PR #765, opened 2014-05-19
Include common Troubleshooting issues section and Windows7 issue

laurencei:
There seems to be a number of 'common' troubleshooting issues with Homestead that people need already.

I've added a troubleshooting section (for future issues) - and included an issue for Windows 7 users that they might need.

See this forum thread for more info: http://laravel.io/forum/05-18-2014-trouble-getting-homestead-to-start-in-windows


ianchanning:
I hit this problem but didn't spot the note that's been added to point this out. It would make more sense if the note was moved to the section where you call `vagrant up` as that's when you hit this problem.

PR #924, opened 2014-09-17
Document Eloquent model routing

dwightwatson:
Add information about the `UrlRouteable` interface and how you can override it to adjust value used by the model in the generation of a route.


taylorotwell:
Done a little differently on 5.0 branch. Thanks.

PR #928, opened 2014-09-21
Update Eloquent table name convention

akshayagarwal:
Add information that Laravel would add underscores to the name of the table if the model name consists of multiple words *and* no table name is explicitly specified within the model.

Eg: For a model named ContentType, Eloquent would automatically assume that this model stores its content within a DB table called content_types


taylorotwell:
Fixed in 5.0 docs.

PR #938, opened 2014-09-30
Reference for @each loop

koomai:
Added a reference to @each (which isn't documented) to the sub-views section.


taylorotwell:
Find each a little confusing rather than foreach and just using include.

PR #968, opened 2014-11-04
[Addition] Quick note to clear a Facade mock

alnutile:
instance. I had to mock a facade in two places of my unit test, and if I did not do this, I ended up getting the state of the first facade overwriting the state of my last facade.


taylorotwell:
This would typically be done with the application refreshes.

alnutile:
got it thanks!

PR #1004, opened 2014-12-26
Info about route names for implicit controllers

lol768:
Documents the third optional parameter for the controller method


taylorotwell:
Added to 5.0 docs.

lol768:
Cheers :)

PR #1008, opened 2015-01-03
Updated Nginx pretty url's

clonerworks:
Removed the pointing to laravel 4.2 docs and added the location info.


GrahamCampbell:
Already fixed.

PR #1009, opened 2015-01-05
Relationship Many-Many tiny fix (role id)

vanimurarka:
Guess it should be "role_id" and not "foo_id"


GrahamCampbell:
You seem to have messed up the branches here.

PR #1011, opened 2015-01-06
Add note about SQS delay limit

davedevelopment:
> The number of seconds (0 to 900 - 15 minutes) to delay a specific message.

- http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html


GrahamCampbell:
:+1:

PR #1013, opened 2015-01-07
[4.2] Don't need to mention that there is a SessionHandlerInterface stub for PHP 5.3

hannesvdvreken:
As PHP 5.3 isn't supported anyway.

[*](https://github.com/laravel/docs/pull/1012#issuecomment-69023347)


taylorotwell:
Removed from master branch.

PR #1042, opened 2015-01-26
Change 'secret' to 'plaintext_password' in security docs

andreyst:
I propose to change 'secret' to 'plaintext_password' (or maybe something else, tell me) to more clearly express that Hash::* accept plaintext password (and not salt or any other kind of secret) as its argument. Not everybody yet knows that bcrypt used by Hash doesn't require any additional salt, so this could be misleading. It actually was in my case, which is the reason I created this pull request.

There is also an instance of 'secret' in Crypt:: example, but I'm not quite sure that 'plaintext_password' should go in its place. Please comment if you think it should also be changed.

### DIFF
--- a/security.md
+++ b/security.md
@@ -28,11 +28,11 @@ The Laravel `Hash` class provides secure Bcrypt hashing:
 
 #### Hashing A Password Using Bcrypt
 
-	$password = Hash::make('secret');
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can you send this to `master` branch with just `plain-text`.

PR #1044, opened 2015-01-26
Added 'string' validation rule

garygreen:


GrahamCampbell:
@garygreen Thanks for your time on this. :)

PR #1046, opened 2015-01-26
Added namespaces to Eloquent relations

JeroenNoten:
Because Eloquent models are now namespaced by default, I added this to the relation definitions in the documentation.


JeroenNoten:
This is to prevent this kind of questions/problems: https://laracasts.com/discuss/channels/general-discussion/retrieving-foreign-key-data-l5

PR #1047, opened 2015-01-27
Adding publish migrations hints

adamgoose:


adamgoose:
Well shoot... In order for this to be an option, [this](https://github.com/laravel/framework/blob/master/src/Illuminate/Foundation/Console/VendorPublishCommand.php#L95-L98) needs to be tweaked to allow merging. How do you (@taylorotwell) propose we make this possible?

brunogaspar:
What/why it needs to be tweaked? 
I tested it earlier today and it worked fine on my end. Doesn't it work on your end? 

Also i don't believe `base_path('database/migrations')` is the correct approach here, what if the application has a different folder structure? Which is completely doable.. 
In that case the migrations will be placed on the wrong directory or maybe it will throw errors due to permissions.

Probably something like `$this->app->databasePath()` should suffice i guess.

adamgoose:
You're probably right, but if you read my linked line, you'll see that if the destination folder exists, he skips publishing said files.

brunogaspar:
Oh i see, sorry i misread it.

brunogaspar:
I guess probably loop through the files and if a file is already present skip it?

GrahamCampbell:
Please resend to 5.0.

PR #1052, opened 2015-01-28
Add back() to responses.md

davidhemphill:
This adds some documentation on the very-handy `back()` method.


juukie14:
Redirecting back is possible with returning ~~~redirect()->back()~~~ and ~~~back()~~~. Very handy!

PR #1066, opened 2015-01-30
Removing repeated parts.

francescomalatesta:
The "Other Configurations" and the "Permissions" parts are already present in the installation.md file. As they are things that I usually do right after install, I thought that we could delete them from here.


taylorotwell:
I like them both places.

PR #1071, opened 2015-02-02
Describing package configs use of .env files

jtgrimes:
I had some trouble figuring out how I could have a package with different configs in different environments. I'm hoping this commit could help someone else having the same problem.


GrahamCampbell:
Please resend to 5.0.

PR #1075, opened 2015-02-03
Add documentation for custom Flysystem adaptors

brenelz:
Feel free to use it or tweak it as you wish.  Figured maybe it would speed up the process, if someone wrote a first draft.

### DIFF
--- a/filesystem.md
+++ b/filesystem.md
@@ -3,6 +3,7 @@
 - [Introduction](#introduction)
 - [Configuration](#configuration)
 - [Basic Usage](#basic-usage)
+- [Custom Filesystems](#custom-filesystems)
 
 <a name="introduction"></a>
 ## Introduction
@@ -98,7 +99,7 @@ The `Storage` facade may be used to interact with any of your configured disks. 
 
 	// Recursive...
 	$directories = Storage::allDirectories($directory);
-
+	
 #### Create A Directory
 
 	Storage::makeDirectory($directory);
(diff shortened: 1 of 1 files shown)


filesystem.md
```diff
@@ -106,3 +107,28 @@ The `Storage` facade may be used to interact with any of your configured disks.
 #### Delete A Directory
 
 	Storage::deleteDirectory($directory);
+
+<a name="custom-filesystems"></a>
+## Custom Filesystems
+If you need to use a filesystem that is not provided out of the box, you may configure one using a service provider. Visit [Flysystem](https://github.com/thephpleague/flysystem#adapters) to view a list of available options.
+
+In order to setup your custom filesystem you call the `extend` method on the `Illuminate\Filesystem\FilesystemManager` class. This class is available to you through the `Illuminate\Contracts\Filesystem\Factory` interface which can be injected into the `boot` method in your service provider.
```
mattstauffer: Jumping straight into the abstract definition here is hard to follow for someone who hasn't done it.

I'd suggest being a little more concrete, or doing the example first? Maybe something like:

```
In order to set up your custom file system, you'll need to create Service a Provider (for example, `DropboxFilesystemServiceProvider`). In its boot method, you'll want to inject an instance of the `Illuminate\Contracts\Filesystem\Factory` contract, and then you'll call the `extend` method of the injected instance.`
```

... etc. etc.

filesystem.md
```diff
@@ -106,3 +107,28 @@ The `Storage` facade may be used to interact with any of your configured disks.
...
+
+The first argument of the `extend` method is the name of the disk you will use, while the second argument is a closure that gets passed both the `$app` and `$config` variables.  
+
+From this closure you must return an instance of `League\Flysystem\Filesystem`. Use the details provided with each filesystem to set this up.
+
+> **Note:** The $config variable will already contain the proper values defined in `config/filesystems.php` for the provided disk.
+
+#### Dropbox Example
+
+	use Illuminate\Contracts\Filesystem\Factory as FilesystemFactory;
```
mattstauffer: I'd consider showing the class name so it's clearer where this should go
```
class DropboxFilesystemServiceProvider etc. etc.
```

filesystem.md
```diff
@@ -106,3 +107,28 @@ The `Storage` facade may be used to interact with any of your configured disks.
...
+	
+	public function boot(FilesystemFactory $filesystem)
+	{
+		$filesystem->extend('dropbox',function($app,$config) {
```
mattstauffer: Make sure to follow Laravel code styles for the examples. For example, this would be 
```php
$filesystem->extend('dropbox', function($app, $config) {
```
(maybe even brace on a new line? Someone else will have to catch that for me.)

Check out other code samples in the docs to see where Taylor puts braces, new lines, spaces, etc.

GrahamCampbell:
I think pointing people to https://github.com/GrahamCampbell/Laravel-Flysystem would make more sense rather than everybody writing the same code over and over.

barryvdh:
I disagree, I say it makes sense to document the extend functionality for the filesystem. No need for a 3rd party package..

brenelz:
It is whatever you guys want.  I can make the updates that Matt suggested, or we can just leave it out.

barryvdh:
@GrahamCampbell Could you re-open this?

brenelz:
Thoughts?

PR #1085, opened 2015-02-04
Add section about public assets

barryvdh:
Add note about publishing package assets to the public folder, including the public tag, --force flag and example public_path() dir


taylorotwell:
Thanks :heart:

PR #1088, opened 2015-02-04
Update links from 5.0 version to use the proper version

filipac:
Right now if i go on the website on Docs, when I am on the 5.0 version and click on a section in the menu it goes back to the master version.


filipac:
Sorry, i was to slow :smile: check #1087

PR #1091, opened 2015-02-04
Shouldn't  it be "homestead up"?

tiran133:


taylorotwell:
No, not unless you are using the homestead PHP helper.

tiran133:
Sure but the whole doc describes the process of using the homestead package. If I would be new to all of this I wouldn't know where to cd in to run the vagrant up command. Just my opinion.

PR #1096, opened 2015-02-04
Added mention of class change Eloquent > Eloquent\Model

cbakker:
When migration a small app I found this missing in the upgrade guide. Fixed it with help from the Laracasts forum: https://laracasts.com/discuss/channels/general-discussion/laravel-50-eloquent-not-found

Added it to the docs to prevent the next users from finding out what happend.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -81,6 +81,8 @@ Feel free to create a new `app/Models` directory to house your Eloquent models. 
 
 Update any models using `SoftDeletingTrait` to use `Illuminate\Database\Eloquent\SoftDeletes`.
 
+If your models extends the Eloquent or Illuminate\Database\Eloquent class, update this to extend the class Illuminate\Database\Eloquent\Model.
+
 #### Eloquent Caching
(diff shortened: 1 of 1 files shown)


upgrade.md
```diff
@@ -81,6 +81,8 @@ Feel free to create a new `app/Models` directory to house your Eloquent models.
 
 Update any models using `SoftDeletingTrait` to use `Illuminate\Database\Eloquent\SoftDeletes`.
 
+If your models extends the Eloquent or Illuminate\Database\Eloquent class, update this to extend the class Illuminate\Database\Eloquent\Model.
```
GrahamCampbell: This isn't really required given that the `Eloquent` alias still exists.

cbakker:
I see, maybe it was because we used 'use Illuminate\Database\Eloquent;' in the header that it did not recognize. Based on the Laracasts post I taught the alias was removed and did not check it myself.

I am sorry guys, will close this PR.

PR #1097, opened 2015-02-04
Fix vagrant for homestead

SelrahcD:
Change `vagrant up` to `homestead up` and `vagrant ssh` to `homestead ssh` as in the documentation in v4.2.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -111,7 +111,7 @@ To add Bash aliases to your Homestead box, simply add to the `aliases` file in t
 
 ### Launch The Vagrant Box
 
-Once you have edited the `Homestead.yaml` to your liking, run the `vagrant up` command from your Homestead directory.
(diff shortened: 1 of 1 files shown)


SelrahcD:
Sorry, might be a duplicate of #1091. Handles `homestead ssh` though.

taylorotwell:
The original wording is correct. Unless you are using the Homestead PHP helper, vagrant up would be correct.

SelrahcD:
I maybe should have provided some more context, sorry.
I struggled for a moment trying to have a webpage showing with vagrant up. I sshed into the VM and discovered that folders weren't shared and guessed the provisionning wasn't done. Using homestead instead did the trick.

Maybe the real issue is with homestead provisioning using vagrant up ? I have possibly missed something in the documentation.

PR #1105, opened 2015-02-05
Changing socialize -> socialite, adding namespace

kingjerod:
I'm not 100% sure if the facade is suppose to have the root namespace, but it's what he's doing on Laracasts: https://laracasts.com/series/whats-new-in-laravel-5/episodes/9


taylorotwell:
That was intentional.

kingjerod:
Why would you want to spell it slightly different? None of the other facades are done that way.

PR #1120, opened 2015-02-06
Fix typo: bad method name of RouteServiceProvider

loicm:
The right method name is `boot` not `before`


taylorotwell:
Wrong branch.

loicm:
You mean I should have proposed the PR on the branch "5.0"?

If so I can do it right a second time.

PR #1124, opened 2015-02-06
Update installation.md

lucasmezencio:
Update installation version, that actually always install the latest version, instead of the documentation version.


GrahamCampbell:
The 4.1 branch isn't maintained.

PR #1125, opened 2015-02-06
Update Socialite docs - authentication.md

sitesense:
Changed to show code example, in line with the composer and facade examples.

### DIFF
--- a/authentication.md
+++ b/authentication.md
@@ -293,7 +293,11 @@ To get started with Socialite, include the package in your `composer.json` file:
 
 	"laravel/socialite": "~2.0"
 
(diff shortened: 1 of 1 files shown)


authentication.md
```diff
@@ -293,7 +293,11 @@ To get started with Socialite, include the package in your `composer.json` file:
 
 	"laravel/socialite": "~2.0"
 
-Next, register the `Laravel\Socialite\SocialiteServiceProvider` in your `config/app.php` configuration file. You may also register a [facade](/docs/5.0/facades):
+Next, register the service provider in your `config/app.php` configuration file:
+
+	'Laravel\Socialite\SocialiteServiceProvider',
+
+You may also register a [facade](/docs/5.0/facades):
```
GrahamCampbell: "alias" is probably a better word to use here rather than "register". It makes it sound like you need to register it in order to use it, which is not true at all.

PR #1128, opened 2015-02-07
Fixed composer installation command & archive link

mnishihan:


GrahamCampbell:
Duplicate. The 4.1 branch is not maintained anymore.

mnishihan:
This is annoying. :-1: 

The official documentation site still provides documentation for 4.0 & 4.1 branches. Also a legacy branch is not officially supported never means that it is not used by anyone on earth. Specially, in enterprise world, it is not always possible to stay up-to-date with a framework's moving pace, when a product has gone into production. I cannot imagine how your reaction will be after hearing that I'm still maintaining a big code base in production that was developed using Laravel 3.x and never felt any need to upgrade to 4.x. This is how enterprise world runs. Laravel is learning this pattern slowly, but from these kind of interactions, I foresee, much longer path to go for Laravel to adopt with enterprise culture.

Lastly, again a legacy branch is not officially supported never means it cannot have update in it's documentation from the community, who are still using the legacy branch somehow in their products.

crynobone:
While I do agree that any typo or incomplete docs should be merged for older version but in this particular case I don't think we should even suggest installing 4.1 anymore (as it no longer maintain, no bugfix or feature).

PR #1140, opened 2015-02-09
Document ability to add custom filesystems

brenelz:
@GrahamCampbell @mattstauffer @barryvdh Sorry about the pull request spam guys.  Just learning a bit about squashing commits and was having trouble.  I think I got it figured out now, so I'll keep this one open and make updates to it as required.

### DIFF
--- a/filesystem.md
+++ b/filesystem.md
@@ -3,6 +3,7 @@
 - [Introduction](#introduction)
 - [Configuration](#configuration)
 - [Basic Usage](#basic-usage)
+- [Custom Filesystems](#custom-filesystems)
 
 <a name="introduction"></a>
 ## Introduction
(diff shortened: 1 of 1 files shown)


brenelz:
Sorry about the pull request spam guys.  Just learning a bit about squashing commits and was having trouble.  I think I got it figured out now, so I'll keep this one open and make updates to it as required.

filesystem.md
```diff
@@ -106,3 +107,41 @@ The `Storage` facade may be used to interact with any of your configured disks.
 #### Delete A Directory
 
 	Storage::deleteDirectory($directory);
+
+<a name="custom-filesystems"></a>
+## Custom Filesystems
+Laravel's Flysystem integration provides drivers for local filesystems, Amazon S3, and Rackspace Cloud Storage. Flysystem is not limited to these and has adapters for many additional storage systems. If you want to use one of these additional adapters in your Laravel application, you'll want to create a custom filesystem.
```
barryvdh: > you'll want to create a custom filesystem.
You want to create a custom driver, not filesystem.

filesystem.md
```diff
@@ -106,3 +107,41 @@ The `Storage` facade may be used to interact with any of your configured disks.
...
+
+In order to set up your custom filesystem you'll need to create a Service Provider (for example, `DropboxFilesystemServiceProvider`). In its `boot` method, you'll want to inject an instance of the `Illuminate\Contracts\Filesystem\Factory` contract, and then call the `extend` method of the injected instance.
+
+The first argument of the `extend` method is the name of the disk you will use, while the second argument is a closure that gets passed both the `$app` and `$config` variables.
```
barryvdh: First argument is the name of the driver.

barryvdh:
Do you also need to add a note that, in order to actually use the filesystem, you need to use the `dropbox` driver in the config, instead of local/s3? Or is that obvious?

brenelz:
You can either change the default driver, or you can use the disk method.  It isn't different from using other drivers but we could add a note.

mattstauffer:
Nice work!

PR #1144, opened 2015-02-09
Remove SSH and HTML

antonioribeiro:
Remote and HTML are not in Laravel, but link is still working: http://laravel.com/docs/5.0/ssh


GrahamCampbell:
Have we removed the html docs too. If not, we should keep these docs.

antonioribeiro:
Keep unlinked docs? It only causes confusion. I googled it, found in 4.2, edited the URL to 5.0 and it worked, so I assumed it was still there, but it's not.

GrahamCampbell:
Agreed. These components still very much exist, and can be still installed into a L5 app.

PR #1152, opened 2015-02-10
Update helpers.md

Donny5300:
The Laravel Requirements are PHP 5.4, but the docs says for the Helpers -> with(): useful for PHP 5.3 :)


GrahamCampbell:
It's not useful for php 5.4 though. :)

Donny5300:
I dont know exactly where to use it, but see it as a notice that its wrong :)

PR #1154, opened 2015-02-11
Change html to php

jasonvarga:
Other Blade snippets use PHP. It looks like HTML doesn't render correctly with Prism.


GrahamCampbell:
It's neither html or php. html is the closest though.

PR #1156, opened 2015-02-11
Fix broken links in docs nav

martinczerwi:
Links had double version number, example: /docs/4.2/4.2/artisan


GrahamCampbell:
fixed already

PR #1159, opened 2015-02-13
Add mention of defining middleware from controller.

edogawaconan:
This would have saved me a lot of time researching how to do it and hopefully will save someone else's time in the future.

### DIFF
--- a/middleware.md
+++ b/middleware.md
@@ -68,6 +68,10 @@ Once the middleware has been defined in the HTTP kernel, you may use the `middle
 		//
 	}]);
 
+### Assigning Middleware From Controller
+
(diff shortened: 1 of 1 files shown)


edogawaconan:
@taylorotwell may I get explanation why it's closed? Is it obvious enough and thus not needed? Or is adding it from controller shouldn't be done?

coffeeshopcoder:
@edogawaconan Specifying middle-ware within a controller seems to be fine as shown in the controllers page of the docs.

As for the content of your contribution, I'm guessing that it would be more appropriate to format it as a "note" rather than a "sub-section". (of course the content would have to be deemed appropriate for addition first.)

PR #1164, opened 2015-02-14
Update docs to show session helper usage

davidhemphill:


GrahamCampbell:
Indentation MUST be with tabs.

PR #1171, opened 2015-02-16
Missing parameters

enigmatic-user:
I guess the variables $email and $password have to be passed as parameters here, or am I missing something?


GrahamCampbell:
These aren't missing. They're just meant to be placeholders.

enigmatic-user:
Well, for people learning PHP and Laravel, this might be very confusing, I guess. It's okay when there's no context at all, just to introduce a method, but here's a (very basic) example implementation, so I think this should be done correctly.

PR #1175, opened 2015-02-17
Update Autoloading Standard in Coding Style

javichito:
Isn't Laravel 5 following the PSR-4 as autoloading standard?


GrahamCampbell:
Yeh, it is. :)

PR #1179, opened 2015-02-20
Add note about X-XSRF-TOKEN and X-CSRF-TOKEN difference

barryvdh:
For https://github.com/laravel/framework/pull/7528

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -86,7 +86,7 @@ In addition to looking for the CSRF token as a "POST" parameter, the middleware 
 
 Laravel also stores the CSRF token in a `XSRF-TOKEN` cookie. You can use the cookie value to set the `X-XSRF-TOKEN` request header. Some Javascript frameworks, like Angular, do this automatically for you.
 
-> Note: The difference between the `X-CSRF-TOKEN` and `X-XSRF-TOKEN` is that the first uses a plain text value and the latter uses an encrypted value, because cookies in Laravel are always encrypted. If you use the `csrf_token()` function to supply the token value, your probably want to use the `X-CSRF-TOKEN` header.
+> Note: The difference between the `X-CSRF-TOKEN` and `X-XSRF-TOKEN` is that the first uses a plain text value and the latter uses an encrypted value, because cookies in Laravel are always encrypted. If you use the `csrf_token()` function to supply the token value, you probably want to use the `X-CSRF-TOKEN` header.
 
 <a name="method-spoofing"></a>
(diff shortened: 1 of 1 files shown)


routing.md
```diff
@@ -73,7 +73,20 @@ Of course, using the Blade [templating engine](/docs/5.0/templates):
...
+	$.ajax({
+	  url: "/foo/bar",
+	  beforeSend: function( xhr ) {
+	    xhr.setRequestHeader('X-CSRF-Token', $('meta[name="csrf-token"]').attr('content'))
+	  }
+	})
+
+Laravel also stores the CSRF token in a `XSRF-TOKEN` cookie. You can use the cookie value to set the `X-XSRF-TOKEN` request header. Some Javascript frameworks, like Angular, do this automatically for you.
+
+> Note: The difference between the `X-CSRF-TOKEN` and `X-XSRF-TOKEN` is that the first uses a plain text value and the latter uses an encrypted value, because cookies in Laravel are always encrypted. If you use the `csrf_token()` function to supply the token value, your probably want to use the `X-CSRF-TOKEN` header.
```
taylorotwell: you probably want* ... you have "your probably want"

barryvdh: Fixed

brunogaspar:
Not sure if it should be on the docs, but it's worth to mention that this can be set using the following on the main layout for example

```
$.ajaxSetup({
	headers: {
		'X-CSRF-Token': $('meta[name="csrf-token"]').attr('content')
	}
});
```

This way you only have to specify the Ajax request and nothing else

```
$.ajax({
    url: "/foo/bar",
});
```

Avoids repetition.

barryvdh:
Yeah, that is perhaps a better example, but both demonstrate the idea I guess :)

brunogaspar:
For sure, just mentioning, as it might be useful for someone :)

TheShiftExchange:
Are headers case sensitive? In your PR the header it checks for is `X-CSRF-TOKEN` - but in your docs you've typed `X-CSRF-Token`?

Also - +1 to the example above - that is how I do it - much easier.

barryvdh:
Not sure, can you submit a PR for consistency and update the example?

TheShiftExchange:
I checked - it is **not** case sensitive - so it doesnt really matter - but I made some slight improvements anyway.

I do it the same way as @brunogaspar - it works flawlessly - so I've odne a PR for that https://github.com/laravel/docs/pull/1182

PR #1187, opened 2015-02-23
Update installation.md to highlight risk of installing to an existing folder

njt1982:
See https://github.com/laravel/framework/issues/5868#issuecomment-75473435 for details.

There is a bug in the installer which, until there is a tagged release, means you cannot install to an existing folder.

This bit me last night. Following the instructions led to an unexpected error.


wmather:
I just ran into this as a first time user of laravel/installer. Ideally the version should be incremented on that repo

PR #1190, opened 2015-02-24
Correct casing of HTML when adding alias

JacobBennett:
In previous versions of laravel, the alias used for the HTML helper was uppercased. Those of us migrating from old installations will still have unresolved classes if using 'Html' for our alias as opposed to 'HTML'.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -181,7 +181,7 @@ You may move your Sass, Less, or CoffeeScript to any location you wish. The `res
 
 ### Form & HTML Helpers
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
PHP is not case sensitive when it resolves these.

JacobBennett:
@GrahamCampbell I'm super confused then as to why mine cared. This happened when resolving names of models as well when I migrated. For instance, if in my tags model I had

```php
public function documents()
	{
		return $this->belongsToMany('document');
	}
```

notice the lowercase, it would fail citing that it couldn't find a class named `document`. If I uppercased it to be.

```php
		return $this->belongsToMany('document');
```

Thoughts? As a side note, I did not namespace all of my models but instead chose to keep them in the global namespace for now. Don't know if that matters.

PR #1194, opened 2015-02-25
fix ViewComposer directory name

lifesign:


taylorotwell:
It was fine before.

GrahamCampbell:
@taylorotwell This issue is that the example above it uses ViewComposers and this one uses Composer.

PR #1195, opened 2015-02-25
added warning note and installation info for the now-removed Form helper

DfKimera:
The Form helper was removed in L5, but the entry is still in Laravel documentation.
The Form helper is now maintained by LaravelCollective

Since it's an old and very well-known feature of Laravel, I propose to leave an advisory note in the Laravel docs indicating the removal, LC's support and installation instructions.


GrahamCampbell:
That package is not official. Our subsplit is still available as illuminate/html.

PR #1197, opened 2015-02-25
Added installation instructions for "Forms & HTML"

DfKimera:
The Forms/Html helpers are no longer bundled in the core, but shipped in a separate package.
The installation instructions for this package are missing, as well as the information that the helpers are no longer bundled in the core.

### DIFF
--- a/html.md
+++ b/html.md
@@ -1,5 +1,8 @@
 # Forms & HTML
 
+> **Warning:** The Form helper was removed from Laravel 5, and is now being maintained by Laravel Collective. For further information, please refer to the [Laravel Collective](http://laravelcollective.com/docs/5.0/html) documentation.
+
+- [Installation](#installation)
 - [Opening A Form](#opening-a-form)
 - [CSRF Protection](#csrf-protection)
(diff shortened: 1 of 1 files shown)


html.md
```diff
@@ -13,6 +14,32 @@
...
 
+<a name="installation"></a>
+## Installation
+
+The Forms & HTML helpers are no longer bundled with Laravel core. 
+
+You can install this package through Composer. Edit your project's `composer.json` file to require `laravel/html`.
+
+    "require": {
+        "laravel/html": "~5.0"
```
GrahamCampbell: Not correct,

taylorotwell:
These documentation pages aren't displayed for the 5.0 docs.

taylorotwell:
Also this information is in the change log.

DfKimera:
If it no longer belongs in the docs, shouldn't the page be removed instead? It still shows up on Google results.

adamgoose:
Live Documentation here: http://laravelcollective.com/docs/5.0/html

taylorotwell:
Removed it just now.​

PR #1199, opened 2015-02-25
filesystem.md: Rename Disk to Storage

svrnm:
Just a minor issue I found while I wrote a custom filesystem: The used facade is called "Disk" there, but it should actually be "Storage"


taylorotwell:
:heart:

PR #1200, opened 2015-02-25
Nginx has a special variable for url arguments

koenpunt:
`$is_args` is set to "?" or an empty string based on the availability of query params


taylorotwell:
Seems to work fine as is.

koenpunt:
Pretty blunt comment if you ask me. I you make all your decisions on what it seems to you.. You could at least asked why this would be better or why I did suggest this change..

PR #1201, opened 2015-02-26
Link to details of Allman style braces, and some dashes.

samwilson:


HipsterJazzbo:
I think this is unnecessary given that everything will change to PSR-2 for 5.1, right?

HipsterJazzbo:
(ICYMI: laravel/framework@c0824c5d2acb46d35125062fa28a76fd0b29feac)

samwilson:
Ah! I did indeed miss that. I'll shift my projects' CS to full PSR2 then, to get a head start. :) Thanks!

PR #1210, opened 2015-02-27
Update cache.md

tomcastleman:
Add section on accessing cache stores other than the default


taylorotwell:
I would put this under cache usage, not give it its own section.

PR #1214, opened 2015-02-28
Add section on routing to helpers document

nikush:


EddyF:
good stuff!

GrahamCampbell:
I wouldn't recommend using these helpers tbh.

taylorotwell:
Routing should be above Strings. Alphabetical Order.

PR #1215, opened 2015-02-28
Added examples for jobs that run on specific days

psampaz:


GrahamCampbell:
Please send all your pulls as one otherwise we're going to get merge conflicts.

PR #1226, opened 2015-03-02
Added example of Filesystem Factory Injection

ben-lowery:


taylorotwell:
We mention you can type hint this contract at the top of the Basic Usage documentation.

ben-lowery:
No problem :) , I thought it might be useful to have an example of how the type hinting would look when expanded out, is type hinting using Factories explained somewhere else that I missed?

PR #1231, opened 2015-03-04
Update Envoy URI.

JerseyMilker:
Fixing Envoy link.


JerseyMilker:
Thanks for adding the Envoy 5.0 docs back in as well.

PR #1235, opened 2015-03-05
Adding initial text addition

jsposato:
Taylor,

Found an article referencing this Blade feature.  I figured I'd submit a PR to document it.  If anything needs to be re-formatted, or if you think more clarification is necessary let me know.


taylorotwell:
I didn't document this intentionally because I personally find it more intuitive to just use a @foreach and and @include.

PR #1237, opened 2015-03-06
Update Homestead docs

joshfng:
Updated the docs to outline using different providers. Also removed the section for the manual `vagrant box add`, as `vagrant up` will automatically download the box for you if it doesn't exist.


joshfng:
I see you added docs. Closing this out.

PR #1239, opened 2015-03-07
[5.0] cache events

lucasmichot:
The documentation was missing

### DIFF
--- a/cache.md
+++ b/cache.md
@@ -134,7 +134,7 @@ In contrast, this statement would remove only caches tagged with `authors`, so "
 <a name="cache-events"></a>
 ## Cache events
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Nice. :+1:

cache.md
```diff
@@ -131,6 +131,27 @@ In contrast, this statement would remove only caches tagged with `authors`, so "
 
 	Cache::tags('authors')->flush();
 
+<a name="cache-events"></a>
+## Cache events
+
+Cache actions can trigger events. You may for instance use `Event` listeners to access the following events.
```
GrahamCampbell: `listeners to access the following events.` Seems odd wording. We would nee a colon here for "the following", but there's not one.

PR #1240, opened 2015-03-08
Update Server Requirements

TheShiftExchange:
Laravel 5 requires `Psy/Psych`. 

`Psy/Psych` requires `Nikic/Php-Parser`

`Nikic/Php-Parser` requires the PHP Exteniosn `tonkenizer` to be installed.

Therefore Laravel 5 requires this extension, otherwise composer will fail to install Laravel 5. Issue was discovered on this SO thread: http://stackoverflow.com/q/28914789/1317935

### DIFF
--- a/installation.md
+++ b/installation.md
@@ -45,6 +45,7 @@ The Laravel framework has a few system requirements:
 - Mcrypt PHP Extension
 - OpenSSL PHP Extension
 - Mbstring PHP Extension
+- Tokenizer PHP Extension
 
 As of PHP 5.5, some OS distributions may require you to manually install the PHP JSON extension. When using Ubuntu, this can be done via `apt-get install php5-json`.
 


GrahamCampbell:
We've required that extension ever since Laravel 4.0. Super closures requires php parser.

neohq:
It should also be on documentation at Server Requirements.

GrahamCampbell:
> It should also be on documentation at Server Requirements

Yes, I'm just explaining that this is NOT a new extension requirement in L5. It's just an existing extension requirement.

neohq:
If a begginer will folow the install tutorial from documentation will bump into this problem and that's the end of the story.

I understand that this is not a direct requirement for L5 but maybe you can point this extension/problem somewhere in documentation.

PR #1241, opened 2015-03-10
Add note about mix.sass() calls

bruno-fernandes:
mix.sass() can only be called once and causes a bit of a confusion as all the others can be called multiple times and documentation doesn't make any reference to it.
https://laracasts.com/discuss/channels/general-discussion/problem-with-mixsass


taylorotwell:
Fixed wording and done.

PR #1242, opened 2015-03-11
Update IronMQ dependency

thangngoc89:
According to ironmq's docs :

Use branch 1.* - laravel-compatible, php-5.2-compatible version. No namespaces. Using default IronMQ servers.

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -27,7 +27,7 @@ The following dependencies are needed for the listed queue drivers:
 
 - Amazon SQS: `aws/aws-sdk-php`
 - Beanstalkd: `pda/pheanstalk ~3.0`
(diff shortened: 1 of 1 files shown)


queues.md
```diff
@@ -27,7 +27,7 @@ The following dependencies are needed for the listed queue drivers:
 
 - Amazon SQS: `aws/aws-sdk-php`
 - Beanstalkd: `pda/pheanstalk ~3.0`
-- IronMQ: `iron-io/iron_mq`
+- IronMQ: `iron-io/iron_mq ~1.0`
```
GrahamCampbell: This needs to be `~1.5`, as per our suggest block in our composer.json...

PR #1244, opened 2015-03-13
Tweaked docs about Blackfire

fabpot:
Blackfire has two sets of credentials: client and server. For homestead, we need the server credentials. Even if it is already mentioned in the docs, people are confused. So, this PR tries to make things a little clearer.

Also, what do you think about renaming the YAML `id` and `token` keys to `server-id` and `server-token` respectively? I can make the needed changes and PRs.


taylorotwell:
Thanks. Yeah I'm fine with the rename - let's support both so we don't break compatibility.

PR #1247, opened 2015-03-14
Fixed order of sections in docs

codenihal:
In a section, where you install with Git, the word alternatively is used. That makes it clear that the composer tool is preferable. But, it comes first, before the Composer - PHP way of installing it. So I switched the orders
Signed-off-by: Nihal Sahu <codenihal@gmail.com>


taylorotwell:
I actually don't prefer the Composer tool.

PR #1252, opened 2015-03-16
Add directory defaults section from original elixir readme

tomhorvat:
When having changed folder structure (for example on shared server with public_html folder) or for custom export directories, elixir should know about that and having elixir.json makes things little easier than having to use parameters in sass or coffee tasks. 

I find this useful and some users don't know where to look in source for this.


taylorotwell:
I'm not a big fan of the elixir.json customization.

PR #1255, opened 2015-03-17
Added example using Form Requests manually

garygreen:
Some cases you might want to call methods on your form request before validation for some more bootstrapping stuff, or maybe you just don't like the dependency injection style of auto validating. It's not immediately obvious how to validate manually, this makes it more clear :-)


taylorotwell:
Hmm, this isn't really the expected usage :)

PR #1259, opened 2015-03-18
Improve installation docs

TheShiftExchange:
This is a small change to clarify the install options.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -60,9 +60,9 @@ If this command fails, you may have an old version of Vagrant that requires the 
 
 ### Installing Homestead
 
-#### Manually Via Git (No Local PHP)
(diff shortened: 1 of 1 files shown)


taiyaei:
The "Option X - " doesn't look fitting when compared to the rest of the docs IMO.  Maybe remove that and add the "Alternatively," to the beginning of the second option?

TheShiftExchange:
Fair point. I'm just trying to highlight that it is either the first method, or the second. It's not immediately obvious.

Currently if you are reading "top to bottom" - it kinda reads as if you do the first option, then the second.

taiyaei:
Agreed, I see that every time I look at the Homestead doc and think to myself "I'll remember to PR a fix for this" but always forget :(

PR #1262, opened 2015-03-20
DropboxFilesystemServiceProvider need to extends ServiceProvider

cahsowan:
The example of DropboxFilesystemServiceProvider Class does not extends anything, and it's not working until it's extending ServiceProvider Class, and also have a register method


GrahamCampbell:
You must indent with tabs.

orloffv:
+1

PR #1276, opened 2015-03-23
Update routing.md

MarkVaughn:
make a note that environments named testing (like my pre-production server) will ignore filters as documented here https://github.com/laravel/framework/blob/4.2/src/Illuminate/Routing/RoutingServiceProvider.php#L32

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -129,7 +129,7 @@ If you need to access a route parameter value outside of a route, you may use th
 
 Route filters provide a convenient way of limiting access to a given route, which is useful for creating areas of your site which require authentication. There are several filters included in the Laravel framework, including an `auth` filter, an `auth.basic` filter, a `guest` filter, and a `csrf` filter. These are located in the `app/filters.php` file.
 
-> **Note:** Filters are disabled for unit testing. This may cause unexpected behaviour if you have an environment named `testing`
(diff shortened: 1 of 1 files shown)


taylorotwell:
Wording is awkward. Why is UnitTesting one word?

GrahamCampbell:
@taylorotwell I thought the 4.2 docs weren't maintained anyway?

MarkVaughn:
Sorry I'm German XD, how can you make it less awkward?

taiyaei:
Hmm, this isn't quite correct and can be very misleading:

> if you have an environment named `testing`

It makes it sound like filters are always disabled no matter what the current environment is as long as there's a `testing` environment.

It'd probably be better if it was something along the lines of:

"When the current environment is \`testing\`, filters are disabled for unit tests."

It's not perfect wording but it's more accurate.

MarkVaughn:
@taiyaei actually... now that I think of it your sentence suggests that it is only disabled for unit tests. When in fact it is disabled completely, which was my point, as the testers on our test servers were able to access restricted areas.

taiyaei:
That does appear to be the case.  Mentioning unit tests should probably just be removed, otherwise the "are disabled for unit tests" part is also misleading. :p

PR #1283, opened 2015-03-25
Mention the ability to queue cookies using the `CookieJar`

jonnybarnes:


taylorotwell:
Just use the Cookie facade.

jonnybarnes:
So `use Cookie;` at the top, then `Cookie::queue()` in a method? Maybe add that as an example in the docs?

taylorotwell:
I don't really think you even need to show an entire controller.

PR #1289, opened 2015-03-26
Removed extra commas from Package documentation

lasselehtinen:


GrahamCampbell:
They're not extra. That's actually our coding style.

PR #1290, opened 2015-03-28
Clarification on accessing data inside a view when passing an array..

morganhein:
using the `view('app.view', $data)` format.

### DIFF
--- a/views.md
+++ b/views.md
@@ -45,7 +45,7 @@ If you wish, you may pass an array of data as the second parameter to the `view`
 
 	$view = view('greetings', $data);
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Using `this` as an example is confusing because PHP uses it to reference a class when you are within an object.

morganhein:
Taylor, I thought about that. I'll update it and make it more clear.

morganhein:
As you can see, i'm still learning to git.

GrahamCampbell:
You must revert the change to the file permissions, and rebase and squash to one commit. Merge commits are not allowed.

PR #1293, opened 2015-03-29
Fix invalid class name

kokokurak:


GrahamCampbell:
No, this is correct as is.

GrahamCampbell:
The other things around it are wrong, and need changing.

kokokurak:
How is that correct? `use` statement calls unexisting class.

GrahamCampbell:
> How is that correct? use statement calls unexisting class.

Yep. the import is wrong. But you changed the only bit that was correct to be wrong, see what I mean?

kokokurak:
Hmm, it's not clear which is correct, it depends how we want to call the class: `PurchasePodcastCommand` (wrong `use`) or `PurchasePodcast` (wrong injection). I guess `PurchasePodcast` is better example for namespaced class, so I chose the latter.

GrahamCampbell:
> Hmm, it's not clear which is correct

No, I'm telling you `PurchasePodcastCommand` is correct.

kokokurak:
Can you explain why?

kokokurak:
This doc: http://laravel.com/docs/5.0/bus also mentions `PurchasePodcast` command, not `PurchasePodcastCommand`.

GrahamCampbell:
It needs the command suffix. Look at the source code, and you'll see why.

GrahamCampbell:
> Can you explain why?

Just a convention.

PR #1296, opened 2015-03-30
templates.md extending Blade example: use createMatcher instead of createOpenMatcher

ArtemFitiskin:
Issue: https://github.com/laravel/framework/issues/8224

> The documentation http://laravel.com/docs/5.0/templates#extending-blade is kinda weird because createMatcher and createPlainMatcher are discussed and when it comes to an example createOpenMatcher is used.

### DIFF
--- a/templates.md
+++ b/templates.md
@@ -151,7 +151,7 @@ The following example creates a `@datetime($var)` directive which simply calls `
 
 	Blade::extend(function($view, $compiler)
 	{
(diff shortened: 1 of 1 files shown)


taiyaei:
I think in your title when you changed it you meant the other way around. :)

ArtemFitiskin:
Sorry, is it OK now? :smiley:

taylorotwell:
I've removed this section form the docs as it is pretty confusing and I don't want to document it until it is less so.

PR #1302, opened 2015-04-02
Made explicit what differentiates before middleware from after middleware

stones:
Small change to the middleware docs to explicitly mention what differentiates the middleware.

This was demonstrated with examples, but never specified *what* differentiated them.

Let me know if you find these small changes annoying, i am just making edits as i read the docs,  trying to understand the new features.


taylorotwell:
I think is OK as is.

PR #1304, opened 2015-04-02
Example for using namespaces in a nested Route::group is wrong

thijsvdanker:
```php
Route::group(['namespace' => 'Admin'], function()
{
    // Controllers Within The "App\Http\Controllers\Admin" Namespace

    Route::group(['namespace' => 'User'], function()
    {
        // Controllers Within The "App\Http\Controllers\Admin" Namespace
    });
});
```

The namespace of the second group should be App\Http\Controllers\Admin\User.

Not sure if this should be the point of the documentation, or that it was nested by accident.

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -254,7 +254,7 @@ You may use the `namespace` parameter in your group attribute array to specify t
 
 		Route::group(['namespace' => 'User'], function()
 		{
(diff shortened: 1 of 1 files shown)


taiyaei:
@thijsvdanker This was merged over in #1306, looks like someone else noticed the same thing

PR #1305, opened 2015-04-02
Update routing.md

thorem:
Updated the comment for Route::group with namespace User

PR #1306, opened 2015-04-03
Add additional "\User" level to namespace example

jholl:
I'm still learning ... but assume nested route::group corresponds to nested namespace.


taiyaei:
Duplicate of #1304 and #1305.

PR #1308, opened 2015-04-03
Added service injection documentation

MarkRedeman:


templates.md
```diff
@@ -138,6 +138,16 @@ To overwrite a section entirely, you may use the `overwrite` statement:
 
 	{{-- This comment will not be in the rendered HTML --}}
 
+#### Service injection
+We can use `@inject` to resolve services from the service container.
+
+	@inject('html', 'HtmlBuilder')
+
+	<div>
+	    {{ $html->image('/assets/images/laravel.png') }}
```
taiyaei: Looks like you switched to spaces here, tabs please

MarkRedeman:
@taiyaei changed and rebased.

taiyaei:
@MarkRedeman Thanks :)

templates.md
```diff
@@ -138,6 +138,16 @@ To overwrite a section entirely, you may use the `overwrite` statement:
 
 	{{-- This comment will not be in the rendered HTML --}}
 
+#### Service injection
+We can use `@inject` to resolve services from the service container.
+
+	@inject('html', 'HtmlBuilder')
+
+	<div>
+		{{ $html->image('/assets/images/laravel.png') }}
```
GrahamCampbell: That is the wrong blade syntax. That would result in escaped html dumped on the page.

templates.md
```diff
@@ -138,6 +138,16 @@ To overwrite a section entirely, you may use the `overwrite` statement:
 
 	{{-- This comment will not be in the rendered HTML --}}
 
+#### Service injection
+We can use `@inject` to resolve services from the service container.
+
+	@inject('html', 'HtmlBuilder')
```
GrahamCampbell: HtmlBuilder is not a class that's shipped with laravel by default, and is also wrong anyway - it's not the FQCN unless you've aliased it. Choose a different example please.

MarkRedeman: Would you rather use an example that is in the Laravel framework (if so, which would be appropriate?)? I think the HtmlBuilder is a nice example as it's probably better to only inject services that really matter when working with html. (vs for example a `StatsService` that could also be included using the `with` function.

templates.md
MarkRedeman: Forgot about the new syntax. Will change it when I change the example service.

taylorotwell:
Would probably rather use an obviously contrived example so people don't assume they can actually inject HtmlBuilder like that out of the box.

MarkRedeman:
@taylorotwell ok, how about the `UrlGenerator`?

GrahamCampbell:
Maybe try something that doesn't have global helper functions - ie, something someone might actually use.

taylorotwell:
Documented.

PR #1309, opened 2015-04-05
Added information about the 'with' attribute on models and eager loading

stones:
BauerUK asked on IRC how to always load the relations and mentioned that it wasn't referenced in the docs.


taylorotwell:
I don't really want to include this as it wasn't really *intended* to work this way and I can't be certain if it works in all various scenarios.

PR #1329, opened 2015-04-13
Fix validation doc.

OlivierBarbier:


taylorotwell:
You can just pass the validator. It still works.

PR #1330, opened 2015-04-14
Added closure serialisation note

drawmyattention:
E-mails that send using Mail::send, can fail to send when using Mail::queue in certain cases.

For example:

    // $data is an object

    Mail::queue('example.template', compact('data'), function($message) use ($recipient)
    {
        $message->to($recipient->email, $recipient->name)->subject('Example subject');
    });

and within the example/template.blade.php

    Outputting something from the data object {{ $data->someval }}

In this instance, due to the serialization of the closure sending can fail. It is therefore necessary to convert `$data` and `$recipient` to an array and change the code within the template.

    $data = $data->toArray();
    $recipient = $recipient->toArray();

    Mail::queue('example.template', compact('data'), function($message) use ($recipient)
    {
        $message->to($recipient['email'], $recipient['name'])->subject('Example subject');
    });

and within the example/template.blade.php, change to:

    Outputting something from the data object {{ $data['someval'] }}

### DIFF
--- a/mail.md
+++ b/mail.md
@@ -135,6 +135,9 @@ If you wish to specify a specific queue or "tube" on which to push the message, 
 		$message->to('foo@example.com', 'John Smith')->subject('Welcome!');
 	});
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
:-1: from me. It's not actually totally correct.

drawmyattention:
@GrahamCampbell care to elaborate?

PR #1333, opened 2015-04-15
Added note to require Guzzle when using Artisan schedule

laurentvd:
See issue at https://github.com/laravel/framework/issues/8433


taylorotwell:
This is really only required when using `thenPing` no?

GrahamCampbell:
Yeh. We just need into include a note abut thenPing really. :)

PR #1334, opened 2015-04-16
IronMQ branch fix

joshhornby:
Have to use 1.* branch has any other branch gives 404 error

See https://github.com/iron-io/iron_mq_php#branches

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -27,7 +27,7 @@ The following dependencies are needed for the listed queue drivers:
 
 - Amazon SQS: `aws/aws-sdk-php`
 - Beanstalkd: `pda/pheanstalk ~3.0`
-- IronMQ: `iron-io/iron_mq ~1.5`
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
No, `~1.5` is correct.

joshhornby:
Doesn't work for me and dropping back down to 1.* did. Also this guide states using v1 http://www.sitepoint.com/ironmq-laravel-setup/

GrahamCampbell:
https://github.com/iron-io/iron_mq_php/pull/55

> Doesn't work for me and dropping back down to 1.* did. Also this guide states using v1 

I don't think you understand what that operator does in composer.

GrahamCampbell:
`1.*` is the same as `~1.5`, only `~1.5` says the min version MUST be `1.5.0` too. That's all.

joshhornby:
My bad, learn something new every day :+1:

PR #1335, opened 2015-04-17
Updating 4.1 upgrade links

jrenton:
The links to upgrade from 4.0 to 4.1 currently point to the master branch which is laravel 5.0


GrahamCampbell:
v4.1.0 is still wrong!

PR #1336, opened 2015-04-17
Added a note to schema.md

rupertrutland:
Changing tables with enum in the column is not supported, similar to renaming columns.


taylorotwell:
Wording is a little off. Will work on it.

PR #1338, opened 2015-04-18
Update commands.md

TheShiftExchange:
Add an example for what you actually need to type to register an Artisan command in the Kernel.

### DIFF
--- a/commands.md
+++ b/commands.md
@@ -126,4 +126,10 @@ Sometimes you may wish to call other commands from your command. You may do so u
 
 #### Registering An Artisan Command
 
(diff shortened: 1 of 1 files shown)


commands.md
```diff
@@ -126,4 +126,10 @@ Sometimes you may wish to call other commands from your command. You may do so u
 
 #### Registering An Artisan Command
 
-Once your command is finished, you need to register it with Artisan so it will be available for use. This is typically done in the `app/Console/Kernel.php` file. Within this file, you will find a list of commands in the `commands` property. To register your command, simply add it to this list. When Artisan boots, all the commands listed in this property will be resolved by the [service container](/docs/5.0/container) and registered with Artisan.
+Once your command is finished, you need to register it with Artisan so it will be available for use. This is typically done in the `app/Console/Kernel.php` file. Within this file, you will find a list of commands in the `commands` property. To register your command, simply add it to this list.
+
+	protected $commands = [
```
GrahamCampbell: This shouldn't be indented, but should be done with a code block.

TheShiftExchange: @GrahamCampbell - in this file - all code is done with indents for some reason. There is no code blocks for any of the other examples?

GrahamCampbell: My bad.

PR #1341, opened 2015-04-20
Remove Option #1 from Homestead installation documentation

mradcliffe:
I was helping someone out who was trying to do `homestead up` without Composer.

If I look at the code in the `./homestead` file I see that there is no possible way that Option 1 can work because the vendor and autoloader would not exist, and the [homestead script requires an autoloader](https://github.com/laravel/homestead/blob/master/homestead#L7). The `./scripts/init.sh` script does not provide these files either. I looked in `./scripts/homestead.rb`, and did not find anything relevant there either. I assume that there is no way to install Homestead without Composer, and I think it would be best to remove this from the documentation in order to improve the experience of new Homestead users.

I'm not familiar with Laravel or Homestead, and so I may be missing some bit of code or Vagrant logic where this magically works, but from [this laracast](https://laracasts.com/discuss/channels/general-discussion/step-by-step-setup-guide-for-homestead-on-os-x) it seems that Option 1 is discouraged and not supported.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -60,19 +60,9 @@ If this command fails, you may have an old version of Vagrant that requires the 
 
 ### Installing Homestead
 
-#### Option 1 - Manually Via Git (No Local PHP)
+PHP and [Composer](https://getcomposer.org/) are required to install Homestead.
 
-If you do not want to install PHP on your local machine, you may install Homestead manually by simply cloning the repository. Consider cloning the repository into a `Homestead` folder within your "home" directory, as the Homestead box will serve as the host to all of your Laravel (and PHP) projects:
-
-	git clone https://github.com/laravel/homestead.git Homestead
-
-Once you have installed the Homestead CLI tool, run the `bash init.sh` command to create the `Homestead.yaml` configuration file:
-
-	bash init.sh
-
-The `Homestead.yaml` file will be placed in your `~/.homestead` directory.
-
-#### Option 2 - With Composer + PHP Tool
+#### Composer + PHP Tool
 
 Once the box has been added to your Vagrant installation, you are ready to install the Homestead CLI tool using the Composer `global` command:
 


taylorotwell:
It certainly does work because that's the way I use Homestead.

adfaklsdjf:
Here's the problem.. the documentation later says to use "homestead up" to start up the VM, but this doesn't work because the "homestead" command requires composer.. Option #1 implies that you can complete the rest of the steps without php and composer installed.

From my perspective, this meant that Option #1 doesn't work.. but another developer on our team, Eric Katz, told us that you can use "vagrant up" instead of "homestead up".  I'd considered that possibility, but I noticed that earlier in the documentation it says you can use "homestead provision" or "vagrant provision".. and in this instance it didn't mention using "vagrant up", which seemed to me to imply that "homestead up" was necessary.

The doc could simply be modified to say "or use 'vagrant up' if you followed option #1 above".  As it stands now, if you take option #1 because you don't have a full php stack running in your local environment.. following the rest of the instructions leaves you at a dead end at "homestead up" unless you know to use "vagrant up" instead.

Eric feels it would be easiest to understand if the rest of the document were split into option #1 and option #2.. but barring that, I really think it should say to use "vagrant up" if you don't have a full local php dev environment set up.

adfaklsdjf:
You can observe another user running into numerous issues because he doesn't have a full php dev environment set up in his local environment here, and being treated like an idiot by the responders because he doesn't, here: https://laracasts.com/discuss/channels/general-discussion/step-by-step-setup-guide-for-homestead-on-os-x

The people in the thread scoff at him for not having composer already set up and ultimately make him set up brew and composer on his local environment so he can finally run the "homestead up" command successfully.  The implication seems to be that if you don't have a full php dev environment, you're not cut out to use homestead or laravel.  It makes it look like Option #1 isn't really an option.

PR #1343, opened 2015-04-21
Add directions for initiating and launching VM without PHP/Homestead CLI

ericnkatz:
This is hopefully a clarity fix in reference to issue #1341. This will help beginners follow through instructions provided from Option 1 in launching their Homestead without PHP being installed on their host computer.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -138,7 +138,7 @@ To add Bash aliases to your Homestead box, simply add to the `aliases` file in t
 
 ### Launch The Vagrant Box
 
-Once you have edited the `Homestead.yaml` to your liking, run the `homestead up` command from your Homestead directory.
+Once you have edited the `Homestead.yaml` to your liking, it is time to boot up your Homestead. If you chose option 1 earlier, run the `vagrant init laravel/homestead` command in your Homestead directory and then the `vagrant up` command to launch. If you installed with option 2 run the `homestead up` command from your Homestead directory.
 
 Vagrant will boot the virtual machine, and configure your shared folders and Nginx sites automatically! To destroy the machine, you may use the `vagrant destroy --force` command.
 


adfaklsdjf:
Please note I commented on #1341 in an attempt to better explain how the docs currently lead beginners into pitfalls.

joelclermont:
@ericnkatz This approach will still fail because the homestead script requires two dependencies that are installed via composer, which requires PHP. In order to truly eliminate any need for PHP on the host, you'd have to refactor all the instructions that reference homestead commands to use vagrant commands.

ericnkatz:
@joelclermont the idea is option 1 doesn't actually use the php pieces of the homestead script. It only uses the [init.sh](https://github.com/laravel/homestead/blob/master/init.sh) script that just creates the ~/.homestead directory and copies the contents of [stub](https://github.com/laravel/homestead/tree/master/src/stubs) to that new directory.

If you followed option 1 you should never be running any `homestead *`commands in your terminal.

taylorotwell:
The 5.1 docs already only present the manual Git way as the only installation option as I really think it's probably the better option long run.

taylorotwell:
I went ahead and just made the 5.1 Homestead docs the 5.0 docs as well.

joelclermont:
@taylorotwell Removing the composer installation instructions simplifies things quite a bit. Thanks! So with this recommended approach, you'd never use the homestead command, nor install any of the composer requirements for the homestead command. 

I'm just thinking this through: Is there any value then in still having the homestead script and the composer dependencies and all the custom commands in src? It seems like if that's all still going to be there, we should have some note explaining that if you have PHP on the host box, you can composer install and use the homestead commands (possibly even adding it to your path)? If that's not recommended at all, then maybe drop those unused extras for simplicity?

PR #1345, opened 2015-04-21
Specify the kilobyte measurement convention

msimkunas:
The documentation should be updated to clear any doubts regarding the size of a kilobyte in the `size` validator rule.

Refer to this question: http://stackoverflow.com/questions/29767234/laravel-4-validator-file-size/29767440#29767440

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -563,7 +563,7 @@ The given _field_ must match the field under validation.
 <a name="rule-size"></a>
 #### size:_value_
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
We could just kibibytes instead.

msimkunas:
Sure, whatever works.

taylorotwell:
Yes, just use Kibibytes.

PR #1346, opened 2015-04-21
Documentation on Blade's directive() helper

franzliedke:
As requested in https://github.com/laravel/framework/pull/8425.

### DIFF
--- a/templates.md
+++ b/templates.md
@@ -164,5 +164,5 @@ The following example creates a `@datetime($var)` directive which formats a give
 
 	Blade::directive('datetime', function($expression)
 	{
-		return "<?php echo $expression->format('m/d/Y H:i'); ?>";
+		return "<?php echo {$expression}->format('m/d/Y H:i'); ?>";
 	});
--- a/templates.md
+++ b/templates.md
@@ -3,6 +3,7 @@
 - [Blade Templating](#blade-templating)
 - [Other Blade Control Structures](#other-blade-control-structures)
 - [Service Injection](#service-injection)
+- [Extending Blade](#extending-blade)
 
 <a name="blade-templating"></a>
 ## Blade Templating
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can you show the generated raw PHP output of the example?

franzliedke:
You mean the PHP code of the compiled template?

taylorotwell:
Right.

franzliedke:
Thanks for the heads-up, I had to fix it.

This code:

    Blade::directive('datetime', function($expression) {
        return "<?php echo {$expression}->format('m/d/Y H:i'); ?>";
    });

For this Blade template:

    @datetime($foo)

Results in this PHP:

    <?php echo ($foo)->format('m/d/Y H:i'); ?>

taylorotwell:
Is that `echo ($foo)->` syntax even valid? That doesn't even parse for me and I've never seen syntax like that.

taylorotwell:
Seems like you would need something like `with{$expression}->format` which would turn into `with($foo)->format()`

franzliedke:
Ah, you're right. This [would only work in PHP 7](http://3v4l.org/B9rAH).

I just changed it to use `with()`. Or should we change the code so that the parenthesis won't be sent to the closure as they're implicit?

TheShiftExchange:
...so is this also valid syntax?

```
Blade::directive('datetime', function($expression, $format) {
    return "<?php echo {$expression}->format($format); ?>";
});
```

For this Blade template:

```
@datetime($foo, 'd-m-Y')
```

Results in this PHP:

```
<?php echo ($foo)->format('d-m-Y'); ?>
```

franzliedke:
@TheShiftExchange No, that would only work in PHP 7. You should do this:

```php
Blade::directive('datetime', function($expression, $format) {
    return "<?php echo with{$expression}->format($format); ?>";
});
```

PR #1347, opened 2015-04-21
Adding guzzle requirement note

adamgoose:


artisan.md
```diff
@@ -170,3 +170,7 @@ In this example, the `foo` command will be run every minute if it is not already
 #### Ping A Given URL After The Job Runs
 
 	$schedule->command('foo')->thenPing($url);
+
+Using the `thenPing($url)` feature require that the Guzzle 4 HTTP library be installed into your application. You can add Guzzle 4 to your project by adding the following line to your `composer.json` file:
```
taylorotwell: I'm pretty sure it's Guzzle 4 or 5, and the "require" should be "requires".

adamgoose: Okay, will update here *and* in the `mail.md`, because that's where I copied this from. ;)

adamgoose:
Does that look good, or would you rather I use `4.*|5.*`?

GrahamCampbell:
We're only recommending guzzle 5 in composer - ie, we're saying we don't support guzzle 4. I think we should document it like that too.

GrahamCampbell:
Even though it would probably work with guzzle 4.

adamgoose:
Okay, so if @taylorotwell agrees, I will remove the recommendation for Guzzle 4, and update `mail.md` to reflect as well.

taylorotwell:
Yes, agree.

taylorotwell:
The "require" typo still wasn't fixed.

adamgoose:
Bahaha sorry... Sending another now.

taylorotwell:
I went back and fixed it. Just don't merge things until I can sign off on them.

PR #1352, opened 2015-04-22
Improve homestead installation instructions

joelclermont:
Added instructions for running composer and setting path for the git-based installation method. Also normalized the post-installation tasks since they work the same for git or composer installs. Finally, removed all references to vagrant commands since these require you to run them from a specific folder depending on the installation method, whereas the corresponding homestead commands can be run from anywhere on the system.

Note: I realize that adding a task for `composer install` in the git instructions means you now have to have PHP installed, but there's no way around it since Homestead relies on two composer dependencies anyways.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -66,11 +66,11 @@ If you do not want to install PHP on your local machine, you may install Homeste
 
 	git clone https://github.com/laravel/homestead.git Homestead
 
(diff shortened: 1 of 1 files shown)


joelclermont:
I'm not happy with this approach since it defeats the whole point of option 1. I have a better idea, which I'll submit as a different PR.

PR #1354, opened 2015-04-23
Update pagination.md

TheShiftExchange:
There is no `lastPage` function.

Updated to fix.


GrahamCampbell:
:+1:

taylorotwell:
There is on `LengthAwarePaginator`

GrahamCampbell:
We need to make that clear in the docs then, otherwise this is confusing.

PR #1364, opened 2015-04-30
Where does not accept closure anymore. Describe the whereNested method

emilv:
The previous example did not work in Laravel 5 since the `where` method expects a string. I found the [whereNested](http://laravel.com/api/master/Illuminate/Database/Query/Builder.html#method_whereNested) method which seems to be the new way of building advanced WHERE clauses.

I also changed some ORs and ANDs around to make a better point. AND takes precedence over OR so in the previous example we wouldn't even need the parentheses.

### DIFF
--- a/queries.md
+++ b/queries.md
@@ -185,20 +185,20 @@ If you would like to use a "where" style clause on your joins, you may use the `
 
 #### Parameter Grouping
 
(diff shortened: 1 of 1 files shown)


emilv:
Wrong branches. Didn't read the contribution guidelines. Sorry.

PR #1365, opened 2015-04-30
Where does not accept closure anymore. Describe the whereNested method

emilv:
The previous example did not work in Laravel 5 since the `where` method expects a string. I found the [whereNested](http://laravel.com/api/master/Illuminate/Database/Query/Builder.html#method_whereNested) method which seems to be the new way of building advanced WHERE clauses.

I also changed some ORs and ANDs around to make a better point. AND takes precedence over OR so in the previous example we wouldn't even need the parentheses.

### DIFF
--- a/queries.md
+++ b/queries.md
@@ -185,20 +185,20 @@ If you would like to use a "where" style clause on your joins, you may use the `
 
 #### Parameter Grouping
 
-Sometimes you may need to create more advanced where clauses such as "where exists" or nested parameter groupings. The Laravel query builder can handle these as well:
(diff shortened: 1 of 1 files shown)


taylorotwell:
It does still accept Closures. Just tested it. You can also see there is a unit test for it that is passing. Must have been user error.

emilv:
You are right, I just tried the exact example from the docs and that works. I'm using Eloquent so perhaps there's something fishy in there instead. This code did not work on a belongsToMany relationship:

    ->where(function ($query) use ($date) {
        $query
            ->where('channel_product.active', true)
            ->orWhere('channel_product.activate_at', '<=', $date);
        })->get();

While the same works with whereNested. Not going to dig deeper into this, but now you know why I thought the docs were wrong.

PR #1369, opened 2015-05-02
update laravel from 5.0.16

a-moses:
more simple for newbie

Handles  with any problems that may appear


GrahamCampbell:
Not FROM, TO is correct.

PR #1372, opened 2015-05-04
Improved Blade templating description

ludo237:
Wrote a simple paragraph that shows the `@parent` method for extending a `@section` without a full overwrite of it.


taylorotwell:
This is already mentioned in this docs page.

PR #1385, opened 2015-05-08
paths fix

a-moses:
paths fix

### DIFF
--- a/elixir.md
+++ b/elixir.md
@@ -183,7 +183,7 @@ elixir(function(mix) {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
assets is correct

GrahamCampbell:
See https://github.com/laravel/docs/commit/dfc531432adc6bc9c27a4f1a73fc0e5bc1074a46.

a-moses:
Hello  GrahamCampbell 

What about the   elixir("css/all.css")  this is correct to ?

I think it is should be  assets and no elixir :)

PR #1388, opened 2015-05-11
Fix docblock in code sample

OlivierBarbier:


GrahamCampbell:
4.2 is not maintained.

PR #1407, opened 2015-05-15
Adding information about the queue method on queued events

christopherrolfe198:
I discovered this earlier by digging through the codebase to find out how to specify what tube my beanstalk queues use.

I guess the overall use could be how to manipulate the queue instance itself for a queued event if someone needed to.


GrahamCampbell:
It might be better send this to 5.1 instead. // cc @taylorotwell

taylorotwell:
Yes.

PR #1409, opened 2015-05-15
Document collection methods

JosephSilber:


GrahamCampbell:
This page now feels very big. Maybe we need to break it up a bit?

JosephSilber:
@GrahamCampbell I don't think so. There isn't really any logical grouping to them.

Who cares that it's a big page. It's a list of method definitions.

I think it's just fine the way it is. Easy to find everything in one place.

lagbox:
I'm glad someone put it together. :)

GrahamCampbell:
@JosephSilber I don't really agree. If you want one place, look at the documentation in the source, and the tests for the class.

psampaz:
I agree @JosephSilber. No one will care that this is a big page. 

Everyone will love to see this in the docs. 

Good job Joseph!

brunogaspar:
:+1: @JosephSilber

PR #1421, opened 2015-05-17
Homestead: Windows bug with npm

JorisVanEijden:
Homestead:
Windows users should use Vagrant 1.7.3 or higher. Older versions have a [bug](https://github.com/mitchellh/vagrant/pull/5495) that prevents `npm install` from working.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -20,6 +20,8 @@ Homestead runs on any Windows, Mac, or Linux system, and includes the Nginx web 
 
 Homestead is currently built and tested using Vagrant 1.7.
 
+> **Note:** Windows users should use Vagrant 1.7.3 or higher. Older versions have a [bug](https://github.com/mitchellh/vagrant/pull/5495) that prevents `npm install` from working.
+
 <a name="included-software"></a>
 ## Included Software
 


taylorotwell:
In general, you should just always use the latest Vagrant version.

JorisVanEijden:
Well, latest Vagrant version is 1.7.2 so I though it'd be nice to let windows users know that npm won't work for them in homestead  ;)

Jakobud:
The problem here is not `npm install` not working. The problem is Windows inherit limit to the length of a file/directory path is 260 characters.

The problem is that when npm is installing dependencies, in many cases it creates these very deep directory structures like `/node_modules/whatever/something/node_modules/this/that/etc/etc/....` and if that path is >260 characters, then Window can't handle it, because this is happening on your Shared Folder that exists on the Windows host machine. So `npm install` starts giving all these errors because it simply cannot create these deep directory structures.

The solution is to remove your shared folder config in your Vagrantfile and instead do this:
```
# Automatically create www directory on host if it doesn't exist
unless File.directory?(File.dirname(__FILE__) + "/www")
  FileUtils.mkdir_p(File.dirname(__FILE__) + "/www")
end

# Create the shared folder manually
config.vm.provider "virtualbox" do |v|
    v.customize ["sharedfolder", "add", :id, "--name", "www", "--hostpath", (("//?/" + File.dirname(__FILE__) + "/www").gsub("/","\\"))]
end

# Mount shared folder on guest environment
config.vm.provision :shell, inline: "mkdir /home/vagrant/www", run: "always"
config.vm.provision :shell, inline: "mount -t vboxsf -o uid=`id -u vagrant`,gid=`getent group vagrant | cut -d: -f3` www /home/vagrant/www", run: "always"
```

This will essentially create a shared folder `www` on your host that is mounted at `/home/vagrant/www` on the guest machine. Only this approach will fix the 260 character limit that Windows usually has. Using this approach, `npm install` will work fine.

The next Vagrant version should have this fixed, but use the above in the meantime.

PR #1422, opened 2015-05-17
Some collections corrections

JosephSilber:


collections.md
```diff
@@ -344,10 +352,11 @@ You may even pass a callback as the default value. The result of the callback wi
 <a name="method-groupby"></a>
 #### `groupBy()` {#collection-method}
```
GrahamCampbell: Unrelated to this PR, but why do all of these have `groupBy()`, or equiv. That's not the function definition. Surely we should ditch the brackets, or actually write the function definition down?

collections.md
```diff
@@ -898,7 +913,7 @@ In addition, you can pass a third argument containing the new items to replace t
 <a name="method-sum"></a>
 #### `sum()` {#collection-method}
 
-The `sum` method returns the sum of all the items in the collection:
+The `sum` method returns the sum of all items in the collection:
```
GrahamCampbell: This reads better with the `the` still there.

collections.md
JosephSilber: That's how I did it originally. Taylor changed it. I guess he likes this better.

GrahamCampbell: Grrr. I'll talk to him about it. :)

Thank you for all your work on our collections, both in docs and implementation. :)

PR #1426, opened 2015-05-17
Updated instalation.md. Added IIS config file for pretty urls

jatubio:
I'm using IIS and I think there are probably also many users running Laravel on IIS ;)


taylorotwell:
Honestly I would rather this be placed in a community blog or something because I simply don't have access to IIS to verify or keep this up to date.

jatubio:
I understand your point of view. :+1: 

Thank you for your fast reply and take my proposal in consideration.

PR #1438, opened 2015-05-19
Adds implicit controller docs back

wayneashleyberry:


taylorotwell:
This is honestly kind of a bad way to route things in my opinion. It will stay in the framework but not really going to document it.

wayneashleyberry:
Thanks for the feedback @taylorotwell, no problem

PR #1440, opened 2015-05-20
cognitive Title

a-moses:
Title  more close to the context


Stichoza:
It's very confusing. **Table Names** is a better title IMHO

a-moses:
maybe it's confusing . but "Names" more confuse me .
the names himself  it is not the subject of this section

PR #1445, opened 2015-05-22
Update.md , changes needed to be able to test.

sleimanx2:
We should add a $baseUrl variable to the TestCase class. Should I create a Test title to post those two changes under it ?


taylorotwell:
I will come back and update these docs.

PR #1457, opened 2015-05-24
Clarification about the use of transactions in tests

jatubio:
Wrapping Tests In Transactions

`will automatically wrap each test case in a transaction` sound like `wrap each method of Test class in a transaction` and I think it's better to clarify that transactions are by 'test case` and not by `test methods`. (I have get confused a few times on reading)


jatubio:
I had one error on my Test clases... lol.

Text is correct, each test method is wrapped in one transaction

PR #1467, opened 2015-05-26
Added Missing Homestead Installation Information

rajankarthiks:
Information for installing Homestead with PHP + composer was missing and those were added accordingly.


taylorotwell:
I removed them on purpose. I want to prefer the Git installation.

PR #1474, opened 2015-05-27
Add `static` keyword.

Andreyco:
Method declaration was missing `static` keyword.


taylorotwell:
Because the method isn't static?

PR #1475, opened 2015-05-28
Remove array_fetch

JosephSilber:


helpers.md
```diff
@@ -153,20 +152,6 @@ The `array_except` method removes the given key / value pairs from the array:
 
 	// ['name' => 'Desk']
 
-<a name="method-array-fetch"></a>
-#### `array_fetch()` {#collection-method}
-
-The `array_fetch` method returns a flattened array containing the values of the specified nested element:
```
crynobone: Shouldn't we say, "This been replaced with `array_pluck`" or something... Since they might be people that been using it and now don't know what to do.

JosephSilber: It hasn't been removed, just deprecated.

It's been marked as deprecated in the source. If you have a proper IDE, it'll warn you that it has been deprecated. The message says what the alternative is.

PR #1480, opened 2015-05-28
Alternate methods for registering commands and middleware

mreschke:
I believe it is important to list the alternate more dynamic methods of registering command and middleware for us package maintainers. I added to the individual middleware/artisan.md AND to packages.md. If anything, at least keep this alternate method documented in the packages.md


taylorotwell:
I think it's perhaps a little better to simply document your middleware and let people insert them into the kernel themselves in the order they want.

PR #1484, opened 2015-05-30
Update testing with missing method.

eriktisme:
The method **seePageIs** causes 

    Call to undefined method


taylorotwell:
I don't like the way onPage... I'll change it.

PR #1496, opened 2015-06-01
Add missing use App\User

antonybudianto:
add missing use statement for User class


mail.md
```diff
@@ -61,6 +61,7 @@ The send a message, use the `send` method on the `Mail` [facade](/docs/{{version
 	use Mail;
 	use Illuminate\Http\Request;
 	use App\Http\Controllers\Controller;
+	use App\User;
```
GrahamCampbell: Please sort imports by length.

PR #1511, opened 2015-06-04
changes to config/app.php cipher

Kryptonit3:
changes to config/app.php cipher

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -11,6 +11,12 @@
 <a name="upgrade-5.1.0"></a>
 ## Upgrading To 5.1.0
 
+### Update `config/app.php`
+
+Update the `'cipher'` key in `config/app.php` to the following:
+
+	'cipher' => 'AES-128-CBC',
+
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This isn't quite true.

GrahamCampbell:
You need to say, IF you were using that cipher, change it to that.

GrahamCampbell:
And if you were using the 256 one, change it to that.

taylorotwell:
I'll get this updated.

Kryptonit3:
If everyone can use a different cipher then why wasn't the one from 5.0 working? Didn't know you could choose your cipher.

GrahamCampbell:
> If everyone can use a different cipher then why wasn't the one from 5.0 working? Didn't know you could choose your cipher.

You can't. We only support two ciphers and that's it.

GrahamCampbell:
We support the one that was the default in laravel 4, and we support the one that's currently the default, and that's it.

PR #1523, opened 2015-06-06
Added note regarding the Laravel 4.2 HTML facade alias

chriskonnertz:
This is only a proposal. However, please add a note to the 5.0 upgrade guide that mentions the differences.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -202,6 +202,8 @@ Next, add these lines to the 'aliases' array:
 	'Form' => 'Collective\Html\FormFacade',
 	'Html' => 'Collective\Html\HtmlFacade',
 
+> **Note:** In Laravel 4.2, the HTML facade alias is 'HTML' - not 'Html'.
+
 ### CacheManager
 
 If your application code was injecting `Illuminate\Cache\CacheManager` to get a non-Facade version of Laravel's cache, inject `Illuminate\Contracts\Cache\Repository` instead.


GrahamCampbell:
:-1:

GrahamCampbell:
This isn't true at all. You can alias it to what you want...

GrahamCampbell:
It also doesn't matter, as class aliases are NOT case sensitive.

chriskonnertz:
Hey @GrahamCampbell,

yes, one can alias it 'HTML' or 'HTML' or whatever. But that's not the point. The point is, if you upgrade the alias is 'HTML' as in 4.1, but the docs recommends to call it 'Html'. So one could name it 'Html' even though it's called 'HTML' in the application's "legacy" Blade templates.

GrahamCampbell:
That's not a breaking change though, and is already obvious.

GrahamCampbell:
Also, it's not been HTML by default for a long time.

chriskonnertz:
It' 'HTML' since... I don't know, *forever*?

https://github.com/laravel/laravel/blob/v4.2.11/app/config/app.php#L169

GrahamCampbell:
Over a year now.

chriskonnertz:
Are we talking about the same topic?

The commit that added the alias notes is not a year old, it's from January:

https://github.com/laravel/docs/commit/3958a6ebd59639917e323ef097424832585721fe#diff-e6fbd6eb4a701f2e57b69c2f4696dbaeR190

chriskonnertz:
Plus, the case matters. I can't tell you why at this moment. Have not figured it out yet.

GrahamCampbell:
Case really doesn't. PHP's classes are not case sensitive.

GrahamCampbell:
It would only matter for autoloading on a case sensitive filesystem, but aliased classes are not autoloaded...

chriskonnertz:
I'm going to test it in another environment.

chriskonnertz:
It's a general bug. I've created a DigitalOcean Droplet with Ubuntu and a LAMP-Stack. Uploaded my Laravel 5 application that follows the upgrade guide regarding the HTML facade. Check out these links:
http://46.101.171.248/ ('HTML')
http://46.101.171.248/test ('Html')

The only difference is the case. I still have no clue why it's a problem - **but it's a problem**.

chriskonnertz:
@GrahamCampbell 
Still can't figure it out. Probably I've missed something - but what?

Tested it on Windows and Linux, with two Laravel projects. I cannot explain this to myself.

GrahamCampbell:
PHP isn't case sensitive when it resolves classes.

chriskonnertz:
Just found this old PR and could not hold me backing from replying. :D 

So I guess we are **both** right.

PHP does not care. It is not case senstive when it resolves classes.
But the _autoloader_ might be and I assume that this was the case here.

PR #1524, opened 2015-06-06
Custom Error Messages Method

auchEnia:
A new sub-section was added to Form Request Validation. This subsection is supposed to be a reference only to show it is possible to override the error messages in a custom Request class. A link is provided to the Custom Error Messages section to give a quick link to more information and examples.


GrahamCampbell:
This should probably go to 5.1 because changes to 5.0 will not get merged into the 5.1 docs.

PR #1535, opened 2015-06-09
Added Illuminate\Broadcasting\BroadcastServiceProvider to providers a…

lukeprentice202:
Added instructions for adding Illuminate\Broadcasting\BroadcastServiceProvider to providers array so that the new broadcasting events work.

The following error was being thrown otherwise in the queue.

```
[Illuminate\Contracts\Container\BindingResolutionException]                  
  Target [Illuminate\Contracts\Broadcasting\Broadcaster] is not instantiable.  
```


GrahamCampbell:
:-1:

GrahamCampbell:
This isn't true. This is needed to use a new feature, so it's not suitable for the "upgrading" guide.

lukeprentice202:
Ah ok, thanks

PR #1540, opened 2015-06-10
[5.1] Show that null is returned if key doesn't exist.

fungku:
This would've made things go just a bit quicker for me today if I didn't have to verify this myself.

https://github.com/fungku/docs/blob/5.1/artisan.md#retrieving-input

### DIFF
--- a/artisan.md
+++ b/artisan.md
@@ -197,7 +197,7 @@ Options may be retrieved just as easily as arguments using the `option` method. 
     // Retrieve all options...
     $options = $this->option();
 
-If either requested argument or option keys don't exist, a `null` value is returned:
+If either requested `argument` or `option` keys don't exist, a `null` value is returned:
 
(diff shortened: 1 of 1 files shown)


artisan.md
```diff
@@ -197,6 +197,15 @@ Options may be retrieved just as easily as arguments using the `option` method.
     // Retrieve all options...
     $options = $this->option();
 
+If either requested `argument` or `option` keys don't exist, a `null` value is returned:
+
+    // If the userId argument isn't defined, $userId will be set to null.
+    $userId = $this->argument('userId');
+
+    // If the queue option isn't set, $queueName will be set to null.
```
fungku: We could also just comment the respective lines when they are used earlier with the same comments instead of re-displaying the same code.

taylorotwell:
Done.

PR #1547, opened 2015-06-10
[5.1] Add Amazon S3 driver upgrade instructions from 5.0 to 5.1

mpociot:
The new flysystem-aws version is already stated in the filesystem documentation, but it's missing from the upgrade guide.


taylorotwell:
Thanks!

PR #1554, opened 2015-06-11
Fixed wrong type in Request docs

rudifrudi:
Methods 'only' and 'except' accept an array. See
http://laravel.com/api/5.1/Illuminate/Http/Request.html#method_only and
http://laravel.com/api/5.1/Illuminate/Http/Request.html#method_except.

### DIFF
--- a/requests.md
+++ b/requests.md
@@ -145,9 +145,9 @@ You may also retrieve all of the input data as an `array` using the `all` method
 
 If you need to retrieve a sub-set of the input data, you may use the `only` and `except` methods. Both of these methods accept a single `array` as their only argument:
 
-	$input = $request->only('username', 'password');
+	$input = $request->only(['username', 'password']);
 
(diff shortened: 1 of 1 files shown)


2tunnels:
As you can see here: https://github.com/laravel/framework/blob/5.1/src/Illuminate/Http/Request.php#L295 `only` method can accept not only an array. Mby docblock should be updated.

rudifrudi:
I stand corrected! Fact remains though that the 'only' method still doesn't accept multiple parameters, so it should be an array in the docs.

2tunnels:
It does accept multiple parameters.

If you pass array, array will be used. If you pass multiple parameters, they will be converted into array.

http://php.net/manual/en/function.func-get-args.php

rudifrudi:
Wow, I seem to have lost my glasses this morning. Sorry. You're totally right. I'll close the pull request :) :+1:

PR #1555, opened 2015-06-11
update composer.json

gaaarfild:


GrahamCampbell:
:-1:

GrahamCampbell:
This is obvious, and is true for every release.

gaaarfild:
I did it as in the previous update guides. If not necessary in this one, okay.

PR #1556, opened 2015-06-11
Fix Naming Controller Routes

vluzrmos:


taylorotwell:
Wrong branch. Also worded wrong.

PR #1557, opened 2015-06-11
Fix cipher in config/app.php

trupedia:
Allow using Laravel 5.1 without Mcrypt.


taylorotwell:
There is more to it than that. If the person is using a 16 character key instead of 32, that would not work.

trupedia:
So how we can address this in the upgrade instructions? 
Think this is a sensitive part, if you just upgrade an app which uses encryption and forget about fixing the cipher?
Or should one recommend not to upgrade an app, which uses a 16 character key?

trupedia:
Reading helps ;)

If you wanna swap out the mcrypt extension entirely and you have the following setup:

16 character key => 'AES-128-CBC' cipher
32 character key => 'AES-256-CBC' cipher

for OpenSSL.

Did I catch it right?

PR #1564, opened 2015-06-12
Insert missing parentheses.

paulboco:


GrahamCampbell:
These are not missing, please close this.

paulboco:
If it is not intended for parentheses to be present, then why does the final PHP generated by this directive (shown on last line of this file) contain parentheses?

GrahamCampbell:
Because the $expression variable already has the parentheses in it...

paulboco:
Cool. Thanks. I guess I don't understand how this works. Perhaps the example could reflect that.

GrahamCampbell:
Looking at the regex under the hood might help you understand. :)

paulboco:
Ok. Will do.

PR #1566, opened 2015-06-13
Update upgrade.md

KevinMellott91:
Added migration instructions for the addition of the 'SoftDeletingTrait' alias, when upgrading from 4.1 to 4.2.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -384,6 +384,10 @@ If you are using soft deleting models, the `softDeletes` property has been remov
 		use SoftDeletingTrait;
 	}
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
:-1: This isn't correct and 4.2 isn't maintained anymore anyway.

taylorotwell:
Graham, what is correct?

GrahamCampbell:
You don't HAVE to add this alias to upgrade so that's why I'm saying this is wrong. :)

KevinMellott91:
Fair enough - thanks for the quick response

PR #1569, opened 2015-06-13
SSH user and configuration sub-section

anaxamaxan:
This seemed a little confusing, so I had to read the package source to understand.


taylorotwell:
Fixed.

PR #1573, opened 2015-06-13
Update installation.md

simondavies:
To prevent an error on running the application, you need to run sudo composer update, so that the system/app sets up this new app namespace.  This alway happens to me when i do this, so though might be best to add it to these docs.

After the renaming and i run say php artisan straight after i get `PHP Fatal error:  Uncaught exception 'ReflectionException' with message..'Class NAMESPACE\Console\Kernel does not exist...`error message.  Then running composer update its fixed.


GrahamCampbell:
:-1:

PR #1575, opened 2015-06-14
Adding Elixir upgrade instructions for 5.1.

orrd:
Users should also be told about changes they may need to make to their Elixir script when upgrading to 5.1.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -123,7 +123,7 @@ Add the protected `$baseUrl` property to the `tests/TestCase.php` file:
 
 ### Elixir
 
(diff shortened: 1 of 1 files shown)


orrd:
Actually, I don't know if the Elixir upgrade is entirely separate from Laravel 5.1, or if the changes are considered to be part of the 5.1 release.  (Laracasts included the changes in their series on 5.1, so I assumed it was.)  But the changes should be in some kind of release notes since it's a breaking change and people using this feature will need to update their gulpfile.js.

taylorotwell:
The changes are really separate.

PR #1582, opened 2015-06-15
Upgrading to 5.1.0 updates

webfractor:
Composer package requirement for Testing
Model Factories section
Update app/, config/bootstrap/ code differences section


GrahamCampbell:
:-1:

PR #1592, opened 2015-06-17
DB is on the root's namespace

luismilanese:
I just installed a clean version of Laravel, without any namespace tweaking on composer.json and this example does not work unless I put a backslash in front of DB in its use statement.

### DIFF
--- a/seeding.md
+++ b/seeding.md
@@ -24,7 +24,7 @@ As an example, let's modify the `DatabaseSeeder` class which is included with a 
 
 	<?php
 
-	use DB;
+	use \DB;
 	use Illuminate\Database\Seeder;
 	use Illuminate\Database\Eloquent\Model;
 


brunogaspar:
Does not work is a bit vague, what's happening? Any error being thrown? 
The example on the documentation seems to be correct.

GrahamCampbell:
This is wrong. :-1:

luismilanese:
@brunogaspar 
It throws an ErrorException. Upon the usage of the backslash in front of DB, it works properly.

```
[ErrorException]                                             
  The use statement with non-compound name 'DB' has no effect
```

@GrahamCampbell 
This is wrong. :-1: is a bit vague. What is wrong?

GrahamCampbell:
You don't need the import at all if you're not using a namespace...

GrahamCampbell:
The slash there is just hiding the error. It's still broken. Please close this.

luismilanese:
They don't need really. Actually, the ```use``` statement is optional in all cases, once you can always specify the entire namespace upon every call you make. I'll gladly close this and accept that my correction is really irrelevant, but as of now I don't quite get your point. :smile: 

You also say that it's hiding an error. What error is there, please? Broken? What is really broken?

GrahamCampbell:
Please close this and ask on the the forums. I'm telling you this change is wrong.

luismilanese:
I am closing this. Someone will try this and fail too. Being less arrogant can surely make things easier for everybody. :wink:

PR #1594, opened 2015-06-18
Update PSR2

antonybudianto:
consistent for the whole page


GrahamCampbell:
This change isn't actually PSR-2 just so you know. It is our code style though. :)

antonybudianto:
Okay, thanks for the info

PR #1605, opened 2015-06-20
Make docs use class name resolution

TheShiftExchange:
The default `ModelFactory` file in Laravel 5.1 uses class name resolution - but the docs do not.

We should be encouraging the same convention...


GrahamCampbell:
:-1:

TheShiftExchange:
why @GrahamCampbell ?

PR #1613, opened 2015-06-24
Update elixir.md

mstnorris:
Added clarity to the output location (and filename) when passing a second argument to the LESS method. The appropriate instructions should be added for SASS too.


taylorotwell:
Fixed.

PR #1616, opened 2015-06-25
Add Chinese docs link

artiship:
I have a proposal, why not support other languages for the documentation. For example, There is a lot of Chinese developers also trying to learn Laravel and use it in real projects, that could be a huge "market". I'm a Laravel fan from China, and I'm trying to translate the latest documentation to Chinese, But I just thought, Why not finally show the Chinese version direct on Laravel.com,


GrahamCampbell:
:-1:

artiship:
:)

PR #1617, opened 2015-06-25
Improve Elixir doc

TheShiftExchange:
Windows users (even when running Homestead) often need to include `--no-bin-links` when doing the install. This is because the symlink function does not work (even in Homestead).

Discussed here: https://github.com/npm/npm/issues/7308


antonybudianto:
:+1: 
Nice

PR #1622, opened 2015-06-25
Amended language on hashing.md

aewing:
Adjusted to use the same language as 5.1 docs.


taylorotwell:
Wrong branch. Should be 5.1.

aewing:
Actually I was trying to fix language in the 5.0 docs since they are still public-facing.

PR #1625, opened 2015-06-26
Added missing global helper functions

uxweb:
I have added some missing global helper functions for routing and miscellaneous. :)

### DIFF
--- a/helpers.md
+++ b/helpers.md
@@ -85,6 +85,16 @@ Laravel includes a variety of "helper" PHP functions. Many of these functions ar
 [url](#method-url)
 </div>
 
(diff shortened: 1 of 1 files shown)


helpers.md
```diff
@@ -85,6 +85,16 @@ Laravel includes a variety of "helper" PHP functions. Many of these functions ar
 [url](#method-url)
 </div>
 
+### Routes
```
GrahamCampbell: I think these are not documented on purpose.

uxweb:
Ok, should i remove them?, for a beginner it is not so obvious they exist.

PR #1627, opened 2015-06-28
Must explicitly specify 5.0

chrispecoraro:
If the installation is run like this, it will install version 5.1. I've updated this to match the format used in the 4.2 docs and it works.


GrahamCampbell:
:-1:

GrahamCampbell:
I don't think we should be documenting how to install old versions anyway.

PR #1628, opened 2015-06-28
Updated testing.md

devinfd:
Added "Adding Relations To Factory Definitions" blurb


taylorotwell:
This isn't really the definitive way to add relations... as we also show examples of how to add relations when actually building the models, which I think is a little more intuitive.

PR #1631, opened 2015-06-28
This can't be right but just wanted to suggest some

alnutile:
docs on this might help.

Had help from here https://laracasts.com/discuss/channels/general-discussion/changing-simple-pagination-nextprev-labels btw


taylorotwell:
We probably just need a simpler way to do this rather than documenting this complex way.

taylorotwell:
I made ->getPreviousButton and ->getNextButton public methods now. They accept a text parameter.

alnutile:
I was thinking the same thing. When it is that hard then it is not "Taylor Approved"

PR #1636, opened 2015-06-30
working command to generate the `Vagrantfile` and `Homestead.yaml` on per project installation

thehamsterofgod:
I followed exactly the steps in http://laravel.com/docs/5.1/homestead#per-project-installation and after running the command `php vendor/bin/homestead make` it did not create `Vagrantfile` and `Homestead.yaml` in my project root and instead it shows this on my cmd: 

capture

i ran this on cmd `php vendor/laravel/homestead/homestead make` and it exactly did what `php vendor/bin/homestead make` supposed to do: created `Vagrantfile` and `Homestead.yaml` with proper configurations inside

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -151,7 +151,8 @@ To install Homestead directly into your project, require it using Composer:
 
 Once Homestead has been installed, use the `make` command to generate the `Vagrantfile` and `Homestead.yaml` file in your project root. The `make` command will automatically configure the `sites` and `folders` directives in the `Homestead.yaml` file:
 
-	php vendor/bin/homestead make
+	php vendor/laravel/homestead/homestead make
+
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hmm, it might be a Windows thing. I will have to look into this more. The command in the docs is correct and how you would typically use a Composer bin though.

ArjanSchouten:
This is a Windows issue. If you've to use windows please use a linux box and do all the stuff from there.

ArjanSchouten:
@taylorotwell I had the same problem. Have a dual booted system so had the same issue on Windows. However I have the same issue on Ubuntu too. 
```
#!/usr/bin/env sh
SRC_DIR="`pwd`"
cd "`dirname "$0"`"
cd '../laravel/homestead'
BIN_TARGET="`pwd`/homestead"
cd "$SRC_DIR"
"$BIN_TARGET" "$@"
```
This file is placed in the bin directory. This isn't a php script. Leaving the ```php``` before the command fixes the problem. Haven't looked at this in detail so don't know what's exactly correct.

PR #1639, opened 2015-06-30
Update elixir.md - match required filename

jbruni:


jbruni:
See http://laravel.com/docs/master/elixir#writing-elixir-extensions

It says

> For example, if you place your extensions in elixir-extensions.js, you may require the file from your main Gulpfile like so:

But then in the following code it requires a file named "elixir-tasks" instead of "elixir-extensions".

PR #1641, opened 2015-07-01
Use correct VMware vagrant box name

nicekiwi:
While VMware Fusion is supported, the correct Vagrant box provider name appears to be `vmware_desktop`, not `vmware_fusion`.


taylorotwell:
vm_fusion works for me.

nicekiwi:
Actually, I can;t get either of them to work. (Under Windows).

PR #1644, opened 2015-07-01
Update authentication.md

lucasmichot:


GrahamCampbell:
I don't think we should include something non-offical in our docs.

lucasmichot:
We already do it

GrahamCampbell:
I think we shouldn't. :)

lucasmichot:
Maybe true yeah

PR #1646, opened 2015-07-02
Add nested validation example

vinkla:
I guess both the English and the example could need some improvements. Maybe with a better real world example. Just wanted to bring this up. 

@taylorotwell Something you would consider adding to the docs?


taylorotwell:
Added a note about this, but wasn't sure it deserved it's on section. Thanks.

vinkla:
Thanks.

PR #1652, opened 2015-07-03
Had trouble figuring out how to run commands like

alnutile:
`migrate:refresh --seed`

Thanks to @gerobk from larachat this should work?

### DIFF
--- a/artisan.md
+++ b/artisan.md
@@ -311,6 +311,12 @@ Sometimes you may wish to execute an Artisan command outside of the CLI. For exa
         //
     });
 
+Another example:
+
+        $exitCode = Artisan::call('migrate:refresh', ['--seed' => true]);
+        //
+    });
+
 Using the `queue` method on the `Artisan` facade, you may even queue Artisan commands so they are processed in the background by your [queue workers](/docs/{{version}}/queues):
 
     Route::get('/foo', function () {


ArjanSchouten:
Formatting isn't correct and it is already mentioned in the [docs](http://laravel.com/docs/5.0/migrations#database-seeding). Don't think this is the place to mention this.

alnutile:
@ArjanSchouten can you show me where it is in the docs? Your link is about a different subject. My focus is http://laravel.com/docs/5.1/artisan#calling-commands-via-code 
I can fix the formatting but I was stumped without the help of the chatroom to run this from inside a particular behat test.

ArjanSchouten:
Calling artisan from code is documented and all migration commands are described in the docs. If you scroll down you see it. Its the last part of that section.

alnutile:
@ArjanSchouten BUT it does not show how to call a command that is only `--seed` with no value.

Why not add a bit more to help the edge cases that can get a new or even "seasoned" user?

ArjanSchouten:
```--seed``` is just an option. If this option isn't set through the command line the option is false. If you pass it it's true. If you call it from code you've to say if you want it or not. So a bit straight forward how you can use this. If you look into commands more and write you're own you are still able to figure this out. Taylor writes very good documentation but can't mention every use case. He used queue as example and can't write every example.

Think Taylor have closed this for this reason. Otherwise he have closed it because the formatting.

alnutile:
Everything you said makes sense BUT not there (in the docs). You still have not helped me as a user going to the docs to figure this one feature out easily. I, being the user who did not know at that time "the option is false", did not find it "..straight forward how you can use this" else I would not be posting here :)

And yes "Taylor writes very good documentation but can't mention every use case" I agree on both. I just guess I feel that this one little fix would have saved me and others some time and is not as obvious as you say.

Thanks for the feedback though!

Cheers,
Al

ArjanSchouten:
We both think from different perspectives and I think both have pro's and con's. We let Taylor decide :)

PR #1654, opened 2015-07-03
Use the ::class notation

borisdamevin:


GrahamCampbell:
:-1:

borisdamevin:
The documation is not identical with the example code.

https://github.com/laravel/laravel/blob/master/database/seeds/DatabaseSeeder.php

ArjanSchouten:
The documentation must be readable. The existing part is more readable than this. I think that's the reason why Graham says this.

PR #1655, opened 2015-07-03
Constructor docblock should return self (PHPStorm warning)

pfuhrmann:


GrahamCampbell:
:-1:

GrahamCampbell:
PHPStorm is talking **** I'm afraid. `void` is the correct return type here.

pfuhrmann:
I agree that `void` is technically correct. However it MAY be omitted as per phpDoc guidelines which would be compromise and people using code examples in PHPstorm will be less confused and warning free.

> functions and methods without a `return` value, the @return tag MAY be omitted here, in which case @return void is implied
http://phpdoc.org/docs/latest/references/phpdoc/tags/return.html

GrahamCampbell:
PHPStorm need to fix this, not us.

taylorotwell:
It says that @return self is implied for constructors, which actually would
mean our stuff is wrong?

GrahamCampbell:
> It says that @return self is implied for constructors, which actually would mean our stuff is wrong?

Those docs are wrong. The current phpdoc standard doesn't say that. Those docs quoted there are from like 10 years ago.

pfuhrmann:
Can you please send link for current docs?

GrahamCampbell:
https://github.com/php-fig/fig-standards/pull/169

GrahamCampbell:
https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md

GrahamCampbell:
Also, it doesn't really matter what any of these standards say anyway, when the offical php documentation already defines what void means for us.

PR #1659, opened 2015-07-06
`$dateFormat` isn't limited to (builtin) timestamps.

nanaya1:
I believe `dateFormat` description should be part of date mutators documentation.


taylorotwell:
Added the note on the mutator page but left it on the original page as well.

PR #1662, opened 2015-07-06
Fix spelling in redis.md

johnberberich:
"Additinoally" -> "Additionally"
"...specify a of Predis client options" -> "...specify a set of Predis client options."


GrahamCampbell:
:+1:

PR #1667, opened 2015-07-09
Sync Accepts Boolean

alexhackney:
Adding a passage about the boolean on sync that could save some headaches.


taylorotwell:
Personally I don't find this very intuitive. If you just want to add records you should be using `attach`. Just my two cents.

alexhackney:
If you try to attach a record that already exists it throws an exception. Sync does not, being able to sync without getting all the current records saves a step.

alexhackney:
Another use case.
https://laracasts.com/discuss/channels/eloquent/belongstomany-and-wanted-duplicates

felipefrancisco:
This is very useful when handling old and new values in the same sync operation. This should be in the docs.

PR #1669, opened 2015-07-10
Added getPrefix function to be implemented.

sleimanx2:
Moreover the add() function is part of the documentation , but its not available in the contract and it's only implemented in MemcachedStore


cache.md
```diff
@@ -244,6 +244,7 @@ To create our custom cache driver, we first need to implement the `Illuminate\Co
         public function forever($key, $value) {}
         public function forget($key) {}
         public function flush() {}
+        public function getPrefix(){};
```
GrahamCampbell: cs

PR #1670, opened 2015-07-10
Added instructions to connect to Redis in Homestead

rafaelbeckel:


taylorotwell:
I think this will confuse people into thinking they need to do something out of the box to make the framework work with Redis.

rafaelbeckel:
Is there any suggestions to adapt it and not confuse people about supporting Redis out-of-the-box? Maybe if we tell upfront "if you do use \[Redis\]\(link_to_redis_doc_page\) ... "; or then listing Redis as a default port like others?

Current documentation confused me about the need of Port Forwarding, which is not actually necessary to connect to Redis in Homestead.

PR #1671, opened 2015-07-12
Documentation for "Touching Parent Timestamps" @eloquent.md

philippzentner:
Sorry, this pull request is done wrong (should've been my first one) and i'm not allowed to delete it.

How to "touch parent timestamps" has been deleted from the documentation in L5.1
The missing part is:

<a name="touching-parent-timestamps"></a>
## Touching Parent Timestamps

When a model `belongsTo` another model, such as a `Comment` which belongs to a `Post`, it is often helpful to update the parent's timestamp when the child model is updated. For example, when a `Comment` model is updated, you may want to automatically touch the `updated_at` timestamp of the owning `Post`. Eloquent makes it easy. Just add a `touches` property containing the names of the relationships to the child model:

	class Comment extends Model {

		protected $touches = ['post'];

		public function post()
		{
			return $this->belongsTo('App\Post');
		}

	}

Now, when you update a `Comment`, the owning `Post` will have its `updated_at` column updated:

	$comment = Comment::find(1);

	$comment->text = 'Edit to this comment!';

	$comment->save();


taylorotwell:
Added.

PR #1673, opened 2015-07-13
Update validation.md

Rkallenkoot:
Replaced the placeholder #foo with the actual name of the link: #other-validation-approaches

same as #1672


taylorotwell:
Fixed.

PR #1677, opened 2015-07-14
Add vagrant provider option for Homestead in VMWare Workstation on Windows

nicekiwi:
To get Homestead working in VMWare on Windows, the provider must be set to `vmware_workstation`. As fusion is exclusive to Mac and Workstation is exclusive to Windows. 

There is no VMWare Application called "Desktop".


nicekiwi:
<3

PR #1684, opened 2015-07-16
Update authentication.md

vdharshan:
Removed Bootstrap column classes to maintain consistency.


GrahamCampbell:
?

vdharshan:
@GrahamCampbell 
In Authentication page, the CSS class '<div class="col-md-6">' has been used only in 1 or 2 fields but other fields are using simply <div></div>. It's better to use empty <div> because it's more clear for the beginners.

PR #1685, opened 2015-07-16
Added semicolons after each Envoy command

JeroenNoten:
This is done because of this issue: https://github.com/laravel/framework/issues/8152
Line breaks disappear when a variable is placed at the end of the line, so a semicolon is necessary in those cases. For consistency, I added them everywhere.

Note that the examples are just not working at the moment, which is in my opinion very bad.

### DIFF
--- a/envoy.md
+++ b/envoy.md
@@ -38,7 +38,7 @@ All of your Envoy tasks should be defined in an `Envoy.blade.php` file in the ro
     @servers(['web' => 'user@192.168.1.1'])
 
     @task('foo', ['on' => 'web'])
-        ls -la
+        ls -la;
     @endtask
 
 As you can see, an array of `@servers` is defined at the top of the file, allowing you to reference these servers in the `on` option of your task declarations. Within your `@task` declarations, you should place the Bash code that will be run on your server when the task is executed.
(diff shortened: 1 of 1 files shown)


taylorotwell:
This isn't exactly necessary. If it is necessary that is a bug. There was a similar bug in Blade but it was fixed. Which version of Envoyer are you on?

JeroenNoten:
I am using Envoy 1.0.24.

JeroenNoten:
I think that it is indeed a bug, but its issue has been closed by GrahamCampbell a couple of minutes ago (see https://github.com/laravel/framework/issues/8152), with the proposal to use semicolons to fix it. So that is the reason that I introduced this update to the documentation.

GrahamCampbell:
Yeh, I closed it so that Taylor could respond here.

GrahamCampbell:
I can re-open that issue if needed. We just need to establish if Taylor thinks this is a bug since it's not for me to ultimately decide.

JeroenNoten:
Ok great! Hopefully this will be fixed soon. Let me know if I can help.

(BTW, you closed the issue it before I created this PR, but no offence.)

GrahamCampbell:
Yeh, I do that so we don't forget to close the issues so we can keep on top everything. :)

JeroenNoten:
Ok, I understand!

PR #1690, opened 2015-07-20
Missing parentheses in Blade::directive example

wecc:


GrahamCampbell:
This is wrong.

GrahamCampbell:
They are not missing.

wecc:
I looked at the source of `BladeCompiler` to understand why and I can see why I was mistaken.

PR #1693, opened 2015-07-21
Update blade.md

mmf:
updated blade.md - replaced curly braces with parenthesis for 'with' helper


mmf:
i checked BladeCompiler and noticed i was mistaken.

PR #1699, opened 2015-07-24
Add global scopes section back to eloquent

dulldave:


taylorotwell:
We are in the middle of exploring re-working this.

PR #1702, opened 2015-07-26
Fix links to non-existent "schema" page

jwcobb:
Instead, point to relevant section within "migrations"


jwcobb:
Missed one. Will re-submit.

PR #1706, opened 2015-07-28
Add missing asset helper

antonybudianto:


helpers.md
```diff
@@ -585,6 +587,20 @@ If the method accepts route parameters, you may pass them as the second argument
...
+#### `asset()` {#collection-method}
+
+Generate a URL for an asset.
+
+	$url = asset('img/photo.jpg');
+
+<a name="method-secure-asset"></a>
+#### `secure_asset()` {#collection-method}
+
+Generate a URL for an asset using HTTPS.
```
GrahamCampbell: It should be pointed out that `asset` will also generate an https link too if the current request scheme was https.

PR #1708, opened 2015-07-28
Update contracts.md

amosmos:
Added a missing class variable in the example

### DIFF
--- a/contracts.md
+++ b/contracts.md
@@ -78,6 +78,11 @@ Likewise, if we want to replace our underlying cache technology (Memcached) with
     class Repository
     {
         /**
+         * The cache.
+         */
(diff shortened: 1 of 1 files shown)


contracts.md
```diff
@@ -78,6 +78,11 @@ Likewise, if we want to replace our underlying cache technology (Memcached) with
     class Repository
     {
         /**
+         * The cache.
```
GrahamCampbell: `The cache instance.` might be better.

GrahamCampbell: You should also include a `@var` annotation.

amosmos: I just copied the exact same thing that was on the earlier example on the page to make it consistent.

PR #1711, opened 2015-07-30
Include Socialite facade for the AuthController

AndreasHeiberg:
A novice user might not catch that they have to include the Socialite facade in the AuthController if they are trying to use Socialite.


authentication.md
```diff
@@ -507,6 +507,7 @@ Next, you are ready to authenticate users! You will need two routes: one for red
     namespace App\Http\Controllers;
 
     use Illuminate\Routing\Controller;
+    use Socialite;
```
GrahamCampbell: Please use the FQCN.

PR #1724, opened 2015-08-06
Update queues.md

familymedia:
This caused me a bunch of grief. The example code is sqs specific, but it is very easy to overlook that since things that need to be changed appear to be marked in red. Except that's not how the docs really work. I don't care what the language is, but this piece of code should be clearer to show that sqs has been set as the service used when you may be using another service, like beanstalkd in my case.


familymedia:
This tutorial in general isn't that good since my problem was trying to use beanstalkd when I really wanted database. Once I figured out I can do queues without a service and just using database my problem were solved.

PR #1726, opened 2015-08-06
Change version method documentation

mercuryseries:
The version method no more accepts a file name relative to the public directory.

It seems that we need now to be explicit about the full path.


taylorotwell:
@JeffreyWay has informed me current documentation is correct.

GrahamCampbell:
Due to https://github.com/laravel/elixir/commit/b407d80c27be381b90e9df76957a0e435b3ba8a8.

PR #1730, opened 2015-08-07
Note about changes that won't work

7talents:
I stumbled across some migration changes that won't work (for example from "text" to "mediumText").


taylorotwell:
Going to hold off on including this for now. I like to try to avoid having dozens of little caveats in the docs if possible. Does Doctrine docs cover this?

PR #1731, opened 2015-08-08
Missing the subfolder "layouts/" when calling the view "child".

lguima:
*InvalidArgumentException in FileViewFinder.php line 137: View [child] not found.


taylorotwell:
I fixed the path.

PR #1737, opened 2015-08-11
Cache: sample names of tagged flush where switched

iNaD:
So, while reading the cache docs I found this error:

`Anne` is tagged as `['people', 'authors']` while `John` is tagged as `['people', 'artists']`.
So if you do `Cache::tags('authors')->flush();` then `Anne` will be deleted - not `John` as suggested by the docs.

Fixed this logical mistake.


iNaD:
Sorry wrong branch... Should be 5.1 not master?!

taylorotwell:
Yes please send to 5.1.

PR #1741, opened 2015-08-12
Update scheduling.md - Add class resolution syntax

ChristianPV:
Update to mirror laravel's current codebase.

### DIFF
--- a/scheduling.md
+++ b/scheduling.md
@@ -43,7 +43,7 @@ You may define all of your scheduled tasks in the `schedule` method of the `App\
          * @var array
          */
         protected $commands = [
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Gah. Why does this need changing?

ChristianPV:
I think it should reflect the current codebase so newcomer's don't get confuse on which one to use because the code is one way and the documentations is in another. I also think it's going to be easier for them to remember only one way instead of multiple ways to accomplish something.

I forgot about the import but if you think this should't be merge i won't add it. Thanks! :)

GrahamCampbell:
What is there to confuse? Both are valid.

PR #1747, opened 2015-08-15
Added clarification to Foreign Key Constraints

GiantCowFilms:
I though that by using $table->foreign, a column would be created. This caused a fair amount of confusion for me, and several other people  (I eventually figured out my problem after seeing a stack overflow post by someone who encounters the same issue). Its a minor change, but will make this clearer

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -321,12 +321,12 @@ Command  | Description
 <a name="foreign-key-constraints"></a>
 ### Foreign Key Constraints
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
This seems to be covered by the existing example, which shows creation and then constraining.

GiantCowFilms:
@taylorotwell  It wasn't very obvious, and I don't see the issue with the clarification. It is an easy thing to overlook, and the extra clarification helps. Thanks anyway, I really enjoy using your framework. :smiley:

PR #1751, opened 2015-08-15
Add/remove hyphens to improve readability

ssfinney:
Some hyphenated words in the docs are unnecessarily hyphenated, since they have unhyphenated English equivalents.

For example, "place-holder" may be replaced with "placeholder".

Other sets of words in the docs *should* be hyphenated. For example, "AJAX based" should be "AJAX-based".

### DIFF
--- a/artisan.md
+++ b/artisan.md
@@ -271,7 +271,7 @@ The `table` method makes it easy to correctly format multiple rows / columns of 
 
 #### Progress Bars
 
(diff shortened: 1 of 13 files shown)


GrahamCampbell:
:-1: Zero point in this. I really don't like this.

ssfinney:
It's *not* pointless, but I understand if you don't prefer this approach.

PR #1753, opened 2015-08-16
Added JSON validation.

bmitch:


taylorotwell:
I think we need an error message in the laravel/laravel repo too :heart:

bmitch:
:+1: Sounds good I'll get to that.

PR #1756, opened 2015-08-18
Docs for new every() collection method

bmitch:


collections.md
```diff
@@ -236,6 +236,23 @@ Return `false` from your callback to break out of the loop:
...
+
+    $collection = collect(['a', 'b', 'c', 'd', 'e', 'f']);
+
+    $collection->every(4);
+
+    // ['a', 'e']
+
+You may optionally pass offset as the second argument:
+
+    $collection->every(4,1 );
```
gmsantosxl: extra space here, it should be `$collection->every(4, 1);` to keep consistency with other doc parts

bmitch: :+1 Thanks. Fixed it up.

PR #1760, opened 2015-08-20
Update installation.md

jmrieger:
Suggested Microsoft IIS installation information; how to get pretty URL's to work


taylorotwell:
I would rather not add this to main documentation as it's hard to keep up to date. It would be nice to maybe do a blog post with it or something. Since I don't have access to IIS, it's hard for us to know if it's correct and test it as time goes on.

PR #1763, opened 2015-08-21
Avoid hardcoding HTTP status codes

edcs:
Most of us know what 404 and 500 HTTP responses are, but how about some of the other ones?

This PR attempts to show readers a useful feature that's available in `Illuminate\Http\Response`. Since it's built on top of `Symfony\Component\HttpFoundation\Response`, we can use all of the HTTP response constants available:

https://github.com/symfony/HttpFoundation/blob/master/Response.php

After all, `Response::HTTP_I_AM_A_TEAPOT` is much more descriptive than `418`.

### DIFF
--- a/errors.md
+++ b/errors.md
@@ -87,7 +87,7 @@ The `render` method is responsible for converting a given exception into an HTTP
     public function render($request, Exception $e)
     {
         if ($e instanceof CustomException) {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Simpler to just use the codes.

PR #1766, opened 2015-08-26
Mentioned dynamic where in query builder docs

jedrzej:
I find dynamic wheres a very convenient way of making code that uses query builder more readable and I think they should be at least mentioned in the docs.

I believe they deserve a separate section in the query builder docs, but on the other hand I see you try to keep the docs as simple as possible. If you like I could add more info about using dynamic where to the docs, just let me know


taylorotwell:
Dynamic where clauses are a little slower to use and are more of a novelty than something that is beneficial. Would prefer people just use explicit wheres.

PR #1769, opened 2015-08-28
Add more examples of how to listen for eloquent model events

fungku:
I had a tough time figuring out how to register a listener class for model events today.

I think it would be nice to add something like this to the docs.

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -524,38 +524,88 @@ Eloquent models fire several events, allowing you to hook into various points in
 
 Whenever a new model is saved for the first time, the `creating` and `created` events will fire. If a model already existed in the database and the `save` method is called, the `updating` / `updated` events will fire. However, in both cases, the `saving` / `saved` events will fire.
 
-For example, let's define an Eloquent event listener in a [service provider](/docs/{{version}}/providers). Within our event listener, we will call the `isValid` method on the given model, and return `false` if the model is not valid. Returning `false` from an Eloquent event listener will cancel the `save` / `update` operation:
+For example, lets look at a couple of ways to register a [listener](/docs/{{version}}/events#defining-listeners) for these model events in a [service provider](/docs/{{version}}/providers). In the event service provider we can either register the listener in the listeners array, or we can specify it directly on the model in the boot method:
 
     <?php
-
+    
     namespace App\Providers;
-
-    use App\User;
-    use Illuminate\Support\ServiceProvider;
-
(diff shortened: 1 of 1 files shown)


alexrussell:
You should make it clearer that the `$listens` array and direct call to `User::created` are two separate options (maybe with comments in the code or two separate code samples?). Someone just looking at the example code (without having properly read what you wrote above it) might think you have to do both and then get a little confused as to why.

eloquent.md
```diff
@@ -524,38 +524,88 @@ Eloquent models fire several events, allowing you to hook into various points in
 
 Whenever a new model is saved for the first time, the `creating` and `created` events will fire. If a model already existed in the database and the `save` method is called, the `updating` / `updated` events will fire. However, in both cases, the `saving` / `saved` events will fire.
 
-For example, let's define an Eloquent event listener in a [service provider](/docs/{{version}}/providers). Within our event listener, we will call the `isValid` method on the given model, and return `false` if the model is not valid. Returning `false` from an Eloquent event listener will cancel the `save` / `update` operation:
+For example, lets look at a couple of ways to register a [listener](/docs/{{version}}/events#defining-listeners) for these model events in a [service provider](/docs/{{version}}/providers). In the event service provider we can either register the listener in the listeners array, or we can specify it directly on the model in the boot method:
 
     <?php
-
+
```
GrahamCampbell: remove all cases where you've added whitespace on blank lines

eloquent.md
```diff
@@ -524,38 +524,88 @@ Eloquent models fire several events, allowing you to hook into various points in
...
+         * The event listener mappings for the application.
          *
+         * @var array
+         */
+        protected $listen = [
+            'eloquent.created: App\User' => [
+                'App\Listeners\UserWasCreated',
+            ],
+        ];
+
```
GrahamCampbell: here too

eloquent.md
```diff
@@ -524,38 +524,88 @@ Eloquent models fire several events, allowing you to hook into various points in
...
+            
+            \App\User::created(\App\Listeners\UserWasCreated::class);
         }
+    }
+
 
+Or let's define an Eloquent event listener right in the model. Within our event listener, we will call the `isValid` method on the given model, and return `false` if the model is not valid. Returning `false` from an Eloquent event listener will cancel the `save` / `update` operation:
+
+    <?php
+
```
GrahamCampbell: like 20 places

taylorotwell:
We should just show one example.

PR #1773, opened 2015-08-31
typo

bmitch:


bmitch:
Ah nevermind I see it logged in #1772 already :)

PR #1774, opened 2015-08-31
"linde" vs "line"

drbyte:


drbyte:
Doh. Duplicate of #1772

PR #1776, opened 2015-08-31
Further explain setting the Stripe secret key

drbyte:
The example shows that an environment setting is expected. This simply adds clarification by showing the setting in the `.env` file.

### DIFF
--- a/billing.md
+++ b/billing.md
@@ -60,12 +60,16 @@ Adding the columns to your model's `$dates` property will instruct Eloquent to r
 
 #### Stripe Key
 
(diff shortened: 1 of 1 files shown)


billing.md
```diff
@@ -60,13 +60,17 @@ Adding the columns to your model's `$dates` property will instruct Eloquent to r
 
 #### Stripe Key
 
-Finally, set your Stripe key in your `services.php` configuration file:
+Finally, in your `services.php` configuration file set it to read the Stripe key from your site's `.env` file:
```
GrahamCampbell: doesn't necessarily have to be the .env file

PR #1777, opened 2015-08-31
Add tip for installing ACL dependencies before upgrading

drbyte:
Silly me, I made the basic mistake of not running `composer update` **first** before adding the ACL AuthController code to one of my apps, and had to undo it all. This just helps others not make the same basic mistake by clarifying the order in which steps should be done.


GrahamCampbell:
:-1:

GrahamCampbell:
It's obvious you need to run composer update, otherwise you've not updated to 5.1.11.

PR #1778, opened 2015-08-31
Typo in "line"

sdebacker:


sdebacker:
There is already a PR for that.

PR #1786, opened 2015-09-03
Add note that flush() clears everything

mfn:
Especially irrespective of the prefix.

I just nuked our production redis database (used for Resque)
because of this behavior (I'm calling `php artisan clear:cache`
on every deploy).

See https://github.com/laravel/framework/issues/5034 for "prior art".

### DIFF
--- a/cache.md
+++ b/cache.md
@@ -189,6 +189,12 @@ You may remove items from the cache using the `forget` method on the `Cache` fac
 
     Cache::forget('key');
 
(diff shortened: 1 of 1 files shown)


trompx:
Would it be possible to prompt the user confirmation with a warning message that the action is going to flush the entire database and not only the laravel session keys ?

I just flushed my whole redis database due to that behavior.

PR #1787, opened 2015-09-04
Adding note about misleading return

adamgoose:


taylorotwell:
I like to try to avoid a lot of little notes like this. Is there anything we can to make the examples itself more clear?

adamgoose:
Well, other than *not* having a return in the example, I can't think of anything. The official gulp examples all return the promise - not sure why. Maybe something like this?

```javascript
 gulp.task('speak', function() {
     var message = 'Tea...Earl Grey...Hot';
 
     // Do not return
     gulp.src('').pipe(shell('say ' + message));
 });

```

PR #1788, opened 2015-09-06
Added backslash before route namespace

raffw7912:
Without backslash Laravel will append namespace to main application namespace and will return an error "Class App\Http\Controllers\Laravel\Cashier\WebhookController does not exist".

### DIFF
--- a/billing.md
+++ b/billing.md
@@ -239,7 +239,7 @@ If the user cancels a subscription and then resumes that subscription before the
 
 What if a customer's credit card expires? No worries - Cashier includes a Webhook controller that can easily cancel the customer's subscription for you. Just point a route to the controller:
 
-    Route::post('stripe/webhook', 'Laravel\Cashier\WebhookController@handleWebhook');
+    Route::post('stripe/webhook', '\Laravel\Cashier\WebhookController@handleWebhook');
 
 That's it! Failed payments will be captured and handled by the controller. The controller will cancel the customer's subscription when Stripe determines the subscription has failed (normally after three failed payment attempts). Don't forget: you will need to configure the webhook URI in your Stripe control panel settings.
 


GrahamCampbell:
No it doesn't. That's not true at all. That will ONLY happen if you wrap things in a namespace group, and the example here is not wrapped in such a group.

raffw7912:
I might be wrong here but I don't have webhook route wrapped in Route::group() so your comment seems to be not true at all to me. All I have at the end of my routes.php is (no group around it):

```php
Route::post('stripe/webhook', '\Laravel\Cashier\WebhookController@handleWebhook');
```

and it works. When I remove the backslash and I try this route via Postman (POST) it gives me an error.

GrahamCampbell:
> All I have at the end of my routes.php is (no group around it):

No, it does. Laravel wraps it in a group when it loads it.

GrahamCampbell:
See: https://github.com/laravel/framework/blob/5.1/src/Illuminate/Foundation/Support/Providers/RouteServiceProvider.php#L85.

GrahamCampbell:
> All I have at the end of my routes.php is (no group around it):

So there is a group. ;)

GrahamCampbell:
Laravel doesn't magically guess the namespace you see. ;)

raffw7912:
Ok, I don't want to argue here, but to me if we want to have the best framework documentation ever, then we have to add notes in the docs. This is not explained and people might just simply copy&paste Taylor's example and it will not work. Majority of people won't go into Postman to test this route, they will rely on Taylor's example. Thanks anyway, but I don't understand why are we doing things like this, ie. "example in docs is not wrapped in group so there's no backslash" but obviously it will be wrapped up in a group automatically! Weird approach.

raffw7912:
I'm also not "magically guessing" why docs example doesn't work. Your comment is really unnecessary here.

GrahamCampbell:
> I'm also not "magically guessing" why docs example doesn't work

I wasn't saying that all. I'm saying Laravel can't magically guess the namespace.

GrahamCampbell:
> but obviously it will be wrapped up in a group automatically! Weird approach.

It's only wrapped in a group because the user does that in their route service provider. That's no secret.

GrahamCampbell:
That said, you are correct in saying that copying and pasting this example will often result in error due to oversight of users, so that's probably good enough reason to change it tbh, so thanks for bringing this up.

PR #1799, opened 2015-09-15
Overriding the job queueing mechanism of dispatcher

halaei:
Related to issue: [DispatchesJobs cannot queue to a different connection in Laravel 5.1](https://github.com/laravel/framework/issues/9896) on laravel/framework

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -216,6 +216,23 @@ By pushing jobs to different queues, you may "categorize" your queued jobs, and 
         }
     }
 
+#### Overriding the job queuing mechanism
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
To me this isn't the most expressive way to do this. I would just do Queue::connection('foo')->push(new Job) from outside of the job itself.

halaei:
That make sense. I don't like it either! I have a proposal [in this comment](https://github.com/laravel/framework/issues/9896#issuecomment-140417682). Would be nice if you share your opinion.

PR #1804, opened 2015-09-17
Add documentation for the `method_field()` helper

JeroenNoten:
See https://github.com/laravel/framework/pull/9477

Explanation is similar to the documentation of `csrf_field()`, but a little more compact, as there is already an HTML example.

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -326,6 +326,14 @@ HTML forms do not support `PUT`, `PATCH` or `DELETE` actions. So, when defining 
         <input type="hidden" name="_token" value="{{ csrf_token() }}">
     </form>
 
+To generate the hidden input field `_method`, you may also use the `method_field` helper function:
+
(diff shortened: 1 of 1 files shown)


routing.md
```diff
@@ -326,6 +326,14 @@ HTML forms do not support `PUT`, `PATCH` or `DELETE` actions. So, when defining
         <input type="hidden" name="_token" value="{{ csrf_token() }}">
     </form>
 
+To generate the hidden input field `_method`, you may also use the `method_field` helper function:
+
+    <?php echo method_field('PUT'); ?>
```
GrahamCampbell: :-1:

JeroenNoten: I did this, because it is also explained this way for the `csrf_field()`.

GrahamCampbell: hmmm

PR #1808, opened 2015-09-22
Removed extra equals sign

danb-humaan:
Double equals would force an evaluation for a value that doesn't exist in the attributes array.


GrahamCampbell:
This is wrong.

GrahamCampbell:
The `==` compares that value to yes and returns true/false.

PR #1809, opened 2015-09-22
Change "vagrant provision" to "homestead provision"

kayladnls:
`vagrant provision` didn't work for me, after updating `Homestead.yaml`. 

`homestead provision`, however, did the trick perfectly.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -186,7 +186,7 @@ To connect to your MySQL or Postgres database from your host machine via Navicat
 <a name="adding-additional-sites"></a>
 ### Adding Additional Sites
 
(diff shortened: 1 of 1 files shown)


crynobone:
Never experience such issue, unless you started off using `homestead up` instead of `vagrant up` during the first provisioning. 

It seem to actually run the `vagrant up` inside your `laravel/homestead` directory (as far I understand) <https://github.com/laravel/homestead/blob/master/src/UpCommand.php#L41>

kayladnls:
Ahh, yes, I did use homestead up. Why aren't those commands/features in the docs?

taylorotwell:
Yeah current docs show the usage of just the plain Vagrant commands. We sort of moved away from the "homestead" commands because it was a little too magical and was confusing. Just sticking with the plain Vagrant commands from the actual Homestead directory is a little more easier for people to grasp.

PR #1810, opened 2015-09-22
Update Model::with() for multiple args

kryogenic:
now takes an array instead of varargs


GrahamCampbell:
Both are valid arn't they?

kryogenic:
When I tried the version in the docs I got very strange behavior for the second eager loaded model: http://laravel.io/bin/9vnxD

kryogenic:
The idea here is that Model::with() behaves differently depending on if you pass it an array or multiple arguments. Please look into this, thanks.

PR #1821, opened 2015-10-03
Added BrowserSync section to Elixir

AndreasHeiberg:


AndreasHeiberg:
I thought we might want to clarify that .version() and .browserSync() are incompatible as per my pull-request https://github.com/laravel/framework/pull/10402.

I didn't know if you wanted to specify this in the docs, so I left it out for the first rev to make merging easier.

What's your thoughts?

PR #1827, opened 2015-10-05
Update homestead.md - Clarify Section Upgrading To PHP 7.0

slavo2:
- using different directory name as an example for new clone
 - adding condition of already having `~/.homestead/Homestead.yml` (not true for older version of Homestead, from which one might also want to upgrade, and not be aware of the difference.
 - clarifying from which directory to run `vagrant up` from


GrahamCampbell:
Please get rid of the merge commits.

GrahamCampbell:
And the other unrelated changes...

PR #1829, opened 2015-10-06
Clarify how `mimes` validation rule maps extensions to mimetypes

jordanlev:


taylorotwell:
This is not the correct location that Mimes are pulled from.

jordanlev:
ok... where are they pulled from then? This confused me for several hours today (trying to figure out why mimetypes were being validated against file extensions), so I feel like at least some explanation in the docs might help out other newbies. Happy to explain it more accurately if you can point me to where these are coming from.

Thanks!

PR #1832, opened 2015-10-08
Restore useful content eliminated in 5.1

mmic-bjohnson:
In 5.1, the `URLs To Controller Actions` section of this page was eliminated, and the explanation of `action()` was jammed into the `Naming Controller Routes` section. This was a step backwards in that the previous section name was more appropriate, and two important methods were scrubbed from the page.

Mention of the `URL::setRootControllerNamespace()` method was eliminated. This is a hugely useful method that is probably under-utilized because it is little-known. My feeling is that it has a place on this page.

Similarly, the `Route::currentRouteAction()` method was eliminated. There is no clear explanation as to why (at least that I'm able to find).

This change restores the `URLs to Controller Actions` section, verbatim, as it appeared in 5.0.

### DIFF
--- a/controllers.md
+++ b/controllers.md
@@ -63,6 +63,8 @@ Like Closure routes, you may specify names on controller routes:
 
     Route::get('foo', ['uses' => 'FooController@method', 'as' => 'name']);
 
+#### URLs To Controller Actions
+
 Once you have assigned a name to the controller route, you can easily generate URLs to the action. To generate a URL to a controller action, use the `action` helper method. Again, we only need to specify the part of the controller class name that comes after the base `App\Http\Controllers` namespace:
 
     $url = action('FooController@method');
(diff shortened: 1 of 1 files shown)


taylorotwell:
The URL::setRootControllerNamespace is called automatically for you in 5.1.

mmic-bjohnson:
@taylorotwell Thank you for getting to this so quickly!

If you have a moment, do you mind elaborating a bit as to how this automatic behavior functions in 5.1?

I ask because I'm on Laravel v5.1.19, and when I comment-out the following line in my package, which had initially allowed this to function the way I required, e.g.,

```php
URL::setRootControllerNamespace('MyVendor\MyProject\Http\Controllers');
```

I receive the following exception:

```
InvalidArgumentException in UrlGenerator.php line 558: Action App\Http\Controllers\MyController@myMethod not defined.
```

Surely, there is something more to it... a brief explanation or a link to the relevant documentation would be most appreciated.

Thanks again, Taylor.

taylorotwell:
Check out the base RouteServiceProvider in the Illuminate\Foundation namespace.

PR #1843, opened 2015-10-16
The "pretend" option for mail isn't in the docs

jtgrimes:
Let's fix that!

### DIFF
--- a/mail.md
+++ b/mail.md
@@ -227,6 +227,10 @@ Another solution provided by Laravel is to set a universal recipient of all e-ma
         'name' => 'Dev Example'
     ],
 
+#### Pretend
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think we removed it because it is deprecated since we now have the "log" mail driver which you can set as your driver in your config/mail.php. The "log" driver basically servers the same purpose and in fact gives even more info such as headers, etc. Let me know if you have any questions :)

jtgrimes:
Pretend is still in the code and in the config file - should it be pulled out of there? If not now, maybe in the next major release. I feel like if it's there, it should be documented, but it's probably not a big deal - I just happened to be playing in the mail docs for something else and noticed it.

taylorotwell:
Yes I think it’s marked as deprecated and would be removed in 5.2 which is released next month.

PR #1850, opened 2015-10-21
Undocumented default option value definition

rkgrep:


taylorotwell:
I don't know that this is ever really necessary.

PR #1899, opened 2015-11-05
Changes to Retrieving Single Models / Aggregates to better understand…

juliobitencourt:
…ing of the exception handler.

I thought that the 404 response was automatically sent with no extra configuration.


taylorotwell:
It is automatically done. You may have an early version of the 5.1 app structure, but it is automatically handled now.

PR #1901, opened 2015-11-06
Document prepend with key

JosephSilber:
Documentation for https://github.com/laravel/framework/pull/10811


taylorotwell:
It may be pretty helpful to show people the use case you mentioned int he original PR for select boxes.

JosephSilber:
@taylorotwell The form builder does not ship with laravel, so I'm not sure we can use it in an example here.

The other option is to manually spin through the options in blade to create `option` tags manually, but that seems a little overkill for this example.

Thoughts?

PR #1903, opened 2015-11-09
multiple databases w/ Eloquent

gnarlycharlie4:
A small snippet of code that specifies how to use multiple database connections with Eloquent.


taylorotwell:
Added to Eloquent docs.

PR #1904, opened 2015-11-11
Include examples of put/delete methods for implicit controllers

susanBuck:
I had a student ask whether Implicit controllers support put/delete in addition to get/post. This doc is *somewhat* vague about this, saying `The method names should begin with the HTTP verb `. You could infer from this that all 4 HTTP verbs are supported, however, in the examples we see only show get/post, so you might second guess this inference. 

In this proposed file change, I'm suggesting adding examples of put/delete. 

Alternatively, you could not add these examples, and instead make the instructions more explicit saying something like `all 4 HTTP verbs (get, post, put, delete) are supported.`

### DIFF
--- a/controllers.md
+++ b/controllers.md
@@ -237,6 +237,23 @@ Laravel allows you to easily define a single route to handle every action in a c
         {
             //
         }
+        
+        /**
+         * Responds to requests to PUT /users/profile
+         */
+        public function putProfile()
+        {
+            //
+        }
+        
+        /**
+         * Responds to requests to DELETE /users/profile
+         */
+        public function deleteProfile()
+        {
+            //
+        }
+        
     }
 
 As you can see in the example above, `index` methods will respond to the root URI handled by the controller, which, in this case, is `users`.


lagbox:
There are more than 4 http verbs.

crynobone:
If Laravel decide to deprecate/remove `Route::controller()` (they are considering it at the moment) I don't think this changes need to be added.

lagbox:
@crynobone good to hear

susanBuck:
@lagbox which am I missing? Patch?

susanBuck:
@lagbox I should clarify, I'm thinking specifically here of the http methods we see mentioned elsewhere in the docs (RESTful controllers, form spoofing, etc.).

susanBuck:
@crynobone If they remove it, then this whole section will be removed. In the meantime, though, it'd still be an accurate change.  I guess this hinges on *when* they plan on removing it... tomorrow, next week, next month, etc. : )

crynobone:
As far as things goes right now is that it going to be deprecated in 5.2 (release this month-ish), and based on how deprecation work in Laravel it going to be removed in 5.3 (release next June).

If this is the case, advertising something that would be missing in less than a year shouldn't happen.

susanBuck:
@crynobone This is not suggesting an advertisement be added, just that the existing advertisement (which as of this writing is applicable) be clearer. 

Was just a question that was raised by a Laravel student who is reading the documentation today (not next month, or next June) and was unclear on how the codebase (today) worked.

PR #1905, opened 2015-11-11
Update elixir.md

acacha:
I removed

```javascript
elixir.config.js.browserify.transformers.push({
    name: 'vueify',
    options: {}
});
```

Sorry because I'm not 100% sure of this change! I think elixir config is not necessary at least at @JeffreyWay https://laracasts.com/series/learning-vuejs/episodes/14 no config is done and it seems to work.

Because no issues section is able at this repo I make this PR but take this more as a question than a PR ;-)

### DIFF
--- a/elixir.md
+++ b/elixir.md
@@ -213,10 +213,7 @@ While Browserify ships with the Partialify and Babelify transformers, you're fre
     npm install vueify --save-dev
 
 ```javascript
-elixir.config.js.browserify.transformers.push({
-    name: 'vueify',
-    options: {}
-});
+
 
 elixir(function(mix) {
     mix.browserify('main.js');


acacha:
It seems now Vueify is now a default Browserify transformer:

https://github.com/laravel/elixir/commit/58c43723f6eab8bc920fce48e498640d481e2aca

JeffreyWay:
It's true that Vueify is now a default transformer with Elixir, but this example is still useful for demonstrating how to plugin in your own transformers. So we shouldn't remove it entirely.

Maybe you can change "vueify" to a different transformer, like "aliasify".

taylorotwell:
Updated per @JeffreyWay suggestion.

PR #1906, opened 2015-11-13
Add a note for JSON fields

rkgrep:
Since `json` is used by default, it will break on earlier versions of MySQL (previously there was a fallback to text).


taylorotwell:
No need for this caveat here. Would go in 5.2 release notes.

PR #1908, opened 2015-11-14
send mail example doesn't work without from set

samthomson:
the mail sending example does not work, a message seems to need the from sender set in order to work. If this is necessary and not optional it should be in the documentation.


taylorotwell:
Wrong branch. I've made the fix on 5.1 branch. Thanks.

PR #1916, opened 2015-11-18
Added array_build() helper

anteriovieira:


helpers.md
```diff
@@ -139,6 +140,17 @@ The `array_add` function adds a given key / value pair to the array if the given
 
     // ['name' => 'Desk', 'price' => 100]
     
+<a name="method-array-build"></a>
+#### `array_build()` {#collection-method}
+
+The `array_build` function build a new array using a callback:
+
+    $array = array_build([1, 2, 3, 4, 5], function($key, $value){
```
GrahamCampbell: missing space before the brace

taylorotwell:
I don't totally understand how this is different from map and doesn't seem to be used internally anywhere.

anteriovieira:
I agree with you, would be the case of depreciating?

vlakoff:
Added in https://github.com/laravel/framework/commit/8683a510b3e5541cb57d1ef6a09d66b713e42a13.

vlakoff:
Apparently the use in [Route::parseFilters](https://github.com/laravel/framework/blob/2375c23596264bf6ac838a7ef2842f4b7c61d003/src/Illuminate/Routing/Route.php#L315) has been the only use in framework, ever.

PR #1919, opened 2015-11-19
Correct property to perform redirect.

DavidStrada:
fixed incorrect property `protected $redirectTo` .


taylorotwell:
actually both work.

DavidStrada:
mmm .. if i use `protected $redirectPath = '/task';` it goes to example.com/home 
however if i use `protected $redirectPath = '/task';` it does what's supposed to do example.com/task.

PR #1922, opened 2015-11-21
Fix method explanation

RDelorier:


taylorotwell:
Should say "or list of values"?

PR #1929, opened 2015-11-25
Use ::class notation in facades section

anik786:
Use class notation instead of string so that it is consistent with the rest of the framework following this commit: laravel/laravel@58db25b and best practices.


taylorotwell:
Incorrect.

anik786:
Sorry i realized after. was meant to delete this branch.

PR #1937, opened 2015-12-02
Statement in wrong place for where in join query

dlepaux:


taylorotwell:
The original code is correct. You can do `where` clauses within join clauses.

dlepaux:
Yes but when i use a `where` insite the join clause, i got an exception..
I found my solution here : http://laravel.io/forum/07-30-2014-missing-where-method-on-join-with-where-clause
I'm sorry if i don't understand...

PR #1938, opened 2015-12-02
Eloquent::save expects an array

aydot:
Eloquent save expects an array, factory() returns an object.


taylorotwell:
Incorrect.

aydot:
Here is my factory 
```
$factory->define(App\Models\User::class, function () {
    return [
        'username' => str_random(10),
        'password' => bcrypt(str_random(10)),
    ];
});
```
calling 
```
 $user = factory(User::class)->make();
```
returns an object

taylorotwell:
Yes, and ->save() accepts an object.

PR #1945, opened 2015-12-05
Added Mcrypt to requirements

vjandrea:


GrahamCampbell:
:-1:

GrahamCampbell:
This is incorrect. It's no longer required in 5.1. Infact, the only thing that uses it is deprecated.

vjandrea:
@GrahamCampbell i just deployed a 5.1.26 on a host that doesn't support mcrypt and i see plenty of exceptions like this:
“**RuntimeException** in *EncryptionServiceProvider.php* line 29: No supported encrypter found. The cipher and / or key length are invalid.”
Do you have any suggestion, apart from enabling Mcrypt?

GrahamCampbell:
> Do you have any suggestion, apart from enabling Mcrypt?

Yes, correct you encryption config.

GrahamCampbell:
> apart from enabling Mcrypt?

That is only there for really old Laravel 4 apps to still work. We're removing it at some point.

vjandrea:
Thanks for the info, i had lost this commit: https://github.com/laravel/laravel/commit/f0cad014cf5b8db69adc93854a25a5792b8e3a7e, actually started the project earlier than june.

GrahamCampbell:
Just a note that we've finally removed the mcrypt gallback in L5.3, which is due to ship in June 2016.

PR #1949, opened 2015-12-07
Added a 'See also' line to the random method

cbtech:
Added 'See also' shuffle to random method.

This was due to me wanting the shuffle method but searched for random. A reference to also look at the shuffle method would have helped
This is following the style on the php.net documentation.

If this style is accepted I'd be happy to apply this to the rest of the documentation.


taylorotwell:
Going to hold off on this.

PR #1951, opened 2015-12-10
Object $response doesn't have a status() method

llnk:


taylorotwell:
It seems to have one: https://github.com/laravel/framework/blob/5.1/src/Illuminate/Http/ResponseTrait.php

llnk:
Are you sure? Whenever I try to run phpunit, I get:

```
Fatal error: Call to undefined method Symfony\Component\HttpFoundation\Response::status()
in ...api\tests\test.php on line 38
```

llnk:
Proof of concept: put this simple test in your folder and run phpunit to see the error:
```
<?php

class ExampleTest extends TestCase
{
    /**
     * A basic functional test example.
     *
     * @return void
     */
    public function testBasicExample()
    {
       $response = $this->call('GET', '/');

        $this->assertEquals(200, $response->status());
    }
}
```

PR #1953, opened 2015-12-11
Update homestead.md

redgluten:
Add a note to signal NFS sharing incompatibility with with Windows hosts or encrypted disks.


redgluten:
Did you close because it was redundant with Vagrant’s documentation? Maybe adding a word to check vagrant’s doc would be useful then. I spent several hours trying to make this work before stumbling on it (my disk is encrypted). I may be not be the only one.

PR #1962, opened 2015-12-18
Update upgrade.md

mikield:
The auth.php config file links to master (the old one).
I think better to link the develop link, and when last config file will be in master - link it in documentation.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -29,7 +29,7 @@ Add `symfony/dom-crawler ~3.0` and `symfony/css-selector ~3.0` to the `require-d
 
 #### Configuration File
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
:-1:

GrahamCampbell:
It will be master when 5.2 is released.

taylorotwell:
@GrahamCampbell why the **** would we leave the wrong file in there **now**?

mikield:
@GrahamCampbell if we have updating manual to 5.2 beta1 in docs, we need to give users right info.
I am thouse person, who made it like said in docs, and had error in auth. I needed to figure out, what I have done wrong? But there was nothing wrong from my side, It was bad link to file update.
So my opinion is to get the info 100% right if I`m saying (in laravel`s case displaing) it.

Sorry for my bad english :)

PR #1969, opened 2015-12-21
[5.2] Remove where clause caveats

acasar:
This has been fixed in 5.2

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -435,24 +435,6 @@ The `withTrashed` method may also be used on a [relationship](/docs/{{version}}/
 
     $flight->history()->withTrashed()->get();
 
-#### Where Clause Caveats
-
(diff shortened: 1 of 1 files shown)


mustafaaloko:
How has it been fixed? I don't think it was a bug from day one.

mustafaaloko:
I mean, how it will handle this situation? 

    User::where('name', '=', 'John')->orWhere('votes', '>', 100);

This will produce the following SQL (if a soft deleted is already used):

    select * from `users` where `users`.`deleted_at` is not null AND `name` = 'John' OR `votes` > 100

If the final where is clause is evaluated to **true** then it will ignore the **`users`.`deleted_at` is not null**, right?

mustafaaloko:
:+1:

PR #1973, opened 2015-12-21
Update eloquent.md

olimorris:
Using `OldScope` implies a historical class that has since been superceded. With prior knowledge of Global Scopes in Laravel, I read this as "this is how we did it before in the OldScope". AgeScope is more apt at describing what the the scope does - apply a constraint to the age.


GrahamCampbell:
Wrong branch. Please send to 5.2.

PR #1975, opened 2015-12-21
Fix Eloquent Global Scope example

mul14:


GrahamCampbell:
Every change here is incorrect. Please close this.

mul14:
Whoops... I still use beta version.

PR #1976, opened 2015-12-21
Updating install docs to reflect that Laravel may not run if directories are not writable.

brandonsavage:


GrahamCampbell:
:-1: I think that's already implied by the docs, otherwise why even bother documenting it, right?

brandonsavage:
Laravel fails silently if these are not writable. It seems worth giving folks a place to look.

GrahamCampbell:
I'm still :-1:. The docs don't say this is optional, rather, laravel will not run.

PR #1986, opened 2015-12-23
Dispatchers section for 5.2 upgrate guide (Commands And Handlers)

halaei:


upgrade.md
```diff
@@ -65,6 +65,10 @@ The `Illuminate\Foundation\Support\Composer` class has been moved to `Illuminate
 
 ### Commands And Handlers
 
+#### Dispatchers
+
+The `dispatchFrom` and `dispatchFromArray` methods are removed from `Illuminate\Foundation\Bus\DispatchesJobs` trait and `'Illuminate\Contracts\Bus\Dispatcher` class. Call `dispatch` instead.
```
GrahamCampbell: That last suggestion is incorrect.

taylorotwell:
See Graham's note. The suggestion is actually incorrect.

halaei:
Oh, I did not mean that `dispatch` work as exactly as the other 2 functions, if that was derived from the last suggestion. Sorry for my bad command of written English. Maybe someone else adds a more appropriate note later.

Thanks in advance.

PR #1989, opened 2015-12-23
[5.2] default value in old helper

vinicius73:
Reports that there is a second parameter in the old helper, the parameter $default


GrahamCampbell:
Please only send one PR.

vinicius73:
Sorry.
As the two documentations do not have this information figured it needed two PRs.
I do not know how to fix in two places with just one PR.

Can you help me?

PR #2009, opened 2015-12-28
Add missing EncryptCookies.php file

antonioribeiro:


GrahamCampbell:
That's not needed to upgrade, so :-1: from me. That's totally optional.

antonioribeiro:
It wasn't for me. :(

antonioribeiro:
Got it.

GrahamCampbell:
> It wasn't for me. :(

Could you describe what went wrong please? I could be wrong here. ;)

antonioribeiro:
In order to do a "real" upgrade from a 5.0 (5.1 ish) app, I needed those middleware groups, so I copied the Kernel.php from 5.2, which had that EncryptCookies class for the web group, which forced me to copy it from laravel/laravel, which made me think something was missing in the docs, but it wasn't, not for those doing a lazy upgrades.

I know I may have not upgraded it the way I should (I never do things the way people do :), but a simple comment on that line ("// This is totally optional.") would've helped.

PR #2014, opened 2015-12-29
Update how to create-project for 5.1

tflight:
If you follow the Composer Create-Project install instructions on the 5.1 page you will currently get 5.2. Now that 5.2 has been released if you want to install 5.1 LTS the version needs to be specified. (Version in quotes.)

### DIFF
--- a/installation.md
+++ b/installation.md
@@ -45,7 +45,7 @@ Once installed, the simple `laravel new` command will create a fresh Laravel ins
 
 Alternatively, you may also install Laravel by issuing the Composer `create-project` command in your terminal:
 
-    composer create-project laravel/laravel --prefer-dist blog
+    composer create-project laravel/laravel blog "5.1.*"
 
 <a name="configuration"></a>
 ## Configuration


tflight:
I don't know quite enough about `--prefer-dist` to know if it should still be included or not.

GrahamCampbell:
The prefer-dist isn't needed because we include it in the laravel/laravel composer.json as a default.

GrahamCampbell:
That must have been left over in the docs from before we did that. I can't remember off the top of my head when the change was made.

Miniwe:
Alternatively you can create 5.1 project as
➜  composer create-project laravel/laravel=5.1 <project_name>

tflight:
Yes, there are several variations that seem to work. Additionally, Taylor needed quotes around the version number for it to work for him, but it worked fine for me with and without.

`composer create-project laravel/laravel=5.1.* blog`
`composer create-project laravel/laravel:5.1.* blog`
`composer create-project laravel/laravel blog 5.1.*`

GrahamCampbell:
➜ composer create-project laravel/laravel=5.1

Will that not give 5.1.0 instead og 5.1.x?

tflight:
Correct, @GrahamCampbell, that will give 5.1.0. I think what is in my PR still works best for those reading the 5.1 docs:
 `composer create-project laravel/laravel blog "5.1.*"`

PR #2015, opened 2015-12-29
Fix Eloquent boot() method visibility

Propaganistas:


taylorotwell:
It's protected though?

Propaganistas:
That's odd; I got a method visibility error when simply pasting the docs
example.

PR #2020, opened 2016-01-02
[5.2] Adjust Validation docs to clarify view $errors

lagbox:
Noted needed middleware


lagbox:
I suppose we should update the Note in the beginner and intermediate tasks that relates to the $errors var as well ?

PR #2023, opened 2016-01-03
Update info on Blackfire

TheShiftExchange:
As per @GrahamCampbell in https://github.com/laravel/framework/issues/11514 - Blackfire is not available for PHP7 Homestead?


taylorotwell:
Just removed it entirely for now until they are ready for PHP 7.

PR #2024, opened 2016-01-04
Add docs for related model factory

vjrngn:
I have added a little bit of documentation explaining how to build up model factories for related tables. I didn't find this in the docs and thought this might be a good addition.


taylorotwell:
I don't really know that I would recommend doing it that way. More sensible to create the users first and then attach N number of posts to each user.

vjrngn:
True but what if you need to attach the users while creating the factory.
Since its already defined in out modelfactory.php.

PR #2025, opened 2016-01-04
Include mention of @push and @stack

gpmcadam:
There's no mention in any documentation of `@push` and `@stack`, which are very useful features, particularly when it comes to complex view structures or layouts. This change introduces the idea and shows two examples of the actions available.


acasar:
Nice :+1:  Didn't even know about this one and had to develop a custom directive...

PR #2027, opened 2016-01-04
Add information about removing the route helpers from laravel 5.2.*

adibhanna:


GrahamCampbell:
Maybe recommending just the facade isn't a good idea. I'd recommend using an actual router instance assigned to a variable.

taylorotwell:
I've fixed it.

PR #2030, opened 2016-01-05
[5.2] Updated Requests - dynamic property

lagbox:
Added note about how the dynamic property resolves what it is asked to return.


taylorotwell:
Fixed.

PR #2034, opened 2016-01-05
Add HTTP Kernel changes into the upgrade guide

cdun:
Running `artisan make:auth` after upgrading from 5.1 to 5.2 breaks out of the box. 
The middlewareGroups array in the HTTP kernel now contains a `web` key by default which is referenced by the scaffolded route.
New projects of course won't be affected so this is probably somewhat of an edge case, but adding for clarity.


taylorotwell:
There is a note in the release notes not to use this during application upgrades.

PR #2036, opened 2016-01-06
5.2 upgrade note about db seeds being unguarded by default

4lun:
Refers to [this change](https://github.com/laravel/framework/commit/41ff1a1fc838007b981ee27b043feae8e4974680#diff-14de5015be74d92950add13254e6eccdR63), seems worth mentioning since not everyone might test their seeds are functioning correctly immediately after upgrading.


taylorotwell:
Were your seeds not functioning correctly?

4lun:
Nope, they broke because of models becoming unguarded by default.

We have a few instances where multiple models (e.g. User and a Profile) get created using the same array of data, we use mass-assignment and the $fillable property so that each model only gets the correct attributes.

PR #2037, opened 2016-01-06
Update installation.md

zaydogan:
I have added the missing extensions Dom and Zip to the list.
I removed Tokenizer from the list, since it is enabled by default since PHP 4.3. The reason is because it is not an extension that you can install, but a compile-time option (that is enabled by default.)

### DIFF
--- a/installation.md
+++ b/installation.md
@@ -20,7 +20,8 @@ However, if you are not using Homestead, you will need to make sure your server 
 - OpenSSL PHP Extension
 - PDO PHP Extension
 - Mbstring PHP Extension
-- Tokenizer PHP Extension
+- Zip PHP Extension
+- Dom PHP Extension
 </div>
 
 <a name="installing-laravel"></a>


GrahamCampbell:
:-1: Every change here is wrong.

GrahamCampbell:
The tokenizer IS still required, even if it may be enabled by default. As for the zip and dom extensions, they're simply not required for normal operation.

zaydogan:
Define normal operation. 
You can't install the installer nor create a new laravel project without dom and zip.

zaydogan:
Tokenizer isn't an extension, but declared as one. Please fix that.

GrahamCampbell:
It can be considered an extension, and composer treats it as one, because php represents it as one internally.

zaydogan:
You can reproduce this results by disabling the zip and dom extensions in your php.ini.

$ php -v
PHP 5.6.14 (cli) (built: Oct 19 2015 20:06:59) 
Copyright (c) 1997-2015 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2015 Zend Technologies
$ laravel new blog                                                                                                                         
Crafting application...
PHP Fatal error:  Class 'ZipArchive' not found in /home/zafer/.composer/vendor/laravel/installer/src/NewCommand.php on line 128
$ 
## after enabling zip
$ laravel new blog                                                                                                                         
Crafting application...
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for phpunit/phpunit 4.8.21 -> satisfiable by phpunit/phpunit[4.8.21].
    - phpunit/phpunit 4.8.21 requires ext-dom * -> the requested PHP extension dom is missing from your system.

Application ready! Build something amazing.
$ 
## after enabling dom, everything builds fine.

GrahamCampbell:
> You can reproduce this results by disabling the zip and dom extensions in your php.ini.

Install with `--no-dev`.

GrahamCampbell:
Yes, the dom extension is needed for development.

GrahamCampbell:
In terms of zip, it's not a requirement of laravel, just our installed.

GrahamCampbell:
And, the installer is totally optional. I've never even needed to use it myself.

GrahamCampbell:
The point is that it's incorrect to say that these extensions are required to run the application in production.

zaydogan:
I agree. I only followed the Installation from the Documentation page. But you won't succeed with the steps provided as is. Therefore I tried to help by updating the information.

PR #2039, opened 2016-01-06
Add a note saying the switch to Paginator::render() is not required in 5.2

clarkwinkelmann:
I was surprised when my app worked without doing this modification. Then I found that it has been added recently: https://github.com/laravel/framework/commit/ad6e7c20df84d3be5b2ea2d360cd61d36377ff33

As the 5.2 pagination documentation uses this newly created alias, I think it is officially safe to keep using it.

I'm not sure how to formulate it in the docs but as there is no Issues page I did my best to change this myself =|

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -693,6 +693,8 @@ If your application code was injecting `Illuminate\Cache\CacheManager` to get a 
 
 Replace any calls to `$paginator->links()` with `$paginator->render()`.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Gonna hold off on this since I do want people to upgrade in case it is ever deprecated.

clarkwinkelmann:
Then maybe you want to update https://laravel.com/docs/5.2/pagination#displaying-results-in-a-view insead. Because for the moment 5.2 newcomers are invited to use the `links()` method.

PR #2041, opened 2016-01-07
add inconsistency note regarding resource route naming inside route groups

EliasZ:
Related laravel/framework#11657

[Fixed in 5.3](https://github.com/laravel/framework/pull/11679#issuecomment-169045871)

Don't know whether this is a good spot to note that inconsistency.


taylorotwell:
Hmm, the wording here is not quite right. I will consider something.

EliasZ:
:+1:

PR #2057, opened 2016-01-11
Added entry to the menu

whoan:
I added Laravel 5.3 in the menu


whoan:
...I think the change would be better if I removed the Laravel 5.3 section of the release documentation. Laravel 5.1 documentation doesn't have information about Laravel 5.2. Why Laravel 5.2 have info about Laravel 5.3?

PR #2067, opened 2016-01-14
[5.2] Update docs with 'getColumnType' method

mustafaaloko:


taylorotwell:
I don't think this is going to be a super common method to be used in user-land to be honest.

PR #2068, opened 2016-01-14
Remove ambiguity associated with 'old' in middleware.md

olimorris:
As per pull request #1974, removed the ambiguity associated with 'OldMiddleware'. This could imply an outdated middleware. AgeMiddleware is more apt.

### DIFF
--- a/middleware.md
+++ b/middleware.md
@@ -23,9 +23,9 @@ There are several middleware included in the Laravel framework, including middle
 
 To create a new middleware, use the `make:middleware` Artisan command:
 
-    php artisan make:middleware OldMiddleware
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
:-1:

GrahamCampbell:
Age doesn't tell you if it's old or new.

olimorris:
@GrahamCampbell yes precisely. The reference of `Old` may refer to an object being redundant or in existence for a prolonged period of time.

The example in the documents uses `Middleware` to prevent access to a page if, and I quote, "the supplied `age` is greater than 200".

In this context, it is more readable to name the `Middleware` as `AgeMiddleware`.

PR #2069, opened 2016-01-14
Referencing Eloquent Collections being passed to views

olimorris:
I couldn't find a reference to passing Eloquent collections to views in the docs. It is somewhat opinionated to propose the use of `compact` to do this but it is elegant and neat in its solution.


views.md
```diff
@@ -55,6 +55,10 @@ As you saw in the previous examples, you may easily pass an array of data to vie
 When passing information in this manner, `$data` should be an array with key/value pairs. Inside your view, you can then access each value using its corresponding key, such as `<?php echo $key; ?>`. As an alternative to passing a complete array of data to the `view` helper function, you may use the `with` method to add individual pieces of data to the view:
 
     return view('greeting')->with('name', 'Victoria');
+    
+You may also neatly pass [Eloquent collections](/docs/{{version}}/eloquent-collections) to your views, utilising PHP's `compact` function:
+
+    return view('admin.profile', compact('data');
```
vinicius73: Syntaxe error

GrahamCampbell:
I don't think this is needed. It's so obvious this is just making it harder to follow the docs.

PR #2070, opened 2016-01-15
explain second argument of sync() function

jrmadsen67:
sync() has an unknown but incredibly useful second argument that I think would be helpful for people know about, and takes little additional write up to explain

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -777,6 +777,13 @@ You may also pass additional intermediate table values with the IDs:
 
     $user->roles()->sync([1 => ['expires' => true], 2, 3]);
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Sync never adds duplicate records regardless of the second parameter.

jrmadsen67:
yes, that's what I said.  but attach() does, from what I've seen, making for workaround necessary if you only want to add new unique items

in any case, there is a second param that is undocumented that is useful to know. I can work it differently if you think it is unclear

taylorotwell:
Hmm, OK let me play with it in a test app and I'll come back to this.
Thanks!

jrmadsen67:
I've always wondered if it wasn't just an unintended side effect, TBH

PR #2071, opened 2016-01-15
Change "may be" to "should be" to prevent errors

JohnMaguire:
Someone came into IRC with this paste:

http://laravel.io/bin/LkJmz

They had read through the upgrade file but interpreted "may be" as optional. Correcting those references to "should be."


GrahamCampbell:
"must" is stronger than "should".

PR #2072, opened 2016-01-15
Update ValidationException use statement

nhowell:
Illuminate\Foundation\Validation\ValidationException is deprecated, use Illuminate\Validation\ValidationException instead.


GrahamCampbell:
This is incorrect. Upgrading to 5.2.0 this won't work. This needs doing on a later 5.2.x version.

nhowell:
Oh, you're correct. And I guess it wouldn't _need_ to be documented in the upgrade guide until the deprecated version is removed (Laravel 6?).

PR #2073, opened 2016-01-16
Added aliases for avg and pluck methods

jerguslejko:
Added aliases for `avg()` - `average()` and `pluck()` - `lists()` methods.


taylorotwell:
I think `lists` is deprecated?

PR #2076, opened 2016-01-17
Add references to JSX transformation in elixir docs

leonaves:
As well as ES2015, by default, Babel 5.x runs with the react (JSX) transformation enabled. Elixir currently uses babel-core 5, but Babel 6 removes all default presets, which might cause some confusion as some developers might expect from looking at the current Babel documentation for the react preset to not be included by default.

### DIFF
--- a/elixir.md
+++ b/elixir.md
@@ -198,7 +198,7 @@ elixir(function(mix) {
 <a name="browserify"></a>
 ### Browserify
 
-Elixir also ships with a `browserify` method, which gives you all the benefits of requiring modules in the browser and using ECMAScript 6.
+Elixir also ships with a `browserify` method, which gives you all the benefits of requiring modules in the browser and using ECMAScript 6 and JSX.
 
 This task assumes that your scripts are stored in `resources/assets/js` and will place the resulting file in `public/js/main.js`:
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Please send to 5.2 branch.

leonaves:
I thought 5.2 used elixir version 4, which includes babelify version 7.2, which uses Babel-core 6.x, so doesn't have this behaviour?

leonaves:
I think this should be re-opened; the issue can cause confusion (I know from experience), and it seems you agree, having asked me to send it to the 5.2 branch, but the changes I have made are not accurate in 5.2, as elixir no longer does JSX transformations by default. 5.1 is an LTS release, and so I think it's important this be added.

JeffreyWay:
@leonaves - 

>  as elixir no longer does JSX transformations by default.

Yes it does.

leonaves:
the laravel/laravel 5.2 repository includes elixir 4.0.0 which only includes the es2015 preset for browserify, as you yourself pointed out here:

https://github.com/laravel/elixir/issues/350#issuecomment-158686325

JeffreyWay:
That's an old conversation.

https://github.com/laravel/elixir/blob/master/Config.js#L280

leonaves:
Thats the master branch, on the 4.0 tag, the config.js file still includes only es2015:

https://github.com/laravel/elixir/blob/4.0/Config.js

leonaves:
You know what ignore me, there's a caret in the package.json.

PR #2083, opened 2016-01-22
unescape html + https

dmxhZGp1c2hh:
Typos.


taylorotwell:
You actually don't need the {!! on those. It will work either way since it uses HtmlString.

PR #2084, opened 2016-01-24
Add note about ties to PHP specific functionality

drowe:
Pull-request to remove dependency on PHP functionality was closed / denied.  Docs should be improved to indicated this dependency.

### DIFF
--- a/encryption.md
+++ b/encryption.md
@@ -11,6 +11,8 @@ Before using Laravel's encrypter, you should set the `key` option of your `confi
 <a name="basic-usage"></a>
 ## Basic Usage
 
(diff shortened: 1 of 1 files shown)


encryption.md
```diff
@@ -11,6 +11,8 @@ Before using Laravel's encrypter, you should set the `key` option of your `confi
 <a name="basic-usage"></a>
 ## Basic Usage
 
+**Note:** Values to be encrypted are passed through `serialize` upon encryption, and subsequently passed through `unserialize` upon decryption. Thus, non-Laravel clients receiving encrypted values will need to `unserialize` the data.
```
GrahamCampbell: non-PHP is more accurate since laravel's encryption component can be used outside of laravel, or infact, anyone else could implement something in plain php that could decrypt the message easily

taylorotwell:
This should be sent to the 5.2 branch.

PR #2092, opened 2016-01-27
Fix language files override path

Omranic:
- Add missing {vendor} segment


taylorotwell:
This is incorrect. You don't need the vendor segment.

Omranic:
My bad, you're right.
I see it depends on the registered namespace :+1: 
`        $file = "{$this->path}/vendor/{$namespace}/{$locale}/{$group}.php";
`

PR #2098, opened 2016-01-29
Change hard-coded class FQNs to class constants

Gawdl3y:
Since Laravel has required PHP 5.5 for a while now, and the move to class constants has been done in most other places in the documentation, this brings the Eloquent relationships examples up to speed.

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -55,7 +55,7 @@ A one-to-one relationship is a very basic relation. For example, a `User` model 
          */
         public function phone()
         {
-            return $this->hasOne('App\Phone');
+            return $this->hasOne(Phone::class);
         }
     }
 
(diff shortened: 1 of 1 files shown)


lagbox:
A similar PR for this was already brought up. #2095

Gawdl3y:
Ah, would you look at that - beaten by a bit over a day. Would be nice to have an explanation for why that was closed, though. Currently, this is an inconsistency in the documentation, since most of it already uses the class name constant. I won't talk about the advantages of using the constant, since they're already well-known and discussed to death.

taylorotwell:
I don't want to make this change.

taylorotwell:
It makes it ambiguous that you have to pass the fully qualified namespace when new users view the docs.

Gawdl3y:
@taylorotwell By that reasoning, wouldn't this mean that the rest of the documentation should be using hard-coded strings as well? I can't think of a situation where FQNs aren't required.
New users really shouldn't be encouraged to use hard-coded string FQNs, period, as it's a bad habit nowadays (since the `class` constant is a thing). If they always use `class` constants, when is there a case where they need to actually know that it requires a fully-qualified name?

That being said, perhaps a sentence or two (it really shouldn't require much elaboration) could be added explaining that FQNs are required, and show how using `User::class` equates to `'App\User'`. I'll go ahead and make this change to see what you think.

Gawdl3y:
See Gawdl3y/docs@2ec239b.

Gawdl3y:
@taylorotwell Thoughts?

PR #2106, opened 2016-02-01
Fix wording in middleware

cdmedia:


taylorotwell:
Original wording is correct. Middleware is plural there.

PR #2107, opened 2016-02-03
Fix the with() syntax in the Blade extensions example

dwightwatson:
Unless I'm totally mistaken, you'll need to use brackets not curly braces.


GrahamCampbell:
You are mistaken actually. Please close this.

GrahamCampbell:
The regex already pulls in the curly brackets, and the braces are merely used to get the string interpolation to work.

dwightwatson:
Ah, gotcha. Thanks for the clarification.

PR #2113, opened 2016-02-04
Updated JSON section of response.md

dhixson:
In 5.2 Illuminate\HTTP\Response no longer has a json function as the functionality has been moved to Illuminate\Http\JsonResponse. Updated the doc to show how to execute json responses.

### DIFF
--- a/responses.md
+++ b/responses.md
@@ -106,15 +106,13 @@ Of course, if you do not need to pass a custom HTTP status code or custom header
 <a name="json-responses"></a>
 #### JSON Responses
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
This isn't true. I use response()->json all the time in 5.2.

dhixson:
You are correct, response()->json works. But when you try to call with parameters, say to pass a status code, it returns the following Call to undefined method Illuminate\Http\Response::json(). In order to pass a status code with a json response, you will need to use JsonResponse.

PR #2114, opened 2016-02-05
Clarify where bug fixes should be sent

vlakoff:
Per "going off-topic" discussions:
* https://github.com/laravel/framework/pull/12046#issuecomment-180022125
* https://github.com/laravel/framework/pull/12173#issuecomment-180297611

If I recall correctly, these notes were written before LTS releases appeared.

### DIFF
--- a/contributions.md
+++ b/contributions.md
@@ -36,7 +36,7 @@ Discussion regarding bugs, new features, and implementation of existing features
 <a name="which-branch"></a>
 ## Which Branch?
 
-**All** bug fixes should be sent to the latest stable branch. Bug fixes should **never** be sent to the `master` branch unless they fix features that exist only in the upcoming release.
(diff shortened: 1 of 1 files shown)


vlakoff:
Ah, and let's ping @GrahamCampbell :)

GrahamCampbell:
Maybe we need to indicate that it must be an actual maintained branch. 4.0 for example has plenty of bugs in it that we're not fixing.

vlakoff:
Agree, maybe "oldest supported branch" would be better.

vlakoff:
Or "oldest maintained branch".

vlakoff:
@taylorotwell I just noticed you only merged this change to 5.1, but haven't merged it to 5.2 nor master. You might want to do it, the 5.2 documentation being the one shown by default.

PR #2118, opened 2016-02-08
[5.0] Adjust install PHP version requirements

lagbox:
Adjustment to exclude PHP 7 do to the psy version in 5.0 using libraries that aren't ready for PHP 7.
>"Cannot use PhpParser\Node\Scalar\String as String because 'String' is a special class name"

### DIFF
--- a/installation.md
+++ b/installation.md
@@ -43,7 +43,7 @@ Laravel ships with scaffolding for user registration and authentication. If you 
 
 The Laravel framework has a few system requirements:
 
-- PHP >= 5.4
+- PHP >= 5.4, PHP < 7
 - Mcrypt PHP Extension
 - OpenSSL PHP Extension
 - Mbstring PHP Extension


daylerees:
It works for the most part on PHP 7. If we're dependant on the library for certain features, then perhaps we should hint at that, rather than discouraging the use of version 7.

GrahamCampbell:
Maybe we should just say that for guaranteed compatibility, upgrade to 5.1+?

lagbox:
@daylerees I hear ya on that since it does work with 7 in general, this is just a psy issue. I just wanted to bring it up since someone mentioned it in the support room yesterday.

lagbox:
@daylerees @GrahamCampbell maybe a 'note' there that it does work with php7, but artisan tinker does not, to not discourage those who have 7 from using 5.0?

lagbox:
@daylerees @GrahamCampbell any thought on adjusting this to be something like

>The Laravel framework has a few system requirements:

>    PHP >= 5.4
    Mcrypt PHP Extension
    OpenSSL PHP Extension
    Mbstring PHP Extension
    Tokenizer PHP Extension

>As of PHP 5.5, some OS distributions may require you to manually install the PHP JSON extension. When using Ubuntu, this can be done via apt-get install php5-json.
>>Note: Due to the changes in PHP 7, `artisan tinker` will not work with PHP 7, this does not effect the rest of the framework and Laravel 5.0 can be used with PHP 7 otherwise.

To make it more obvious that 7 works with the framework, it is just one element that does not.

GrahamCampbell:
We only support Laravel 5.1 anyway.

PR #2119, opened 2016-02-08
Grammer check

curogihu:
In terms of grammar, checked sentences except for comments in  sample codes

### DIFF
--- a/authentication.md
+++ b/authentication.md
@@ -57,7 +57,7 @@ Laravel provides a quick way to scaffold all of the routes and views you need fo
 
     php artisan make:auth
 
-This command should be used on fresh applications and will install registration and login views, as well as routes for all authentication end-points. A `HomeController` will also be generated, which serves post-login requests to your application's dashboard. However, you are free to customize or even remove this controller based on the needs of your application.
+This command should be used for fresh applications and will install registration and login views, as well as routes for all authentication end-points. A `HomeController` will also be generated, which serves post-login requests to your application's dashboard. However, you are free to customize or even remove this controller based on the needs of your application.
 
 <a name="included-views"></a>
 ### Views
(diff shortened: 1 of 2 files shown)


authentication.md
```diff
@@ -221,7 +221,7 @@ We will access Laravel's authentication services via the `Auth` [facade](/docs/{
         }
     }
 
-The `attempt` method accepts an array of key / value pairs as its first argument. The values in the array will be used to find the user in your database table. So, in the example above, the user will be retrieved by the value of the `email` column. If the user is found, the hashed password stored in the database will be compared with the hashed `password` value passed to the method via the array. If the two hashed passwords match an authenticated session will be started for the user.
+The `attempt` method accepts an array of key / value pairs as its first argument. The values in the array will be used to find the user in your database table. So, in the example above, the user will be retrieved from the value of the `email` column. If the user is found, the hashed password stored in the database will be compared with the hashed `password` value passed to the method via the array. If the two hashed passwords match an authenticated session will be started for the user.
```
daylerees: Not sure about this one. I think `by` works.

Think of repository methods. `$repo->getUserByEmail();`

authentication.md
```diff
@@ -380,7 +380,7 @@ After the password is reset, the user will automatically be logged into the appl
 
     protected $redirectTo = '/dashboard';
 
-> **Note:** By default, password reset tokens expire after one hour. You may change this via the password reset `expire` option in your `config/auth.php` file.
+> **Note:** By default, password resets tokens expire after one hour. You may change this via the password reset `expire` option in your `config/auth.php` file.
```
daylerees: Also this was correct in the original sentence.

taylorotwell:
Not all of these changes are correct.

curogihu:
Excuse me. If I think wrong grammar happens 8 times, should I do pull request 8 times because all my thinking may be wrong?

authentication.md
torkiljohnsen: PR is closed, so not that it matters, but I think "the user will be retrieved by the value" sounds like the value is retrieving the user. Instead, the user should be retrieved by using the value.

I would go for "by using", or just "from", as was suggested :)

PR #2122, opened 2016-02-09
Added note to `requests.md` about cookie names

hassankhan:


taylorotwell:
Where do we actually make this conversion. Also, would probably go to 5.2 docs.

hassankhan:
From my testing, I see that it only happens after the response is returned from the controller. This PR was because of https://github.com/laravel/framework/issues/12212, wasn't sure if it was intended behaviour or not. I'll try and make a PR to show the exact problem.

PR #2123, opened 2016-02-09
[5.1+] One can directly pass the middleware class name

vlakoff:
I was irritated by the "`Http\Kernel $routeMiddleware`" extra step, until I realize I could avoid it.


taylorotwell:
Could you send this to 5.2 branch please.

PR #2124, opened 2016-02-09
Update JSON post example to use `$this->json`

rtablada:
The current docs use `$this->post` to send a request to JSON APIs.
However, the `$this->post` method sends form encoded data.
While this works for APIs using `Request::get`, it breaks for APIs which rely on `Request::json`.

This PR updates the docs to use the `$this->json` method for testing JSON APIs.


taylorotwell:
Thanks

PR #2131, opened 2016-02-10
Add note about retrieving JSON input

dabernathy89:
Newcomers may not know that Laravel automatically maps JSON data to the request input, especially if they are just sending data with something like jQuery and not setting their Content-Type header.

There was a note about accepting JSON input in the 4.2 docs, which is where I borrowed the language from.

At the very least, this will be a hint for anyone who comes to the Requests page of the docs and does a cmd+F for 'json'.

### DIFF
--- a/requests.md
+++ b/requests.md
@@ -169,6 +169,8 @@ You may also access user input using dynamic properties on the `Illuminate\Http\
 
 When using dynamic properties, Laravel will first look for the parameter's value in the request payload and then in the route parameters.
 
+> **Note:** Some JavaScript libraries may send input to the application as JSON. If the request's `Content-Type` header is set properly, you may access this data via the request instance methods as normal.
+
 <a name="old-input"></a>
 ### Old Input
 


GrahamCampbell:
:-1: This unrelated to laravel.

dabernathy89:
It was mentioned in the 4.2 docs:

https://laravel.com/docs/4.2/requests

Not sure how it's unrelated? Somebody new coming to the framework would have no idea that a request can be JSON rather than form data by reading the docs. And if they try to pass JSON it won't work unless they set their headers properly, which they might not know to do.

taylorotwell:
Just curious how are you accessing your request data when you POST JSON to your application? Request::json or Request::input()?

dabernathy89:
just using `input()`. I think there are a couple of scenarios where this note would be useful:

- someone reading the docs before trying out Laravel. They get to the section on handling input, but it doesn't say anything about JSON. They might assume that they have to decode it manually.

- someone using the framework for the first time, and posting JSON data but not setting their Content-Type header.

I think just having some mention of JSON on the page would be useful, considering the work that the Request class does to handle JSON input elegantly.

GrahamCampbell:
> someone using the framework for the first time, and posting JSON data but not setting their Content-Type header.

Nothing to do with the framework though. This is required everywhere because that's how HTTP works.

dabernathy89:
I disagree:

https://github.com/laravel/framework/blob/5.2/src/Illuminate/Http/Request.php#L661

If you send XML with the correct Content-Type header, Laravel won't magically let you access it as standard form input, right? The framework goes out of its way to account for JSON. Shouldn't that at least be noted in the docs?

PR #2133, opened 2016-02-13
Added "Setting Paths to Your Public and Private SSH Keys" to "Configuring Homestead" section

MeiRct:
The explantion was in previous branches and omitted on this branch for some reason.
I deciced to re-add it mainly for Windows users which encounter an error due to missing SSH key.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -80,6 +80,29 @@ Once you have cloned the Homestead repository, run the `bash init.sh` command fr
 The `provider` key in your `~/.homestead/Homestead.yaml` file indicates which Vagrant provider should be used: `virtualbox`, `vmware_fusion`, or `vmware_workstation`. You may set this to the provider you prefer:
 
     provider: virtualbox
+    
+#### Setting Paths to Your Public and Private SSH Keys
+
+The `authorize` property of your `Homestead.yaml` file holds the path to your public SSH key file:
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
In general I think it's just extra work since you can already do `vagrant ssh` to access the box without having to configure any of this?

TerrePorter:
@taylorotwell You do realize that not all of the programmers in the world are on a Mac and the vagrant ssh does not work straight out of the box on windows. So this would actually be useful information.

MeiRct:
@TerrePorter Even before the `vagrant ssh` command, most of Windows users who follow the docs, will fail to pass the first `vagrant up` command.

@taylorotwell As I mentioned in my first comment, large amount of Windows users who follow the installation steps in Homestead docs, will probably encounter an error during `vagrant up` command, due to absence of `id_rsa` and `id_rsa.pub` SSH keys in `~/.ssh/` directory by default, as well as any other SSH keys in their machine, and even absence of any knowledge about SSH keys.

I don't think it's a good idea to have a known and very possible error unhandled, in any kind of tutorial.

Are you okay with users who follow every step and instruction of the docs, and still find themselves search the internet for an error message they ran into, and then find the answer only in earlier branches of same docs, or in laracasts forum?

PR #2135, opened 2016-02-15
Event implements a mediator pattern, not an observer pattern

moon0326:
I always thought Laravel's event system is based on a mediator implementation, but the doc says it is an observer implementation. As the Laravel acts as a central point for all the listeners, isn't it acting as a mediator?

### DIFF
--- a/events.md
+++ b/events.md
@@ -17,7 +17,7 @@
 <a name="introduction"></a>
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think it's observer.

moon0326:
Not trying to bring that miserable facade thing or something similar, but how is this an observer? Would you mind explaining a little?

You never listen to an object directly. Listen/Dispatch are done through the dispatcher class. Isn't this what a mediator?

PR #2140, opened 2016-02-16
Add array syntax for dropping indexes

barryvdh:
This add the easier syntax, instead of manually creating the index name.


taylorotwell:
Interesting. I didn't know you could do this, heh.

barryvdh:
Hah, didn't notice it before today as well, but it's in there at least since L4 ;)

If it wasn't a BC break I would say it would have been a better default behavior as well :)

PR #2143, opened 2016-02-17
Updated docs with missing step for 5.1 to 5.2

jeremymlane:
Added a missing step for the upgrade guide from 5.1 to 5.2. The except parameter for the guest middleware in AuthController.php needs to be updated when going from 5.1 to 5.2.


taylorotwell:
This isn't actually correct. You would only need to make this change if you re-defined you routes.

PR #2145, opened 2016-02-18
Add suggestion for `dump` method

timkrins:
Many people see this page and go looking for a ```dd``` method that does not halt the execution of the script.

I'm not sure if here is the best place to put it, but for lack of a better place, it sure would save some time for those in the same situation as myself.

### DIFF
--- a/helpers.md
+++ b/helpers.md
@@ -754,7 +754,7 @@ The `dd` function dumps the given variable and ends execution of the script:
 
     dd($value);
     
-If you do not want to halt the execution of your script, try use the `dump` function included from Symfony instead. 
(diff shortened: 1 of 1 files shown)


dabernathy89:
FYI `try use` should be `try using`. I would also change `from Symfony` to `with Symfony`, but I don't think that is as necessary grammatically.

timkrins:
No worries, updated. Cheers

TerrePorter:
Good note to have in the docs. I ended up doing this https://gist.github.com/TerrePorter/a3b2eceff35636705308 for a dd that doesn't die.

PR #2149, opened 2016-02-19
Add docs for morph map

phroggyy:
This PR adds documentation about the polymorphic relationship morph map


eloquent-relationships.md
```diff
@@ -424,6 +424,24 @@ You may also retrieve the owner of a polymorphic relation from the polymorphic m
 
 The `imageable` relation on the `Photo` model will return either a `Staff` or `Product` instance, depending on which type of model owns the photo.
 
+#### Custom polymorphic types
```
devonzara: Should probably use title case here. :)

PR #2150, opened 2016-02-19
change FQCN to use ::class keyword

BostjanOb:
Changed passing FQCN as string and instead use ::class keyword.
If we use ::class keyword in configs, middlewares, services, .... we should use it here. And it's even shorter to write, because model classes are frequently in same namespace.


lagbox:
#2095  #2098 For reference

PR #2157, opened 2016-02-22
[5.2] Blade - Missing parens in string interpolation

ferhtgoldaraz:
Given that `with` is a helper function, if `$foo = 44`, `"with{$foo}"` is
`"with44"`, not `"with(44)"`.


brunogaspar:
https://github.com/laravel/docs/pull/1564

ferhtgoldaraz:
I see.

PR #2161, opened 2016-02-24
Added important note about APP_ENV and APP_DEBUG.

Donatello-za:
My previous pull request failed due to me not opening the pull request from my forked repo. 

Regarding the note:
This note could save developers hours of debugging if they decide to change either the APP_ENV or APP_DEBUG settings in their local .env file and the changes are not reflected during runtime. The issue of course being that these two settings are also located in php-fpm.conf and basically overrides the two corresponding settings in the developers .env file on Homestead.

### DIFF
--- a/configuration.md
+++ b/configuration.md
@@ -40,6 +40,8 @@ Your `.env` file should not be committed to your application's source control, s
 
 If you are developing with a team, you may wish to continue including a `.env.example` file with your application. By putting place-holder values in the example configuration file, other developers on your team can clearly see which environment variables are needed to run your application.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't think that's the case on recent versions of Homestead.

Donatello-za:
Thanks, but myself and a co-developer had this issue with Homestead version 3.0.1 very recently. I noticed that there is a homestead/scripts/clear-variables.sh in the homestead repo which would clear those settings but I'm not sure when it is executed or whether it should be manually executed.

PR #2162, opened 2016-02-24
auth middlewares upgrade guide

spescina:
added auth middlewares update in the 5.1 -> 5.2 upgrade guide


taylorotwell:
They don't specifically require changing when you are upgrading I don't think?

spescina:
If you don't change them multi auth feature does not work when specified like this `auth:api`

PR #2164, opened 2016-02-24
[5.2] Add Routing to Upgrade Guide

chrisloftus:
To make users aware of the new `web` middleware group for routing.


danieliancu:
I agree some users are confused by this, but when you upgrade to `5.2` this is not a breaking change, it will work just like you expect it.

It's written in docs at the correct place and only happens on new projects or if you manually change your `routes.php` file in which case you should be fully aware of this.
This has been changed in laravel/laravel not laravel/framework.

Maybe a note about this should be added, but I don't agree with the way you have written it because it confuses people even more making them think of this as a breaking change and experiencing the opposite issue when the web group middleware is not defined.

chrisloftus:
@danieliancu I had problems with my login form after upgrading from 5.1 to 5.2. Wrapping my routes inside the web middleware group fixed it. Are you sure it's not a breaking change? If not then fair enough.

I agree the wording can be changed - it's pretty much the same as on the routing docs. There are a lot of people having problems on StackOverflow, etc with sessions or CSRF not working, because their routes aren't in the group. I feel that new upgraders should be made aware.

danieliancu:
@chrisloftus yes.

`5.1`: https://github.com/laravel/laravel/blob/5.1/app/Http/Kernel.php
`5.2`: https://github.com/laravel/laravel/blob/master/app/Http/Kernel.php

If you upgrade a project from `5.1` to `5.2` as described in the upgrade guide you will not experience this issue at all.

As I have said above, it will only happen when you upgrade and change middleware manually in `app/Http/Kernel.php` but don't update `routes.php` accordingly.

chrisloftus:
Maybe we should add something like:

> If you want to use the web middleware group for your routes like in 5.2, make sure you add `$middlewareGroups` to your `app/Http/Kernel.php` file. Like so: https://github.com/laravel/laravel/blob/v5.2.15/app/Http/Kernel.php

taylorotwell:
I agree that this is not necessarily a required part of the ugprade.

PR #2166, opened 2016-02-25
->cron example

AlexeyMez:
It should be `->cron('* * * * *') instead of ->cron('* * * * * *')` I guess

### DIFF
--- a/scheduling.md
+++ b/scheduling.md
@@ -75,7 +75,7 @@ Of course, there are a variety of schedules you may assign to your task:
 
 Method  | Description
 ------------- | -------------
-`->cron('* * * * * *');`  |  Run the task on a custom Cron schedule
+`->cron('* * * * *');`  |  Run the task on a custom Cron schedule
 `->everyMinute();`  |  Run the task every minute
 `->everyFiveMinutes();`  |  Run the task every five minutes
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
No?

danieliancu:
The PR is correct, but the comment isn't.

There are 5 stars in cron: minute, hour, day of the month, month, day of the week.

AlexeyMez:
Sorry, there is a mistake in comment. The problem is there are six asterix in documentation, but it should be five.

GrahamCampbell:
See https://github.com/laravel/docs/pull/1950.

danieliancu:
@GrahamCampbell You are correct.

Looking at https://github.com/laravel/framework/blob/5.2/src/Illuminate/Console/Scheduling/Event.php clearly reflects that it has been implemented with 6 arguments, maybe the first one is for seconds?

I find it very confusing because the Linux cron is implemented with 5 arguments. I'll discuss it on Slack.

taylorotwell:
There is an optional 6th "year" argument available: http://www.nncron.ru/help/EN/working/cron-format.htm

danieliancu:
@taylorotwell thanks for the pointer.

AlexeyMez:
Thanks for the explanation.

PR #2178, opened 2016-02-29
Added NETWORKS homestead configuration to the docs. (5.2)

cpriego:
Hi Taylor, I added this new entry to the docs because I had a hard time figuring it out. Hope it helps everybody.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -97,6 +97,27 @@ To enable [NFS](http://docs.vagrantup.com/v2/synced-folders/nfs.html), just add 
           to: /home/vagrant/Code
           type: "nfs"
 
+#### Configuring Network Interfaces
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
What is the typical use-case of configuring networks like this?

cpriego:
The most common use cases are:
- Exposing the homestead machine directly so you can access it from other devices in the network without opening your machine's ports.
- Allowing the VM to access resources in your network (like DB servers)

Also I think it is useful since the homestead config already allows for it but it is nowhere documented.

TerrePorter:
+1
I actually already have this setup on one of my homestead installs. This is so I can show progress to others who want to see without needing to have to come to my desk or try to copy the site to a shared dev server. 
I get that it likely would not be common for most users, but it would have been nice if it was documented for those could use it as it would have saved me some time finding how to make it work.

PR #2184, opened 2016-03-02
Added Bail validation rule note

joel-james:
`bail` validation rule was added to Laravel 5.2 validator class. Need to mention that too in the release doc right?


taylorotwell:
Hmm yeah I guess we need a good policy for new things we add to a current release. I wonder if we should make a section for the point release that it was included in?

joel-james:
At least a small list of newly added features with the reference link will do?

PR #2187, opened 2016-03-02
Add backslash before Illuminate reference

perkola:
Required for correct referencing and avoiding a not found exception


GrahamCampbell:
:-1:

PR #2188, opened 2016-03-02
Updated MIME Type validation rules documentation

jarichypma:
I updated the MIME Type validation rules documentation. 
The current description does not reflect the actual working (anymore?).

The MIME validation `mimes` rule doesn't validate against the MIME type of the file, it only checks for a valid file extionsion using the method `guesExtension()`

In order to validate against a MIME Type I documented the use of the validation rule `mimetypes`. This validation rule wasn't documented yet.

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -502,7 +502,8 @@ Below is a list of all available validation rules and their function:
 [IP Address](#rule-ip)
 [JSON](#rule-json)
 [Max](#rule-max)
-[MIME Types (File)](#rule-mimes)
+[MIME (File)](#rule-mimes)
+[MIME Types (File)](#rule-mimetypes)
 [Min](#rule-min)
 [Not In](#rule-not-in)
 [Numeric](#rule-numeric)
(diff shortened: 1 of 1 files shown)


taylorotwell:
The Symfony guessExtension method still examines the MIME type of the file underneath the hood. In fact it calls the ->getMimeType() method.

taylorotwell:
The `mime` rule doesn't just do a blind check of the passed extension from the browser.

taylorotwell:
see the following code in Symfony\Component\HttpFoundation\File\File:

```
    /**
     * Returns the extension based on the mime type.
     *
     * If the mime type is unknown, returns null.
     *
     * This method uses the mime type as guessed by getMimeType()
     * to guess the file extension.
     *
     * @return string|null The guessed extension or null if it cannot be guessed
     *
     * @see ExtensionGuesser
     * @see getMimeType()
     */
    public function guessExtension()
    {
        $type = $this->getMimeType();
        $guesser = ExtensionGuesser::getInstance();

        return $guesser->guess($type);
    }
```

jarichypma:
Hi Taylor,
Thanks for your response. Reading your response I'm afraid I didn't cleary describe what I mean to address. What I'm trying to document is this issue I ran into today:

1. Validating an uploaded .xlsx (Excel) file against the `mimes` validation rule doesn't work with the Mime type argument `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`.
2. When the validateMimes-method examines the uploaded Excel file, the guessExtension returns `xlsx`
3. The in_array filter returns false comparing `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` with `xlsx`
4. As a result the validation fails
5. Using the validateMimetypes validation rule solves this. Because the validateMimetypes method uses getMimeType for the in_array filter.

Jarich

PR #2196, opened 2016-03-05
Eloquent: public -> protected for timestamps

hansott:
Shouldn't the $timestamps property be protected?


lagbox:
It is public in Model.php

hansott:
I see... Thanks @lagbox! :bow: I'll close my PR.

PR #2200, opened 2016-03-09
Added morphMap details

skcin7:
Added that the morphMap should be added in the boot function of a service provider, and added the FQCN of the Relation class used to the example code, so that users are not confused about what/where the namespace of the Relation class.


lagbox:
:+1:

skcin7:
@lagbox  Will this go live in the 5.2 docs? I'd feel mighty proud and accomplished if it did ^_^

PR #2203, opened 2016-03-09
Remove optional year field from Cron example

vlakoff:
Did not understand what that 6th asterisk was. It's an optional year field added by the [Cron Expression](https://github.com/mtdowling/cron-expression) library.

I suggest removing it, as it is non-standard Cron (therefore confusing) and seldom useful.

### DIFF
--- a/scheduling.md
+++ b/scheduling.md
@@ -75,7 +75,7 @@ Of course, there are a variety of schedules you may assign to your task:
 
 Method  | Description
 ------------- | -------------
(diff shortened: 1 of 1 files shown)


vlakoff:
Maybe it should be removed in [`Event.php`](https://github.com/laravel/framework/blob/5.2/src/Illuminate/Console/Scheduling/Event.php) as well? (methods `hourly`, `daily`, etc.)

taylorotwell:
It's supported.

PR #2207, opened 2016-03-10
add sample workflow

garethnic:
Based on people struggling to get to grips with creating packages. This section gives an overview of what creating a package might look like.


taylorotwell:
Going to hold off on recommending this as an official approach.

PR #2211, opened 2016-03-13
added psr-http-message-bridge component requirement

appkr:
This is remedy for

```bash
FatalThrowableError in Router.php line 1085:
Fatal error: Class 'Symfony\Bridge\PsrHttpMessage\Factory\HttpFoundationFactory' not found
```

This happened, when I send an email using mailgun driver. After some research, I found that the required Class was not included out of the box. 

As you see in diff, what I did was just add a line in mail.md

```bash
"symfony/psr-http-message-bridge": "~0.2"
```

I'm using a framework version 5.2.22. Thanks.

### DIFF
--- a/mail.md
+++ b/mail.md
@@ -17,7 +17,8 @@ Laravel provides a clean, simple API over the popular [SwiftMailer](http://swift
 
 The API based drivers such as Mailgun and Mandrill are often simpler and faster than SMTP servers. All of the API drivers require that the Guzzle HTTP library be installed for your application. You may install Guzzle to your project by adding the following line to your `composer.json` file:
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Have never heard of such a package being required.

GrahamCampbell:
Yeh, this isn't needed to get mail working.

GrahamCampbell:
> FatalThrowableError in Router.php line 1085:

Looking at the error there, you can see clearly it's unrelated to mail.

appkr:
I just don't want people to be in trouble like me. Anyway the fact is that the class is not there. I hope there be correction, whatever it is. Thanks.

PR #2216, opened 2016-03-16
Add namespacing warning

DaniAkash:
There is an option in PHP artisan to name the app that is being created by

`php artisan app:name newname`

However if any special characters are used in the application name it will cause an almost fatal error making the application useless hence the warning must be mentioned in a place which can be read by everyone who see the documentation.


taylorotwell:
I prefer people just don't rename their apps.

PR #2217, opened 2016-03-17
add documentation on starting the server

thenickcox:
I'm a Rails developer, and my first ever exposure to Laravel went like this: Laravel site -> Installation instructions -> Quickstart. That seemed like a reasonable path for me to follow.

I got about halfway through the quickstart and realized that I wanted to start up the server as a sanity check to make sure everything was working. But it's nowhere to be found.

I thought this was odd, especially since instructions on starting the server [in the docs for 4.2](https://laravel.com/docs/4.2/quick).

I mention the piece about being a Rails developer and my exposure to Laravel because I had not yet seen the `php artisan serve` command and had to Google to find it. I finally found it and ran it and realized I had to fix a couple things, which derailed my progress on the quickstart. I would have been even more confused and frustrated had I gotten to the end of the tutorial (by which time the server is still not started) and realized I had no idea how to check on my progress and, when I found issues, I'd have no idea at which step in the process they were introduced.

This PR adds (or, rather, re-adds) a quick line or two about how to start the server at what seems like the appropriate time: the first point in the tutorial in which the user would get any output in the browser. It's possible that this was removed intentionally, but I can tell you as a new user that this is a really key piece of information to have in the quickstart.


thenickcox:
Sorry, not sure what happened to my git history. I'll try again.

PR #2223, opened 2016-03-21
Warn about .env values with spaces

tomcastleman:


taylorotwell:
Feel like this should maybe be in the config section.

tomcastleman:
@taylorotwell could you clarify whether you mean I should add this to the config docs in addition or instead of in the upgrade guide? I can add it to the configuration docs too, but I added it here because it is a breaking change between 5.1 and 5.2 (between `vlucas/phpdotenv` v1 and v2).

Thanks

tomcastleman:
A little frustrating this has been closed with no decent feedback. Any thoughts @GrahamCampbell ?

tomcastleman:
@GrahamCampbell Following on from [your comment](https://github.com/laravel/framework/issues/11715#issuecomment-204737355), what further context to can I provide on this? We just need to put something in along these lines so give me something to go on or just add it in yourself.

tomcastleman:
The complete lack of response on this is disappointing.

philbates35:
I fell victim to this last week. The resulting stacktrace has no mention of an invalid value in the `.env` file either, so it took far too long to work out what was wrong.

PR #2227, opened 2016-03-24
Update queries.md

dsshafer:
Mention use of orderBy() with chunk(), and include in code examples

### DIFF
--- a/queries.md
+++ b/queries.md
@@ -71,9 +71,9 @@ If you don't even need an entire row, you may extract a single value from a reco
 
 #### Chunking Results From A Table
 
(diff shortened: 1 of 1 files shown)


dsshafer:
Taylor, on this commit you included the change to the sample code for chunk(), but not the explanation about _why_ a person would use orderBy() with chunk(). I can't imagine the hours people have spent troubleshooting what's already an understood issue-- I lost a few days on it. If this isn't going to be addressed in the framework, then can we at least mention it in the docs? I understand you don't like the ideas presented thus far for changing the behavior, but we really need at least a hint in the docs about what's going on. People are trying to help. Here are all the related issues and PRs people have submitted (all closed without resolution):

https://github.com/laravel/framework/issues/11302
https://github.com/laravel/framework/pull/11490
https://github.com/laravel/framework/pull/11489
https://github.com/laravel/framework/pull/11510
and of course this PR, which was a last-ditch effort to at least document the issue.

PR #2229, opened 2016-03-27
Add note regarding testing environment variables

carlosafonso:
The behavior I reported in https://github.com/laravel/framework/issues/12874 turns out to be the expected one. And it has indeed a pretty straightforward solution, however I couldn't find a mention to this in the Testing section of the documentation.

This PR adds a note addressing this particular topic. I think it might save developers some time if they come across this.

### DIFF
--- a/testing.md
+++ b/testing.md
@@ -29,6 +29,8 @@ When running tests, Laravel will automatically set the configuration environment
 
 You are free to create other testing environment configurations as necessary. The `testing` environment variables may be configured in the `phpunit.xml` file.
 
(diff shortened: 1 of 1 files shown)


skvskv:
@carlosafonso , have you by any chance found a way to clear config from e.g. phpunit.dist.xml, so that you would not have to do it manually?

carlosafonso:
You could clear the cache from within the bootstrap script defined in `phpunit.dist.xml`.

Here's something similar from the Symfony docs: https://symfony.com/doc/current/testing/bootstrap.html

Would this work for you?

PR #2237, opened 2016-03-29
add a note to migrations about dropping foreigns when renaming a table

dbhynds:
I'm new to Laravel, and didn't know you should drop foreign keys and re-add them when renaming tables. After making this mistake for a while, I eventually had rebuild all of the migrations on one of my projects. I figure I can't be the only one who's made this mistake.

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -169,6 +169,8 @@ To drop an existing table, you may use the `drop` or `dropIfExists` methods:
 
     Schema::dropIfExists('users');
 
+> **Note:** Before renaming a table, you will need to drop any existing foreign keys. After the table has been renamed, you will need to add them back on the new table name.
+
 <a name="creating-columns"></a>
 ### Creating Columns
 


taylorotwell:
What happens if you don't?

dbhynds:
The foreign key references the old table name. I think I've actually figured out how to duplicate it finally. I'll post over on laravel/laravel about it in more depth and an example of how to duplicate it.

dbhynds:
Errr, actually, I'll just put it here. Try running this in a migration.

```Schema::create('test',function(Blueprint $table){
    $table->increments('id');
    $table->integer('user')->unsigned();
    $table->foreign('user')->references('id')->on('users');
});
Schema::rename('test','new_test');
Schema::table('new_test',function($table){
    $table->dropForeign(['user']);
});```

I get an error: General error: 1025 Error on rename of './bnbapp/new_test' to './bnbapp/#sql2-471-d1' (errno: 152) (SQL: alter table `new_test` drop foreign key `new_test_user_foreign`)

It seems that the foreign key is still named `test_user_foreign`, not `new_test_user_foreign` as I (and the dropForeign method) expected it to be.

taylorotwell:
OK clarified the note a bit. You can get around this by assigning a explicit name ot the foreign key in your migrations.

PR #2239, opened 2016-03-30
Remove unnecessary 'web' middleware

JeffreyWay:
Salutations. This PR updates the quick start example to remove the 'web' middleware group reference.

Bye.


GrahamCampbell:
Seems legit. We might want to add a small caveat saying this was changed in a patch release in 5.2, so people who newed up an app a few weeks ago, then came back to it aren't confused.

PR #2245, opened 2016-04-02
Echo csrf field as unescaped

BenCavens:


lagbox:
I don't think it does escape the return value of `csrf_field()` (`HtmlString`) using `{{ }}`.

GrahamCampbell:
Yeh, this can be closed please.

PR #2250, opened 2016-04-05
fix withoutGlobalScope examples

jadjoubran:


lagbox:
The docblock says it takes a string or a scope

    @param  \Illuminate\Database\Eloquent\Scope|string  $scope

jadjoubran:
So would the `AgeScope::class` work?

GrahamCampbell:
Both should be fine.

lagbox:
Yes, the class constant returns a string.

jadjoubran:
Then it seems I have a particular issue that's not making it work :fearful: 
Sorry for the noise

lagbox:
No worries.
You may have an issue with namespacing.

PR #2256, opened 2016-04-07
Add dropForeign array notation to 5.1 docs

CausaMortis:


barryvdh:
:+1: If so, also backport the other paragraph about `Dropping Indexes` (few lines above)

CausaMortis:
Ah, I wasn't aware. Added

PR #2258, opened 2016-04-08
is this a typo in the code example

summerblue:


lagbox:
I don't think so. You can map them how ever you want to name them.

PR #2262, opened 2016-04-11
Remove documentation about composite primary key

nanaya:
It's not supported by eloquent as per laravel/framework#12442 (and in many other issues).

Mentioning it in migration documentation but then not being able to actually use it (or usable in broken way) would lead to sadness.

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -318,7 +318,6 @@ Laravel will automatically generate a reasonable index name, but you may pass a 
 Command  | Description
 ------------- | -------------
 `$table->primary('id');`  |  Add a primary key.
-`$table->primary(['first', 'last']);`  |  Add composite keys.
 `$table->unique('email');`  |  Add a unique index.
 `$table->unique('state', 'my_index_name');`  |  Add a custom index name.
 `$table->index('state');`  |  Add a basic index.


nanaya:
Anyone using it for many to many intermediate table should just use composite unique key instead.

GrahamCampbell:
:-1:

GrahamCampbell:
Just because eloquent doesn't support something, it doesn't mean you don't want to write migrations for it. You might be using doctrine with laravel, or whatever you want.

nanaya:
unfortunately nowhere in the documentation is mentioned that whilst it can be created, nothing in default laravel can use it. I'd suggest either remove it or add note that eloquent can't use it.

GrahamCampbell:
This is unrelated to the migrations though?

nanaya:
But related to the whole framework? Or am I wrong assuming that everything written in migration page (part of laravel framework) is supported by eloquent (part of laravel framework) unless mentioned otherwise?

GrahamCampbell:
> Or am I wrong assuming that everything written in migration page (part of laravel framework) is supported by eloquent

Yes, that is an incorrect assumption.

nanaya:
Okay, fine.

PR #2265, opened 2016-04-12
Add Public Driver explanation

barryvdh:
This is explained here: http://taylorotwell.com/spark-storage/
But is not explained on the docs. I added it to make the 'standard' clear for everyone.

### DIFF
--- a/filesystem.md
+++ b/filesystem.md
@@ -28,6 +28,16 @@ When using the `local` driver, note that all file operations are relative to the
 
     Storage::disk('local')->put('file.txt', 'Contents');
 
+#### The Public Driver
+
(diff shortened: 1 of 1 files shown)


philbates35:
From the laravel/internals#34 issue, @ tomschlick suggested the following:

> You could always write a quick artisan command to do that... storage:symlink-public which would check the OS and create the appropriate symlink for that system during bootstrap. 

Do you think that would be something worthwhile adding to the framework?

barryvdh:
Sure

taylorotwell:
Going to adjust this a bit because there is no "public" **driver**, it's just a disk.

brian-lamb-software-engineer:
Coming from #34 here, [This doc](http://taylorotwell.com/spark-storage/) is no longer accessible and I'm on the hunt to understanding how to not make it public, that is to keep sensitive files more protected.  

For e.g., I see that the directory my sensitive files are currently uploading to is `storage/app/public/uploads`.

 I'm concerned about that word `public`.  Is there any documentation on how to ensure its never public?  I wouldn't want something automated to happen which causes the files to be publicly accessible.  

Regards.

PR #2269, opened 2016-04-13
Add a note explaining the pitfalls of ->get

kmark:
See laravel/framework#13138.

I'm not confident on the wording or placement here. I imagine we'd have to summarize the argument put forth in the above issue in just a sentence or two. Is "using Laravel's validation" descriptive enough or too broad?

One alternative I've considered:
> **Note:** When using Laravel's validation, do not use `$request->get('name')` in place of `$request->input('name')` as the former may potentially return a different value than the one Laravel has validated against.

### DIFF
--- a/requests.md
+++ b/requests.md
@@ -139,6 +139,8 @@ When working on forms with array inputs, you may use "dot" notation to access th
 
     $names = $request->input('products.*.name');
 
+> **Note:** When using Laravel's validation, do not use `$request->get('name')` in place of `$request->input('name')` as the former may potentially return a different unvalidated value.
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
`$request->get()` is never mentioned in the documentation though? I'm not sure why anyone would try to call it?

kmark:
It's a publicly exposed/available method so it will show in, as an example, PhpStorm method listings and appear to work as one might expect given its name. It seems here that the Symfony and Laravel methods conflict rather than compliment each other. I do agree that bringing up this method out of nowhere in the documentation specifically for this warning could be odd.

PR #2271, opened 2016-04-15
Add SQLite section and indicate .env configuration required

mpavel:
I spent quite a bit of time trying to figure out how to configure the local SQLite database within the `.env` file.

In config/database.php, the sqlite section contains:
```
'driver' => 'sqlite',
'database' => env('DB_DATABASE', database_path('database.sqlite')),
```

I couldn't find in the existing documentation a way to (re)use this function `database_path` and it was only after searching online and trying both relative and absolute paths that I made it work. The answer is that it works with an absolute path.

It might be useful for others to see how to configure SQLite within .env file.


mpavel:
That's correct, but if you want to use another sqlite file, and not the default named `database.sqlite`? There is no indication on how to reference this.

PR #2277, opened 2016-04-19
Add phpunit path explanation

CausaMortis:
found myself Googling this several times. Might as well save others the hassle by being explicit

### DIFF
--- a/testing.md
+++ b/testing.md
@@ -23,10 +23,7 @@ Laravel is built with testing in mind. In fact, support for testing with PHPUnit
 
 An `ExampleTest.php` file is provided in the `tests` directory. After installing a new Laravel application, simply run `phpunit` on the command line to run your tests.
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
:-1:

GrahamCampbell:
Laravel does not recommend global installs.

CausaMortis:
ok so can we at least explain that you should be using the full path to the binary rather than calling the alias?

GrahamCampbell:
Yes.

CausaMortis:
is there anything I need to do in order to include this in the other branches going back to 4.2 as well?

gmsantos:
@GrahamCampbell Laravel doesn't recommend only for phpunit, right?

For Laravel Installer it does recommend global installation:

https://laravel.com/docs/5.2/installation

taylorotwell:
Going to hold off on this.

PR #2281, opened 2016-04-22
Update views.md

neovive:
Illuminate\Users\Repository is no longer available.


taylorotwell:
This should be changed to something else. I'll fix it.

PR #2289, opened 2016-04-27
Added mention on how to test/instanciate ServiceProviders within TestCases

crazycodr:
In reference to: https://github.com/laravel/framework/issues/13335#issuecomment-215083712


taylorotwell:
I'm not sure if this needs to be explicitly documented? Service providers are never registered in the container.

PR #2298, opened 2016-04-30
Reloading Many to Many Polymorphic Relationships

KongrooDev:
From my understanding many to many polymorphic relationships are only loaded once in testing. This means that they must be reloaded in order to check to see if they have been modified. I did not see any mention of this in the docs so I wanted to add a short section on it.


taylorotwell:
This basically applies to all relations.

PR #2299, opened 2016-05-05
Use pushd & popd instead of cd in convenience func

bdunn313:
The current suggested convenience function in the Homestead documentation causes the user's current shell to actually change directories to the specified Homestead directory depending on how their default shell works.

Moving to pushd & popd circumvents this issue. The downside is that the function is less compact, but its just an additional line, with the inclusion of sending output to /dev/null so that pushd & popd don't announce the directory change.

Its possible that this does not affect enough users to warrant a change, but it seemed worth the pull request anyway.


bdunn313:
Apologies - meant to say what I tested on:

Tested on: 
(X)Ubuntu 16.04 LTS with bash and zsh
Mac OSX El Capitan bash and zsh

taylorotwell:
You can edit it as needed for your system. Thanks.

bdunn313:
Understood - figured I would pass it along if it was deemed useful.

Thanks @taylorotwell - well done with all things Laravel!

PR #2302, opened 2016-05-07
Add explanation of fifth argument in hasManyThrough method

acacha:
Explanation of fifth argument in hasManyThrough method is missing at docs


acacha:
See also an extra detailed explanation with example from @themsaid at: https://github.com/laravel/framework/issues/13195

PR #2308, opened 2016-05-09
Improved homestead function for daily usage

mabasic:
It no longer changes your current working directory when using it.

Also it solves issue mentioned in this PR: https://github.com/laravel/docs/pull/2299


bdunn313:
Ah - much better solution than what I proposed! Well done @mabasic

PR #2310, opened 2016-05-11
Laravel 5.2 returns 400 status instead of 422

mikefrancis:
Think this may have changed between 5.1 and 5.2 but wasn't updated in the docs, unless I'm doing something wrong!


mikefrancis:
Sorry, ignore this.

PR #2325, opened 2016-05-24
Added token based authentication chapter.

mabasic:
I could not find any documentation for this feature, so I added a very short and minimal instruction on how to get it up and working.

- users table `api_token` column migration
- generating a unique token suggestion code
- route protection using `auth:api`
- three ways of sending the token to authenticate

I am sure that this is enough for anyone to get started with this.


taylorotwell:
I want to improve this for 5.3 and add proper docs then.

PR #2326, opened 2016-05-24
Fix log documentation

kohenkatz:
Although the default log setting is `single`, as mentioned above on the same page, this paragraph said that the default is `daily`. This PR corrects that.

### DIFF
--- a/errors.md
+++ b/errors.md
@@ -116,7 +116,7 @@ The views within this directory should be named to match the HTTP status code th
 <a name="logging"></a>
 ## Logging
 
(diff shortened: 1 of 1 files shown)


kohenkatz:
I also want to add a note about `daily` logging that Monolog [officially recommends](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/RotatingFileHandler.php#L20-L21) using LogRotate instead of the built-in daily rotation, but I wasn't sure if that should be the same PR or another one.

kohenkatz:
@taylorotwell Any comment on my comment above?

PR #2330, opened 2016-05-26
Add docs for optional argument with flatten

outofcontrol:
Not sure if my example code properly expresses what depth accomplishes and most likely needs updating.


taylorotwell:
@JosephSilber do you have an idea of a better example you can PR?

PR #2340, opened 2016-06-01
Fix Command to run Homestead CLI

Kamshak:
The file vendor/bin/homestead is a shell script, not a php file. This executes the command correctly

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -146,7 +146,7 @@ Once Homestead has been installed, use the `make` command to generate the `Vagra
 
 Mac / Linux:
 
-    ./vendor/bin/homestead make
+    sh vendor/bin/homestead make
 
 Windows:
 
--- a/homestead.md
+++ b/homestead.md
@@ -146,7 +146,7 @@ Once Homestead has been installed, use the `make` command to generate the `Vagra
 
 Mac / Linux:
 
-    php vendor/bin/homestead make
+    ./vendor/bin/homestead make
 
 Windows:
 


taylorotwell:
It's not: https://github.com/laravel/homestead/blob/master/homestead

Kamshak:
@taylorotwell Did it use to be? I don't really understand it in this case, this is with a Lavarel 5 project, following per project instructions from the docs:
Image

taylorotwell:
Oh you’re on Windows. I’m actually not sure how it’s supposed to work on Windows. Regardless, that syntax you gave would fail on Mac.

Kamshak:
Hmm, it's Linux actually (Homestead VM on Windows, Linux code 3.19.0-25-generic #26~14.04.1-Ubuntu SMP Fri Jul 24 21:16:20 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux). I don't have a mac to test unfortunately but it [should](http://stackoverflow.com/questions/733824/how-to-run-a-sh-script-in-an-unix-console-mac-terminal) work.

Kamshak:
@taylorotwell: I think the correct way to call the binaries on either System would be to use ./vendor/bin/homestead, see https://github.com/composer/composer/blob/e2c3bc5e8951cc9197c4b8c7d090b5a856628154/src/Composer/Installer/BinaryInstaller.php#L180 for Details on how Composer installs binaries. There is always an executable Proxy, so the current command is incorrect for all Systems (including mac).

PR #2344, opened 2016-06-04
Note about automating upgrades with Laravel Shift

jasonmccreary:


jasonmccreary:
I didn't to repeat the note for every version or break the symmetry of *Upgrading to...* by adding a link. Felt this placement was the best compromise. Also not 100% on the wording. Tried to match town and verbiage. Open to recommendations.

taylorotwell:
One quibble here is that it makes it sounds like Laravel Shift is an official supported service offered by "Laravel" as in "Laravel LLC" the company. We should probably make it clear it's a community maintained product.

jasonmccreary:
Yes, I use *Laravel Shift* to distinguish the products. Glad to call it just *Shift* in this context add whatever wording to denote it's separate. Which do you prefer?

jasonmccreary:
Added *community-driven* to the wording to follow suit with other third-party references in the Upgrade Guide such as the Laravel Collective.

jasonmccreary:
Cool. Is it okay if I make this same PR for the previous version branches?

PR #2346, opened 2016-06-05
Set custom values for Authentication Throttling

whoan:


taylorotwell:
I would rather the default values just be sane then having people customize them all the time.

PR #2352, opened 2016-06-06
Note about automating upgrades with Laravel Shift

jasonmccreary:


jasonmccreary:
Did you not want to backfill the documentation for the previous versions? These are still the top links in search results. Just want to ensure we're not missing that portion of the audience.

PR #2357, opened 2016-06-08
Added `Vagrant::HostsUpdater` note

tillkruss:
Refers to https://github.com/laravel/homestead/pull/328


taylorotwell:
I don't really want to support issues with it if people have problems since I would be officially recommending it.

PR #2360, opened 2016-06-09
Fix proper namespacing

samuelsan:


taylorotwell:
You should just import it into your class.

PR #2361, opened 2016-06-10
Add info on `size` validation rule

samundra:
* when `size` gets array value it checks against count of an array


taylorotwell:
Wrong branch. I will fix it for you.

samundra:
:+1: sure

PR #2370, opened 2016-06-17
Empty lines also needed here

manniL:
Another two empty lines to comply to CodeMark specs


GrahamCampbell:
Thanks. Does it not parse ok anyway?

manniL:
Nope, it doesn't display the MD properly when using my Firefox MD plugin/extension. Also my IDE doesn't highlight the MD if the two empty lines are not there.

parsing

PR #2382, opened 2016-06-25
Updated composer bin path

kimett:
Path seems to have changed according to the Homestead installation I did today.
It pointed out the updated path after I did `composer global require "laravel/installer"`

### DIFF
--- a/installation.md
+++ b/installation.md
@@ -34,7 +34,7 @@ First, download the Laravel installer using Composer:
 
     composer global require "laravel/installer"
 
-Make sure to place the `~/.composer/vendor/bin` directory (or the equivalent directory for your OS) in your PATH so the `laravel` executable can be located by your system.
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
> It pointed out the updated path after I did composer global require "laravel/laravel"

We definitely don't recommend that you globally install `laravel/laravel`.

GrahamCampbell:
This path doesn't seem correct. On most systems it will be `~/.composer/...`. I don't understand this PR?

kimett:
global was of "laravel/installer" my fault while typing. Commands taken from the installation doc itself.

The outputted path and path of the bin was what I got from composers output.
Virtualbox and vagrant was downloaded and installed today which is the same day the homestead box was set up. Why it gives a different path is unknown to me so far.

taylorotwell:
I also have mine located in ~/.composer

taylorotwell:
If someone has more info on this though I would definitely be interested.

PR #2384, opened 2016-06-27
Adding new view file extensions

kfriend:


views.md
```diff
@@ -200,3 +200,43 @@ The `composer` method accepts the `*` character as a wildcard, allowing you to a
 View **creators** are very similar to view composers; however, they are fired immediately when the view is instantiated instead of waiting until the view is about to render. To register a view creator, use the `creator` method:
 
     view()->creator('profile', 'App\Http\ViewCreators\ProfileCreator');
+
+### Adding Extensions and Assigning a View Engine
+
+Laravel supports both `.blade.php` and `.php` view file extensions by default. Additional extensions can be added using the `view()->addExtension('extension', 'engine)` method. This can be used for including view files via Laravel's standard means, while retaining a file extension that is more useful to your editor/IDE. You must supply the desired extension and the view engine to handle it. Out of the box, Laravel supports the `blade` and `php` rendering engines.
```
arcanedev-maroc: Missing a `'`:
```php
view()->addExtension('extension', 'engine)
```
To:
```php
view()->addExtension('extension', 'engine')
```

kfriend: thanks @arcanedev-maroc! 👍

arcanedev-maroc: You're welcome :+1:

taylorotwell:
Hmm, I'm not sure I really like this examples as documentation examples. Are there any other more practical examples?

kfriend:
Laravel doesn't ship with a "plain" engine, so examples might seem more contrived than they actually are, especially without an explanation.

The end goal of adding extensions are:
- Allow loading non .php and .blade.php extensions via common Laravel methods (view('...'), View::make('...'), @include('...')). 
- Allow using a more semantic file extension. .txt for a plain text file is more clear about intension than .php or .blade.php
- Retain automatic syntax highlighting within your editor/IDE

Here's a simplified real world example that you might find more useful. If you approve, I'll format everything and place in the PR. This is really just an explanation for you @taylorotwell.

```
// Add support for loading Vue templates
view()->addExtension('vue', 'php');
```

Then elsewhere, in another view file (e.g. a layout file)

```
...
<div id="navComponent">
    @include('components.nav')
</div>
...
```

Again, there's no "plain" engine, so the vue extension must be handled by the php engine. The goal is to allow including Vue templates and have your editor/IDE (Sublime in my case) automatically use the correct syntax highlighting for the file. .blade.php or .php will mean every time a Vue template file is opened, you have to change the current syntax.

Markdown files are also a valid example. Adding support for .md extension will allow editing the file with Markdown syntax used automatically.

```
view()->addExtension('md', 'php');

// Elsewhere in a controller
...
$markdown = view('pages.about');
return MarkdownParser($markdown)->toHtml();
...
```

Let me know any thoughts.

PR #2385, opened 2016-06-28
Explain there will be .env.exmaple for placeholder

setkyar:


GrahamCampbell:
The sentence straight after says `rename`, not `copy` though?

PR #2386, opened 2016-06-28
No need of type hinting on explicit binding

batusa:
When an explicit binding is registered, there's no need of type hinting.

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -363,7 +363,7 @@ To register an explicit binding, use the router's `model` method to specify the 
 
 Next, define a route that contains a `{user}` parameter:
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
There is no reason not to have it though. It's there to help document what type that thing is to anyone reading the docs.

batusa:
By reading the docs, in the first example it shows implicit binding, where just by type hinting a model you get it loaded.

In the second example, it shows explicit binding, where one goes to the route provider and binds it. By removing the type hinting in the closure (as I did), it makes more clear that one doesn't need to type hint it anymore.

Well, that's just a suggestion. It's up to you guys.

PR #2398, opened 2016-07-06
Users want to remove 'public' from url

gauravmak:
Like me, so many users are asking questions in Stackoverflow and Laracasts to open the site without 'public' in the url. It would be good to specify details in this document. Am no expert with all this so your suggestions for proper instructions would be helpful.


taylorotwell:
I added a note regarding this in the documentation.

gauravmak:
Okay!

PR #2402, opened 2016-07-08
added timezone setting to homestead.md

BartHuis:
i tried multiple ways to set the timezone by adding rows to diffrent files, but they made the script giving errors or dont do anything at all, untill i found this much simpeler way.


taylorotwell:
I think this is something that more belongs in a blog post or something. Thanks.

PR #2404, opened 2016-07-09
Add a missing space to block-style control structures

Ingramz:
Rest of similar control structures in the file have a space between the directive name and opening parentheses. This will correct the inconsistency.


GrahamCampbell:
Do they?

PR #2411, opened 2016-07-14
Correct the doc when passing an array to the where method

darronz:


taylorotwell:
The original example works it just needs an extra parameter. I will fix it.

PR #2416, opened 2016-07-18
Added Requiring Elixer in Gulp heading

bjrnblm:
This step seems to be skipped until Writing Elixir Extensions, which is an advanced topic. 
I think it is more beginner friendly to put it higher in the docs.

### DIFF
--- a/elixir.md
+++ b/elixir.md
@@ -69,7 +69,15 @@ If you are developing on a Windows system or you are running your VM on a Window
 <a name="running-elixir"></a>
 ## Running Elixir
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think we don't mention this because it's already in the file, which would be confusing.

bjrnblm:
Ah, you are totally right! The confusion came from updating an old project with already a gulpfile.js present, and I just copied bits and pieces from the docs to use elixir.

I just installed a fresh laravel and the gulpfile looks fine.

PR #2417, opened 2016-07-19
Different PATH variable for Laravel Installer 1.3.3?

fordarnold:
Not sure if my system (Ubuntu 14.04 LTS), or if Laravel Installer 1.3.3 is the one which installs the composer configurations in `~/.config/composer/vendor/bin/` and **NOT** `~/.composer/vendor/bin`.


taylorotwell:
We already say "or the equivalent directory on your OS".

PR #2419, opened 2016-07-21
Document the @php function

daankuijsten:
The documentation for this function seemed missing


taylorotwell:
I'll document this. I don't think it deserves its own entire section.

PR #2424, opened 2016-07-26
[5.2] CSRF protection is for session user, not necessarily authenticated user.

JesseLeite:
CSRF protection is for session user, not necessarily the authenticated user.  The docs are a bit misleading as CSRF protection is important, even without a user authentication layer.

Also see #2425 for same changes to 5.3 docs.

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -254,9 +254,9 @@ You may also use the `prefix` parameter to specify common parameters for your gr
 <a name="csrf-introduction"></a>
 ### Introduction
 
-Laravel makes it easy to protect your application from [cross-site request forgery](http://en.wikipedia.org/wiki/Cross-site_request_forgery) (CSRF) attacks. Cross-site request forgeries are a type of malicious exploit whereby unauthorized commands are performed on behalf of an authenticated user.
+Laravel makes it easy to protect your application from [cross-site request forgery](http://en.wikipedia.org/wiki/Cross-site_request_forgery) (CSRF) attacks. Cross-site request forgeries are a type of malicious exploit whereby unauthorized commands are performed on behalf of another user.
 
-Laravel automatically generates a CSRF "token" for each active user session managed by the application. This token is used to verify that the authenticated user is the one actually making the requests to the application.
+Laravel automatically generates a CSRF "token" for each active user session managed by the application. This token is used to verify that the intended user is the one actually making the requests to the application.
 
 Anytime you define a HTML form in your application, you should include a hidden CSRF token field in the form so that the CSRF protection middleware will be able to validate the request. To generate a hidden input field `_token` containing the CSRF token, you may use the `csrf_field` helper function:
 


taylorotwell:
I don't *think* I totally agree with your change here. A CSRF attack would try to perform something on the authenticate user's account (for that browser session).

JesseLeite:
Imagine a user is not authenticated, sitting on any public form (maybe a contact form, or a login form, where there is no authenticated user in session) for over 120 minutes (the default session lifetime) without refreshing the browser.  If they try to submit that form without refreshing first, they'll get a token mismatch exception.

>A CSRF attack would try to perform something on the authenticate user's account (for that browser session).

While true that CSRF protection is often intended to protect an authenticated user, the middleware is enabled and enforced by default, whether the developer has a need to scaffold out user authentication or not.

If a newer Laravel developer is debugging token mismatch exceptions, and reads the documentation, it leads them down the path of "authentication", rather than session lifetime.  I see token mismatch questions on Larachat quite a bit.  It's a confusing subject to newer developers I think.

stayallive:
This resulted in a few unexplained errors in our apps too for a while (untill we noticed that the user was null on all requests, thank you Sentry).

Maybe it would be better to add a tip that an expired session can also generate a token mismatch. To me this change does not clarify this to be honest.

sisve:
Imagine that I know your router's ip. (Some even come with resolvable internal domain names...)

Scenario 1: You are authenticated but the router does not use CSRF.
Scenario 2: Router does not use authentication but accepts all commands coming from the internal network.

```
<form method="post" action="http://192.168.0.1/set-dns-server" id="evil_form" target="evil_iframe" >
    <input type="hidden" name="ip" value="1.2.3.4">
</form>
<iframe id="evil_iframe" style="display: none"></iframe>
<script type="text/javascript">
    document.getElementById("evil_form").submit();
</script>
```

Result 1: You are authenticated and the router accepts the command.
Result 2: You are anonymous, but the router sees the request coming from an internal network address and accepts the command.

Both scenarios ended up changing your dns settings in your router without you knowing it. The only thing that can protect the user in these cases are 1) referral checks (the posted form was hosted on the router url) or 2) CSRF checks (the posted form was generated by the router).

My point: CSRF is needed for anonymous users too.

geggleto:
I believe @JesseLeite . 

Consider CSRF on a authentication form. 
You would not be "authenticated", on the system as a whole.

 The reason for putting CSRF on your login form would be to prevent offline login attacks [saving file to desktop, launching a huge amount of logins perhaps through a CLI.... mostly to save on bandwidth etc].
^ I have seen this in real life ;) It's weird and an edge case, but I think a valid point that you would want to cover.

JesseLeite:
A non-authenticated guest user is still a user, and token mismatch exceptions can still affect those guest users without authentication.

I see Taylor’s point, maybe it’s most commonly used to protect authenticated users, but it also protects guest users by session.  The CSRF middleware is applied and enforced by default on guest POST requests, contact forms, etc.  For new Laravel users debugging token mismatch exceptions in those situations, the current description in the docs seem too focused on ‘authentication’ to help them.

@stayallive, maybe the wording changes suggested by my PR aren't optimal, but my point still stands.

PR #2425, opened 2016-07-26
[5.3] CSRF protection is for session user, not necessarily authenticated user.

JesseLeite:
CSRF protection is for session user, not necessarily the authenticated user.  The docs are a bit misleading as CSRF protection is important, even without a user authentication layer.

Also see #2424 for same changes to 5.2 docs.


taylorotwell:
Closing this one since 5.2 merges will be merged forward.

PR #2433, opened 2016-07-27
small typo fix

friedrichschoene:


taylorotwell:
This isn't really a typo, its just to make it clear that they are probably going to be IDs that are injected in.

friedrichschoene:
Either way, the variable names should be consistent.

Issue #2428, opened 2016-07-27
Collection::slice() has no 3 parameters option.


kt81:
in: https://laravel.com/docs/5.2/collections#method-slice

Collection::slice() is introduced as it can be passed 3 parameters to preserve original keys or not.
But actually this method does not use 3rd parameter and always preserves original keys.

I found the related issue and PR below. Since these conversation, It seems the document should be corrected.
https://github.com/laravel/framework/issues/13692
https://github.com/laravel/framework/pull/13698

jhoff:
Pull request created to update docs starting at 5.2 onward:

> The returned slice will preserve keys by default. If you do not wish to preserve the original keys, you can use the `values` method to reindex them.

kt81:
@jhoff Thank you! 😃

PR #2437, opened 2016-07-28
Slice preserves keys by default

jhoff:
Issue raised here https://github.com/laravel/docs/issues/2428

The documentation for [`Collection::slice`](https://laravel.com/docs/master/collections#method-slice) currently states the following:

> The returned slice will have new, numerically indexed keys. If you wish to preserve the original keys, pass true as the third argument to the method.

Not only is this false in that the current behavior preserves keys by default, but slice doesn't even support a 3rd argument.

This separate pull request introduces the third argument and keeps the current behavior unchanged: https://github.com/laravel/framework/pull/14527

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -993,7 +993,7 @@ If you would like to limit the size of the returned slice, pass the desired size
 
     // [5, 6]
 
-The returned slice will have new, numerically indexed keys. If you wish to preserve the original keys, pass `true` as the third argument to the method.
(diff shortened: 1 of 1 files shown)


jhoff:
Per https://github.com/laravel/framework/pull/14527, boolean options are going away and the default behavior is to preserve keys. The documentation has been updated accordingly.

jhoff:
Closing this to rebase and create new pull request on 5.1

PR #2443, opened 2016-07-31
Added note about Valet not working out of the box

DCzajkowski:
Many people (including myself) didn't get Valet working out of the box. It was caused by Apache already running in the background. Simple restart of the Apache and Valet using `sudo apachectl stop; valet restart` worked.

https://laracasts.com/discuss/channels/laravel/valet-v112-update-just-keep-getting-the-it-works#reply-165001


taylorotwell:
I think this can be worded differently. I will look into it.

taylorotwell:
Ah, just noted we already have a big red note about this at the very top of installation.

PR #2447, opened 2016-08-02
Missing =

marcusmoore:


jhoff:
=== ?

marcusmoore:
That would work. I was actually copying the [5.2](https://laravel.com/docs/5.2/authorization#defining-abilities) section that had `==` but I see that the [5.1](https://laravel.com/docs/5.1/authorization#defining-abilities) section had `===`.

michaeldyrynda:
The strict comparison isn't going to work anyway, because of the foreign value being a string. The primary key will be returned [as an int](https://github.com/illuminate/database/blob/master/Eloquent/Model.php#L2734) >= 5.2, but the foreign key is still a string unless you explicitly cast it in the model.

PR #2453, opened 2016-08-04
[5.3] Change routes.php to routes/web.php

adrianyg7:
`routes.php` file no longer exists, I guesss `routes/web.php` should be good, or simply `web.php`.


adrianyg7:
Just realised that Upgrade Guide doesn't mention about the routes.php being moved to a top level in the app because it's not really necessary.

PR #2460, opened 2016-08-06
Specification about count column name

gauravmak:
If relation is 'contactLists', it returns 'contact_lists_count'


taylorotwell:
The example already shows it in snake case.

gauravmak:
@taylorotwell I meant to specify that if `{relation}` is in 'came case', it will be converted to **snake case** as well. Not so major thing, anyway.

PR #2469, opened 2016-08-09
[4.2] Move test seeding text below heading

ntzm:


GrahamCampbell:
I think text was intentionally above to lead to this heading?

ntzm:
Oh right, #2466 was approved yesterday, for the exact same change?

PR #2472, opened 2016-08-09
[5.2] Add Cashier currency configuration

josebailo:
After a while searching how to configure the Cashier currency I found a solution that I think must be in the docs.

I tried to explain it the easiest way possible.


taylorotwell:
Can you send this to the 5.3 branch?

josebailo:
Sure, here #2495 you have it!

PR #2479, opened 2016-08-10
add abort_if and abort_unless

spyric:


helpers.md
```diff
@@ -715,6 +717,28 @@ The `abort` function throws a HTTP exception which will be rendered by the excep
...
 
     abort(401, 'Unauthorized.');
+    
+<a name="method-abort-if"></a>
+#### `abort_if()` {#collection-method}
+
+The `abort_if` function is a syntax sugar for next expression:
+
+    if($condition) {
+    	abort(401);
```
GrahamCampbell: If we're including 401 here, maybe change the above bit to include it too?

> The `abort_if(401)` function is a syntax sugar for next expression:

spyric: I can rewrite it to

The `abort_if` function is a syntax sugar. `abort_if($condition, 401)` will become to:
```
if($condition) {
    abort(401);
}
```

is it ok?

GrahamCampbell: looks good 👍

PR #2486, opened 2016-08-12
Mention info on session::now

BenCavens:


taylorotwell:
I honestly think this is sort of a weird way to use the session and don't really want to advertise it.

BenCavens:
I agree. 
Possible usage is to override flashed session data. For me, it turned out to be of good use for populating site notifications, which already pulled their value from session.

PR #2487, opened 2016-08-12
[5.3] Fixed missing = symbol

jmarcher:
Missing = symbol at Mail documentation 
and Broadcasting documentation, as I tried Echo it asked me for a pusherKey and not a key parameter, maybe was just me.

### DIFF
--- a/broadcasting.md
+++ b/broadcasting.md
@@ -64,7 +64,7 @@ When using Pusher and [Laravel Echo](#installing-laravel-echo), you should speci
 
     window.Echo = new Echo({
         broadcaster: 'pusher',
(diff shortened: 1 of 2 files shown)


taylorotwell:
The echo change is not right. I will fix the other issue. Thanks. Your echo copy might be out of date.

jmarcher:
weird, but I just checked my echo.js in my node_modules and it says:

```
createClass(PusherConnector, [{
        key: 'connect',
        value: function connect() {
            this.pusher = new Pusher(this.options.pusherKey, this.options);
        }
    }
```

PR #2494, opened 2016-08-13
Temporary change version to dev-develop

mustafaaloko:
Not sure if this would make sense or not until official release.


taylorotwell:
It will make sense when official release happens.

PR #2498, opened 2016-08-13
add note on cascaded configuration

samundra:
The usage of one config variable inside another config variable results in intermittent behaviour.


GrahamCampbell:
👎

GrahamCampbell:
You should never read config from inside another config file.

samundra:
@GrahamCampbell I realized that we shouldn't be reading config from config. So, the PR here informs people regarding config usage inside another config. I didn't see it documented though I remember there was some discussion on it. Please close this PR if this was already documented.

Thanks for the update on it.

PR #2500, opened 2016-08-14
Warn readers about false positive cache results

tillkruss:
Using PhpRedis instead of Predis with Laravel’s default `RedisServiceProvider` will result in false-positive cache results across the framework, because PhpRedis returns `false` instead of `null` if a key does not exist.

This PR warns the user about this unexpected behaviour and links to a drop-in replacement.


taylorotwell:
Would it better to drive to solve this at the framework level so that it supports PhpRedis?

taylorotwell:
I think I would rather try to solve this as mentioned above.

tillkruss:
Great, love hearing that! I'll submit a patch.

PR #2501, opened 2016-08-14
Add data and object helper documentation

mtdavidson:
- Document data_fill, data_get, data_set and object_get


taylorotwell:
I'm not sure there is ever really a reason to use the object_get function rather than data_get. Also there are some typos here.

PR #2506, opened 2016-08-15
LTS Release Verbiage Update

iolson:
Added a line for LTS releases, that some clients have when asking how often are LTS releases released.


taylorotwell:
I have no idea if a new LTS will be released at the end of the 5.1 cycle.

PR #2511, opened 2016-08-16
add clerify for auth in constructor

ac1982:
You can not read authenticated user in constructor.


taylorotwell:
This is not the best place for this.

PR #2521, opened 2016-08-18
[5.2] Remove the Bus facade

nesk:
This facade [is missing](https://github.com/laravel/laravel/blob/982e769d2aaa9b8a4121ac933ee1360246a43a3f/config/app.php#L179-L212) in the application, so no need to document it.


taylorotwell:
It still exists and can be added if desired.

PR #2528, opened 2016-08-19
Updated Scout documentation, PHP 7.0 system requirement

wcjonathan:
When installing Scout, i received the error

`Cannot bind an instance to a static closure`

Updating PHP version to 7 fixed this issue and i was able to use the Scout library.

Updated documentation to reflect this.

### DIFF
--- a/scout.md
+++ b/scout.md
@@ -2,6 +2,7 @@
 
 - [Introduction](#introduction)
 - [Installation](#installation)
+    - [Server Requirements](#server-requirements)
     - [Queueing](#queueing)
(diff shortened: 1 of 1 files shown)


taylorotwell:
So, probably we should just fix the PHP 7 requirement in Scout.

taylorotwell:
Please provide the stack trace of the error.

taylorotwell:
I have tagged a new 0.1.1 version that I think should address the issue.

wcjonathan:
Brilliant! Thanks

PR #2531, opened 2016-08-19
Fix code example for multiple middleware on route.

samrap:
Docs say, "Use an array to assign multiple middleware to the route:" but the code example does not use an array.


taylorotwell:
You actually don't have to use an array. I'll update the docs.

samrap:
Cool, I knew this was the case but didn't know if you wanted to update the code example or the text. Thanks!

PR #2535, opened 2016-08-20
Update helpers.md

Shipu:
add str_replace_first helper in documentation


taylorotwell:
Didn't even know this existed in the core and not sure that it should. It is not used by the framework.

PR #2538, opened 2016-08-22
Update releases.md

BartHuis:
wrong link


BartHuis:
not only in master, also in 5.3 ;)

PR #2540, opened 2016-08-22
Add more information about calling policy methods with class names

clemir:


taylorotwell:
This is not new to 5.3.

PR #2541, opened 2016-08-22
Use the proper namespace to Crypt facade.

ndeet:
use Crypt; does not work, as of issue (laravel/lumen-framework#307) I learned that Crypt is only an alias, but that does not work out of the box so the full namespace to Crypt facade makes sense here?

### DIFF
--- a/encryption.md
+++ b/encryption.md
@@ -21,7 +21,7 @@ For example, we may use the `encrypt` method to encrypt a secret and store it on
 
     namespace App\Http\Controllers;
 
-    use Crypt;
+    use Illuminate\Support\Facades\Crypt;
     use App\User;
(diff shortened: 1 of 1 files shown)


taylorotwell:
If it's a Lumen issue I don't think it would go to the Laravel docs.

ndeet:
Hi @taylorotwell sorry for the confusion, BUT, I found the hint in the lumen issue queue, the problem is the same for both laravel and lumen as both use the same framework classes and both are missing an alias for "Crypt" so the full facade namespace is needed?

ndeet:
@taylorotwell please look at my last comment again, still true for laravel docs. If you do not want to merge for whatever reason, please let me know, I will delete the PR.

PR #2543, opened 2016-08-23
Update releases.md

BartHuis:
since 5.3 became master, 5.3 links didn't work anymore on the docs view


taylorotwell:
They are correct, and they will work when 5.3 is actually released.

BartHuis:
ah i see, didn't know the rc1 to real release would be that fast, thought it would stay with wrong links for another few weeks with rc1, but this time was very short ;)

PR #2551, opened 2016-08-23
Add information about the missing logger attribute in the base exception handler class

ThibaudDauce:
The LoggerInterface is now missing from the base exception handler class.

See laravel/framework@a92651e

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -246,6 +246,10 @@ The base exception handler class now requires a `Illuminate\Container\Container`
 
     parent::__construct(app());
 
+#### Logger
+
+The base exception handler `\Psr\Log\LoggerInterface` attribute has been removed. You should use `$this->container['log']`, `app('log')` or the facade `Log::` instead of `$this->log`.
+
 ### Middleware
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not sure that this was ever really a documented feature.

ThibaudDauce:
Maybe it wasn't documented but it was usuble and I found it really easily with PHPStorm inspection for example. For me it was a breaking change (and hopefully I have a lot of PHPUnit tests but for others it may break their application when an exception occures…).

conorsmith:
I encountered this issue after upgrading. While it wasn't a documented feature, it is not wild to expect a user to access a protected property of a framework class they've extended (especially when that class is actually extended by default on installation). The fact that it wasn't private would lead the user to believe this is a safe thing to do.

In this particular case I'd argue that it's important enough to be in the upgrade guide. It breaks reporting of all uncaught exceptions in an application and prevents the application from rendering it as an HTTP response. This could lead to errors being unnoticed in a production environment.

PR #2552, opened 2016-08-23
Provider boot() method and RouteServiceProvider changes.

webtechnick:
Upgrading to 5.3 was a bit more than what the upgrade guide detailed. I also had to copy over the providers in github as the boot definition changed and threw errors. As I was updating the RouteServiceProvider it setup new route locations to routes/web.php routes/api.php, so had to update that as well. Moving Http/routes.php to routes/web.php and creating a new routes/api.php from github default.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -343,7 +343,17 @@ If you are queueing jobs using this syntax, Eloquent models will no longer be au
 
     Queue::push(new ClassName);
 
+### Application Service Provider changes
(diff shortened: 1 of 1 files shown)


taylorotwell:
You don't really have to change your route file location but agree the other things need to be mentioned in docs.

taylorotwell:
I've updated docs for this.

PR #2555, opened 2016-08-24
Make S in Bugsnag lowercase

dwightwatson:
Looks like the brand name for Bugsnag has the S lowercase, so I've updated it in the docs.


GrahamCampbell:
👍  Yep, I can confirm this is correct. We get all kinds of funny spellings besides this. :)

PR #2557, opened 2016-08-24
[5.3] Add composer.json changes to Upgrade Guide

tzurbaev:


GrahamCampbell:
👎

tzurbaev:
@GrahamCampbell  Am I missing something? composer.json update is not required to upgrade from 5.2?

GrahamCampbell:
We don't do this on any of the guides because it's already obvious. Infact, the guides specifically describe what to do after you change the version.

tzurbaev:
But [Upgrading To 5.2.0 from 5.1](https://laravel.com/docs/5.3/upgrade#upgrade-5.2.0) contains this section :)

But ok, I'll close this, thanks for reponse.

GrahamCampbell:
Does it? Hmmm, maybe this is fine then.

GrahamCampbell:
We should probably point out people should bump their symfony deps in require-dev too.

PR #2568, opened 2016-08-24
update collections.md each()

timbroder:
with info about skipping an each() iteration


taylorotwell:
This seems true of basically any callback? You can always return out of a function so I'm not sure this should be documented.

timbroder:
@taylorotwell true. I guess I just hadn't thought through that yes, if I return from a callback, the next one in the collection iteration will automatically fire. I had to google a bit and figured I could save someone else the trouble

PR #2571, opened 2016-08-25
Add Laravel TestTools Chrome extension to Application Testing section

mpociot:
The extension is already used by more than 4.000 users per week. Adding it to the official documentation will make even more people get started with application testing (which is good, right?)


taylorotwell:
No offense intended, but I don't really want to include too many community tools in the documentation because it could create awkward situations later. For example, what if another person writes a similar but slightly different tool and also wants to be included in the docs? Then I have to make decision to either remove one or have a list of community tools, etc.

mpociot:
Not feeling offended at all. Valid reasons.

PR #2577, opened 2016-08-26
Add --ignore-platform-reqs to homestead require

TheoKouzelis:
In the past I have had to add the "--ignore-platform-reqs" to get composer to ignore my local PHP setup to bring down the Homestead package.


taylorotwell:
I don't think this be generally recommended. It could let people install Homestead on a machine that doesn't have the right PHP version.

TheoKouzelis:
Ah yes, sorry, I didn't check to see Homesteads dependencies.

PR #2578, opened 2016-08-27
Typo in Broadcasting authorization

stevethomas:
When authorizing a channel, I can't think of any good reason why you would to create a new model from the BroadcastServiceProvider. This would create a situation where every request is authorized and result in junk data.

Pretty sure this was a typo.


GrahamCampbell:
Nope, this was not a typo.

GrahamCampbell:
Previously, it just create a model in memory, but now it fatal errors.

GrahamCampbell:
At no point does this temp model actually get written to the db, so the existing code is fine. :)

taylorotwell:
As noted, existing code is fine.

PR #2582, opened 2016-08-27
Add better explanation how to fix new routes directory structure

loekiedepo:
I struggled for a few minutes how to make sure my app knew where my new routes were located. Tried to explain it better in the docs, so other people can go straight ahead in upgrading instead of searching around in their app for what is throwing that 404 all the time.


taylorotwell:
There is no need to change the location of your routes files when updating.

PR #2596, opened 2016-08-30
Update eloquent-serialization.md

Quezler:
Clarify why $appends does not work in tinker when just using e.g. `all()` or `first()`

### DIFF
--- a/eloquent-serialization.md
+++ b/eloquent-serialization.md
@@ -145,3 +145,5 @@ After creating the accessor, add the attribute name to the `appends` property on
     }
 
 Once the attribute has been added to the `appends` list, it will be included in both the model's array and JSON representations. Attributes in the `appends` array will also respect the `visible` and `hidden` settings configured on the model.
+
+> {note} When using `php artisan tinker` to return a modal without `toArray()` or `toJson()` appends does not show in its output.


Quezler:
### Test case
```
<?php

namespace App;

use Illuminate\Database\Eloquent\Model;

class FooBar extends Model
{
    protected $appends = ['foo'];

    public function getFooAttribute() {
        return 'bar';
    }
}

```

```
>>> App\FooBar::first()
=> App\FooBar {#701
     id: "1",
     created_at: "2016-08-30 08:06:28",
     updated_at: "2016-08-30 08:06:28",
   }
```

```
>>> App\FooBar::all()
=> Illuminate\Database\Eloquent\Collection {#697
     all: [
       App\FooBar {#702
         id: "1",
         created_at: "2016-08-30 08:06:28",
         updated_at: "2016-08-30 08:06:28",
       },
     ],
   }
```

```
>>> App\FooBar::first()->toArray()
=> [
     "id" => 1,
     "created_at" => "2016-08-30 08:06:28",
     "updated_at" => "2016-08-30 08:06:28",
     "foo" => "bar",
   ]
```

```
>>> App\FooBar::first()->toJson()
=> "{"id":1,"created_at":"2016-08-30 08:06:28","updated_at":"2016-08-30 08:06:28","foo":"bar"}"
```

## Conclusion

I feel like this should be added to the docs since the behavior is slightly non obvious.
Personally have spend 20 minutes searching the internet on why it did not work, without results.

And indeed, if you read the Laravel docs to the letter, it is intended behaviour.
But since tinker returns the output after every command, it might be confusing.

🎩

taylorotwell:
I would be more interested in knowing why.

Quezler:
Knowing why what?

PR #2597, opened 2016-08-30
Added 5.3 upgrade option for Route::resource nested in a Route::group…

anzeznidarsic:
… with a prefix

If a user had a 2 resource routes with same named controllers in 2 different Route::groups this breaks in L5.3. It was before mentioned that the solution is using names option on Route::resource, but that means we have to specify each method separately. If we simply wish to prefix entire route resource we can use 'as' option (this feature is not documented, but it is working in L5.3).


anzeznidarsic:
In 5.2 we used:
`Route::group(['prefix' => 'admin'], function () {
  Route::resource('maps', 'MapsController');
});`

We had route names admin.maps.create, admin.maps.store, etc.

To achieve the same in 5.3 we have to add 'as' option in Route::resource:
`Route::resource('maps', ['as' => 'admin', 'uses' => 'MapsController']);`

PR #2598, opened 2016-08-30
Add code example for jobs table migration

redgluten:
Add code example for jobs table migration regarding changes made to queues’ database driver


taylorotwell:
I'm going to hold off on this because it would be a different actoin in your app is still in dev.

redgluten:
I don’t understand why, this passage is only relevant if the migration has been made before right?

PR #2605, opened 2016-09-01
Remove Invalid anchor link for Multiple DB Connections

alcis-dev:
Does not currently have a section in v5.3+ docs for multiple db connections on this page


alcis-dev:
I would note that it exists already further up the page

PR #2611, opened 2016-09-03
Parameters out of order for array_where

JackEllis:
$key before $value. It's fine in 5.2 backwards just not 5.3.


taylorotwell:
Did you test it? They are not out of order.

PR #2612, opened 2016-09-03
Use consistent naming

tanarurkerem:
This naming and order is used In the example code below.


taylorotwell:
closure shouldn't be singular. Also callback is commonly used to refer to Closures / anonymous functions passed into a method.

PR #2613, opened 2016-09-03
Add a note so the reader can find the install instructions for Laravel

bobbybouwmann:
This PR comes out of this thread on Laracasts: https://laracasts.com/discuss/channels/laravel/laravel-new-does-not-work-after-installing-valet-command-not-found

Let me know if you want a PR for the other versions of Laravel as well :)


taylorotwell:
This seems too basic to include.

bobbybouwmann:
Fair enough ;)

PR #2614, opened 2016-09-03
Add notice about storage engines and transactions

Thijmen:
Some default MySQL-setups have `MyISAM` as the default storage engine. It took me quite some time to find out that `MyISAM` is not supported with DatabaseTransactions.


taylorotwell:
I don't like two notes close together. They don't render correctly.

PR #2617, opened 2016-09-05
Update authorization.md

naneri:
The last bracket should be removed


brayniverse:
You should probably remove the semicolon as well.

PR #2619, opened 2016-09-05
For laravel Passport: add note about openssl being required

themsaid:


taylorotwell:
OpenSSL is in the laravel/framework composer.json require so that should alert them.

PR #2621, opened 2016-09-06
Final step for upgrade

gauravmak:
Obvious step but lets avoid confusion.


taylorotwell:
It seems weird to put this in the middle of the instructions.

PR #2625, opened 2016-09-06
add info to upgrade Form Requests

pakogn:


taylorotwell:
This isn't really part of the upgrade necessarily. If you are upgrading you actually do have a base Request in your application so you can still extend that.

PR #2631, opened 2016-09-07
small addition to csrf & mysql db

ctf0:
1- add Vue.js settings to x-csrf to facilitate using Vue with laravel
for newbies.

2- add a small doc for strict mode true in mysql as an extra help point
instead of only relaying on
`https://laravel.com/docs/5.3/upgrade#upgrade-5.2.0 MySQL Dates` point.


taylorotwell:
We already set the CSRF token if you are using Vue in 5.3.

PR #2632, opened 2016-09-07
Add `limit` method to the query builder docs.

ericbarnes:
I found this missing when searching for limiting to a specified number of records.


taylorotwell:
I merged this but I think you pushed a commit at the same tiem. Anyways, its merged :)

PR #2637, opened 2016-09-08
Updates SQS disclaimer.

wuori:
This update adds some context to the SQS disclaimer without adding too much more text. Since SQS handles queues in such a different manner, some further description could be helpful. I originally spotted the SQS disclaimer while re-reading the docs and the callout was definitely useful in the fact that I could disregard my failed-jobs issues, but left me with nowhere to go. Not sure how Laravel likes to use vendor URLs in documentation, so please close this PR if that's not acceptable, or I can re-write to use link-less descriptions.


taylorotwell:
Wouldn't you just use the Laravel failed jobs facilities? Why would you need to use the Dead Letter stuff?

wuori:
Whoops! Sorry, late-night brain mode. I was also dealing with some failed-job issues with SQS/Mongo and I linked them together in my mind. 

I updated the commit to remove the Dead Letter line and ONLY reference the Default Visibility Timeout if you still think it's worth updating the docs with. 

I'll double-check any future PR's to avoid any mishaps like this again.

PR #2644, opened 2016-09-09
Add missing import for Route facade

aaemnnosttv:
Small addition to the code snippet in the Introduction to Facades.


taylorotwell:
This isn't typically necessary in your routes file since we use un-aliased by default there.

aaemnnosttv:
Perhaps a note should be added about that then since that seems to be a bit of an implicit context to the example?  That or maybe modify the example to only use the Cache facade?

I find this to be a bit confusing myself, but I see now that it is aliased.
image

Why import Cache then in the example when it is also aliased?  It seems strange to have that inconsistency.

Aliases only appear to be mentioned in the upgrade guide.

PR #2646, opened 2016-09-09
As behaves differently than the previous prefix

fernandobandeira:
Related to laravel/internals#181

Since laravel/framework#15375 wasn't merged, now ```as``` behaves differently than what ```prefix``` did before.

It requires an ```.``` at the end of the name in order to it be equal to the previous ```prefix``` naming.

Using ```prefix``` you would get ```admin.*``` meanwhile ```as``` generates ```admin*```

Also note that ```as``` doesn't append the name to the url, so you need to use both to maintain the behavior., not just replace ```prefix``` with ```as``` the doc's currently state's.


taylorotwell:
Thanks

PR #2648, opened 2016-09-09
Update controllers.md with extra middleware examples

malhal:
I didn't know controller middleware could be add using its class name so added it to the docs. And I also added the new middleware closure feature from [#15080](https://github.com/laravel/framework/pull/15080)


taylorotwell:
Just documented this in a slightly different way.

PR #2652, opened 2016-09-12
added command for directory configuration

rjcrystal:
adding this command will help new users set correct directory permissions who are not using homestead configure laravel installation.

### DIFF
--- a/installation.md
+++ b/installation.md
@@ -67,7 +67,7 @@ All of the configuration files for the Laravel framework are stored in the `conf
 
 #### Directory Permissions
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I would rather not go to this level of detail. That is more something for the Linux documentation, etc.

brunen9:
@taylorotwell, in my case after installing Laravel and create first project I got just a blank page instead of welcome screen. It takes a while to get the solution - as usual, Laracasts&StackOverflow helped. But don't you think that including small advice about permissions in documentation is more appropriate than force people to find answers in web? After all, documentation's target is to give the most complete information how to use a tool, isn't it?

PR #2654, opened 2016-09-12
array_where callback argument order

bezko:
($key, $value)  not ($value,$key) [unless I am missing something]


taylorotwell:
The original is correct. Make sure you are viewing the correct version of the Laravel docs for your Laravel version.

Issue #2593, opened 2016-08-29
Document on Custom Validation Attributes


dionysiosarvanitis:
Is this somewhere documented except for validation.php lang file?

````php
    /*
    |--------------------------------------------------------------------------
    | Custom Validation Attributes
    |--------------------------------------------------------------------------
    |
    | The following language lines are used to swap attribute place-holders
    | with something more reader friendly such as E-Mail Address instead
    | of "email". This simply helps us make messages a little cleaner.
    |
    */
```

paulodealmeida:
Look this https://laravel.com/docs/5.2/validation#localization

dionysiosarvanitis:
@paulodealmeida the difference is that this one describes the way of translating the message of a rule (based on a specific attribute) but not the attribute itself.

paulodealmeida:
@dionysiosarvanitis I can't understand your doubt.
I use this custom messages as the docs describes and it works. You just need to create a file named validation in resources/lang/en/validation.php.
How can we improve the documentation for other users?

dionysiosarvanitis:
@paulodealmeida I am talking about the difference between [custom](https://github.com/laravel/laravel/blob/master/resources/lang/en/validation.php#L98) and [attributes](https://github.com/laravel/laravel/blob/master/resources/lang/en/validation.php#L115)?

paulodealmeida:
good question...

bobbybouwmann:
Custom is used for custom messages per rule. So if you want to change the the message for a the required rule for only the email field you do something like this
```
'custom' => [
        'email' => [
                'required' => ':attribute is required you silly user!',
        ],
],
```

Attributes are used to translate the name of the field to a human readable field. So lets say your column is called `end_date` then your validation message will look like this "The end_date field is required". You can then use the attributes array to change that
```
'attributes' => [
        'end_date' => 'End Date',
],
```
This will result in this "The End Date field is required"

bobbybouwmann:
Created a PR for this ;)

dionysiosarvanitis:
Issue can be closed since https://github.com/laravel/docs/pull/2658 merged.

PR #2658, opened 2016-09-13
Add section for validation attributes

bobbybouwmann:
See #2593


dionysiosarvanitis:
Maybe it's better to stay consistent with the comment already exists in the config file. So E-Mail Address instead of "email" is more appropriate example. 

Also, attribute is not necessarily a column name. Maybe it should be change to field name.

PR #2661, opened 2016-09-14
Artisan migrate required before auth is fully functional

joevallender:
Avoid confusing new users who are just getting started with the 'missing users table' error


taylorotwell:
I think there is sort of no end to this kind of documentation though. Before running migrate you need to configure your .env file. Before doing that you need a hosting environment, etc.

PR #2665, opened 2016-09-15
Taggables

gizburdt:
In the Table Structure it is called `taggables`, but in the Model Structure the table name passed to `morphToMany` is `taggable`. In this PR this is changed to `taggables` so it's consistent.


brayniverse:
I believe the second argument is the name of the relation i.e. `taggable_id` and `taggable_type`. Changing this will make it `taggables_id` and `taggables_type`. Eloquent gets the table name from `App\Tag`.

taylorotwell:
See above comment.

PR #2667, opened 2016-09-17
Typo fix on middleware.md

sunscreem:
An 's' was missing from one of the first sentences.


taylorotwell:
I think the original is correct because middleware is plural in this case.

PR #2671, opened 2016-09-18
Authorizing a full resource controller (close #2444)

### LINKED ISSUE #2444: Add example of authorizeResource in authorization.md
While browsing through a preview of Matt's new book I noticed he mentioned using `authorizeResource()` in the constructor of a controller as a replacement for  having to do `$this->authorize(...)` for each function in the controller.

Looks like a pretty awesome time saver and would be great to see it in the docs.

Omranic:


taylorotwell:
Honestly holding off on this for now because it's not very explicit as to what middleware is actually added. Prefer to just document the explicit way of authorizing things.

ralphschindler:
For what it's worth @taylorotwell, we use this method when we have strong 1-1 relationships between models->resources->policies.  I think this particular PR could be cleaned up to discuss the middleware added by authorizeResource() and add the map of actions it will protect (view, create, update, etc..); and it will be very useful.

PR #2673, opened 2016-09-18
Use Class Notation in Authentication Events

LKDevelopment:
Use Class Notation in EventServiceProvider for Auth Events instead of 'Namespace\Classname'-Notation


taylorotwell:
See comment on previous PR.

PR #2674, opened 2016-09-18
Use Class Notation in EventServiceProvider Events instead of 'Namespace\Classname'-Notation

LKDevelopment:
Use Class Notation in EventServiceProvider Events instead of 'Namespace\Classname'-Notation where it is possible


taylorotwell:
This is not correct because the classes would have to be imported. That's why I don't list them this way.

PR #2677, opened 2016-09-19
Remove erroneous php prefix

Kindari:
Using the command as stated outputs the contents of the file because it is not a PHP file.
Composer installs a "proxy" file that handles cygwin detection and forwards the request to the real script.
( https://getcomposer.org/doc/articles/vendor-binaries.md#what-about-windows-and-bat-files- )
The real script uses the environment shebang invocation method to resolve php.

Example:

image

Note: tab spacing difference brings it inline with the windows equivalent of the same command in this section.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -138,7 +138,7 @@ Once Homestead has been installed, use the `make` command to generate the `Vagra
 
 Mac / Linux:
 
-    php vendor/bin/homestead make
+	vendor/bin/homestead make
 
 Windows:
 


taylorotwell:
That's probably a Windows only problem?

taylorotwell:
If so, I think we should specify that.

Kindari:
These commands were actually ran from inside of homestead itself and not through cygwin, but I will confirm on a linux server.

Kindari:
Looked into this a bit further and performed some testing on some of my other environments. The proxy file is generated under a few different scenarios but the main one being when a symlink cannot be created  to the target script. For me this has happened in two environments:

1) Inside of a homestead/vagrant vm on a filesystem shared from windows. Symlinks are not able to be created so the proxy method is used.
2) Linux-only shared filesystem (gluster) with symlink support disabled (security reasons)

The second one I think is unlikely for most people to encounter which means I agree this is mainly a windows issue.

Would you be interested in a note being added in a similar style as the introduction?

image

Thanks!

PR #2686, opened 2016-09-20
Added naming application in configuration file

paulodealmeida:
Added naming application in configuration file


taylorotwell:
I don't really think it's that great of an idea to rename your application to be honest.

paulodealmeida:
Ok, I just added this because it alredy exists on artisan. Thanks for reply.

PR #2692, opened 2016-09-21
Correct Class naming in morphMap

akoury:
I tried creating a morphMap in a project with this syntax, only to notice that i had to add a \ before the class


taylorotwell:
That just depends on where you are registering it.

PR #2693, opened 2016-09-21
Added missing helper

santigarcor:
Added the with helper, it was missing in the 5.3 docs.


taylorotwell:
It's not necessary in recent PHP versions.

PR #2694, opened 2016-09-21
PhpRedis Support

tillkruss:
What's still missing from the documentation is the difference in connection parameters.

Predis and PhpRedis both support:

- `host`
- `password`
- `port`
- `database`
- `timeout`
- `persistent`

Predis supports these:

- `scheme`
- `path`
- `async`
- `read_write_timeout`
- `alias`
- `weight`
- `iterable_multibulk`
- `throw_errors`

PhpRedis supports these:

- `read_timeout`
- `prefix`

Especially the `read_write_timeout` and `read_timeout` is noticeable. I tried to prevent that, but https://github.com/laravel/framework/commit/bae60487cd2749eb57d2e2eeb7bf8c3ce71ef3ec#diff-dbfdb06f6a76b9f93f18f77fa7f4b298 reverted my attempt.

I'm not quite sure how to mention that without adding new sections. Any suggestions?

### DIFF
--- a/redis.md
+++ b/redis.md
@@ -2,6 +2,8 @@
 
 - [Introduction](#introduction)
     - [Configuration](#configuration)
+    - [Predis Configuration](#predis-configuration)
+    - [PhpRedis Configuration](#phpredis-configuration)
 - [Interacting With Redis](#interacting-with-redis)
     - [Pipelining Commands](#pipelining-commands)
 - [Pub / Sub](#pubsub)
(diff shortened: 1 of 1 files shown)


taylorotwell:
I would suggest adding a new sub-section after "Configuration" called "PhpRedis" or something where we can detail how to use the PhpRedis client.

tillkruss:
Done. Let me know what you think.

boynet:
Hi is it possible to add info about this comment?

> If you have the PhpRedis PHP extension installed via PECL, you will need to rename the  Redis alias in your config/app.php configuration file.

there is no info about what its need to be changed to

PR #2697, opened 2016-09-22
Fixed typo

jayywolff:
Hashing time 'decreases' as hardware power increases.


taylorotwell:
The point of the work factor is that you can increase the amount of time it takes to create the hash.

PR #2701, opened 2016-09-23
Updated contains method documentation

Codeklopper:
Removed the $key parameter. Otherwise it would evaluate the $key parameter as the value.


taylorotwell:
This isn't correct. It's $value, $key... make sure you are using Laravel 5.3 if you are PRing the 5.3 docs.

PR #2707, opened 2016-09-26
Undocumented alias methods

nedmas:
I just used `doesntHave` in a project after finding it through a Google search. So I thought I'd add a tip to the documentation as it was the first place I looked and it isn't in here.


taylorotwell:
This would need proper documentation not just a tip I think.

PR #2709, opened 2016-09-27
Document new routes folder in upgrade.md

benmag:
Adds some documentation around the new routes location in 5.3

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -461,9 +461,15 @@ If you are queueing jobs using this syntax, Eloquent models will no longer be au
 
 ### Routing
 
-Laravel 5.3 introduces a `routes` folder at the root of the project. Your old `app/Http/routes.php` file is now split into two files `routes/web.php` and `routes/api.php`. These files are automatically loaded by the framework and means `app/` is now fully PSR-4 compliant. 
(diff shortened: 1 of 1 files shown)


lagbox:
Im not sure if this needs to be in the upgrade docs at all. The framework didn't change to force this. This is at the application level, which you already fully control, so its very optional.

Im a little worried the wording on this PR might be confusing and make someone expect to see a 'routes' folder, which wont exist, and those route files, which wont exist; as the person might not realize that composer doesn't touch the actual application. They also wont have the new RouteServiceProvider (default one from laravel/laravel 5.3) so making this change to the routes would break their application.

benmag:
@lagbox I see your point. I'll reword it so that it's more instructional and has information about the new `RouteServiceProvider`

taylorotwell:
This isn't necessary to upgrade to 5.3.

lagbox:
I agree it isn't necessary, but should the Authentication section in the Upgrade Guide  be mentioning `routes/web.php` without any explanation of the fact you won't have that file when upgrading? Maybe the reference to `routes/web.php` should just be removed from that section.

PR #2710, opened 2016-09-27
Remove return from global scope

shadoWalker89:


taylorotwell:
The current docs appear correct to me. If they are not open a new PR with a description.

Issue #2715, opened 2016-09-29
Envoy: Should the hostname in the very first example have array brackets?


jamesggordon:
The very first task example in the Envoy docs defines a host as follows:

```
@servers(['web' => ['user@192.168.1.1']])
```

Whilst this works when running tasks, if you issue `envoy ssh web` then you get this error:

_Array to string conversion in /Users/name/.composer/vendor/laravel/envoy/src/Console/SshCommand.php on line 38_

Removing the brackets from the hostname so that the line reads like this, works:

```
@servers(['web' => 'user@192.168.1.1'])
```

I noticed that none of the other examples include the brackets, so I can only assume that they are not supposed to be there in this instance either.

bobbybouwmann:
Created a PR for you ;)

jamesggordon:
Thanks @bobbybouwmann!

laurencei:
Closing as this issue was discussed in the PR.

PR #2717, opened 2016-09-29
Remove brackets from example

bobbybouwmann:
See #2715 

> Note: This issue doesn't exist in the 5.2 docs.. So not sure what happend here

### DIFF
--- a/envoy.md
+++ b/envoy.md
@@ -39,7 +39,7 @@ You may also use Composer to keep your Envoy installation up to date. Issuing th
 
 All of your Envoy tasks should be defined in an `Envoy.blade.php` file in the root of your project. Here's an example to get you started:
 
-    @servers(['web' => ['user@192.168.1.1']])
+    @servers(['web' => 'user@192.168.1.1'])
 
     @task('foo', ['on' => 'web'])
         ls -la


taylorotwell:
Brackets are fine.

taylorotwell:
We should fix the root issue not modify the docs to satisfy broken behavior.

jamesggordon:
@bobbybouwmann I had a deeper look at this issue and discovered that you can define a server alias to point to more than one server.

So this, for example, works:

    @servers(['web' => ['user@192.168.1.1', 'user@192.168.1.2']])

If you run **envoy run foo on web** it will run the task on both servers referred to by the "web" alias. This functionality is not mentioned in the Multiple Servers section of the docs.

If you were to run **envoy ssh web** (which, admittedly, doesn't make much sense) then you will get *Array to string conversion in envoy/src/Console/SshCommand.php on line 38*. Unfortunately, you also get the error even if you have only one server in the array. So I think the fix that Taylor really wants is a couple of checks when running the ssh command. (1) If the server is defined as an array but there is only one item, then pass that item into the ssh command and (2) if there are multiple servers defined in the array then throw an error. Seems like a lot of work for relatively little gain.

Personally, I think it makes more sense to disallow the syntax above, as it provides, IMO, very little extra benefit over and above the facility that is described in the Multiple Servers section, which also works nicely with the ssh command. However, if Taylor prefers it this way then who am I to argue. :smile:

The real test will be how many people lodge issues related to the ssh command not working when they define a server name within brackets and how many of those it takes before Taylor finally cracks and disallows the syntax above. :wink:

bobbybouwmann:
@jamesggordon Thanks for explaining, learned something new today ;)

PR #2719, opened 2016-09-30
Point to 5.2 version of config/auth.php (rather than master)

CrookedNumber:
Otherwise instructions like "take special note of the passwords.users.email configuration option in the new auth.php" are confusing, as passwords.users.email no longer exists in the master branch version of the file.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -524,7 +524,7 @@ Add `"symfony/dom-crawler": "~3.0"` and `"symfony/css-selector": "~3.0"` to the 
 
 #### Configuration File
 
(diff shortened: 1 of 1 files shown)


upgrade.md
```diff
@@ -524,7 +524,7 @@ Add `"symfony/dom-crawler": "~3.0"` and `"symfony/css-selector": "~3.0"` to the
 
 #### Configuration File
 
-You should update your `config/auth.php` configuration file with the following: [https://github.com/laravel/laravel/blob/master/config/auth.php](https://github.com/laravel/laravel/blob/master/config/auth.php)
+You should update your `config/auth.php` configuration file with the following: [https://github.com/laravel/laravel/blob/5.2/config/auth.php](https://github.com/laravel/laravel/blob/5.2/config/auth.php)
```
anteriovieira: `{{version}}` It is automatically replaced by the current version. Maybe it is more appropriate.

PR #2721, opened 2016-10-01
Homestead - Fix "Per Project Installation" command

Djuuu:
`vendor/bin/homestead` is a shell script, not a PHP script.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -138,11 +138,11 @@ Once Homestead has been installed, use the `make` command to generate the `Vagra
 
 Mac / Linux:
 
-    php vendor/bin/homestead make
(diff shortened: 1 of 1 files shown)


taylorotwell:
Isn't this only the case on Windows?

Djuuu:
Yeah, I thought the two alternatives were just about _directory separator_ styles. (Using Cygwin, I use Unix-style anyway). 
I didn't realize that the shell script was [a proxy generated by Composer](https://getcomposer.org/doc/articles/vendor-binaries.md#what-about-windows-and-bat-files-).

My bad, sorry for the noise.

PR #2722, opened 2016-10-02
passport installation additional info

Dlahacz:
Only change that's not mentioned in text.


taylorotwell:
We already show the import in the example.

PR #2732, opened 2016-10-06
Add vagrant-hostsupdater instructions

seanmangar:
Add instructions on how to install vagrant-hostsupdater to automate
entry of relevant files to the hosts file.


taylorotwell:
I don't really want to recommend this because A) I myself have had problems with it.. and B) I don't want to provide technical support for it when people can't get it to work.

seanmangar:
Yeah that makes a lot of sense. Thanks for reviewing it :)

PR #2734, opened 2016-10-06
Update scout.md

fridzema:
This is the way to go i think?


taylorotwell:
The original docs are correct.

PR #2744, opened 2016-10-11
Remove required argument

whoan:


taylorotwell:
I think this was intentional to show using both kinds of arguments at the same time.

whoan:
Maybe, but `{user}` is not being used in the example:

    php artisan email:send --id=1 --id=2

I didn't try it but I am sure that won't work. Do you want me to change the example?

PR #2754, opened 2016-10-14
Update eloquent.md :: Function updateOrCreate

reno1979:
The updateOrCreate function can not be called / used statically when using multiple database connections. 

It caused my application to check in connection 1 and insert in connection 2....

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -377,6 +377,18 @@ The `firstOrNew` method, like `firstOrCreate` will attempt to locate a record in
     // Retrieve the flight by the attributes, or instantiate a new instance...
     $flight = App\Flight::firstOrNew(['name' => 'Flight 10']);
 
+> Note: When using multiple connections, using the on() method on a static will make the updateOrCreate() fail.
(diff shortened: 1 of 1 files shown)


taylorotwell:
Why don't we just fix the bug instead of document the bug?

reno1979:
Ehm,

I kinda hoped the laravel team would. And thus a fix would be great,
reporting An issue did not seem possible on github.

I discovered that the bug also happens when using An other than default
connection on An instantiated model.

Op 14 okt. 2016 10:22 p.m. schreef "Taylor Otwell" <notifications@github.com
>:

Why don't we just fix the bug instead of document the bug?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/laravel/docs/pull/2754#issuecomment-253909872>, or mute
the thread

taylorotwell:
Issues can be reported at laravel/framework repository. I have also added this to our internal Trello to take a look at. Thanks.

PR #2757, opened 2016-10-17
Add Carbon\Carbon trait to AuthServiceProvider in configuring passport

DrowningElysium:
Based on someone asking for help for forgetting to add use Carbon\Carbon;


MicroDroid:
That was me

PR #2759, opened 2016-10-19
Remove the unnecessary "you" from the sentence.

tiggreen:
It should be "Don't worry if this directory does not exist ...".


taylorotwell:
Thanks

PR #2760, opened 2016-10-19
Let people know they can pass synced folder options in the homestead.yml

etherealite:
It took me far longer than I care to admit to learn that you could pass-through arbitrary configuration options to vagrant synced folders from within the homestead.yml. Maybe my rsync case makes a bad example, but I think the docs should mention the fact that you can do this.


tpraxl:
This would have saved quite some time on my end as well.

PR #2774, opened 2016-10-25
Document the existence of updateOrCreate and provide an example

jtgrimes:
I'd like to have this in the docs just because I can never remember whether it's updateOrCreate or updateOrNew.


taylorotwell:
Thanks! However, the act of creating this PR has probably cemented the syntax into your brain so that you will never need this documentation again! 😛

PR #2779, opened 2016-10-26
[5.3] Authorization - Customizing `email` to `username`

brianseitel:
This PR adds some additional clarity on how to change the default user key for authorization from `email` to something else, such as `username`. This is a common question that comes up from other devs, and since this behavior changed between 5.2 and 5.3, it seems like something that the docs could benefit from having.

Happy to make copy-edits, if requested.

### DIFF
--- a/authentication.md
+++ b/authentication.md
@@ -79,7 +79,7 @@ When a user is not successfully authenticated, they will be automatically redire
 
 #### User Field Customization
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Adding a username method to your RegisterController would not do anything.

brianseitel:
Whoops! Brain fart. Will fix when I get back to a computer.

brianseitel:
Updated. Thanks, @taylorotwell.

PR #2788, opened 2016-10-29
Document unchangeable column types.

Kender2:
Due to the doctrine dbal being used for the change method, not all column types can be changed.

As suggested by @GrahamCampbell in issue [#9636](https://github.com/laravel/framework/issues/9636) I'm submitting a PR for the docs.

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -283,7 +283,7 @@ Before modifying a column, be sure to add the `doctrine/dbal` dependency to your
 
 #### Updating Column Attributes
 
-The `change` method allows you to modify an existing column to a new type or modify the column's attributes. For example, you may wish to increase the size of a string column. To see the `change` method in action, let's increase the size of the `name` column from 25 to 50:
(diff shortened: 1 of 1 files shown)


taylorotwell:
I mean maybe it would be easier to document the ones that **can** be changed? Since this is a pretty big list?

Kender2:
The list of columns that can be changed is only slightly shorter:
bigIncrements, bigInteger, binary, boolean, date, dateTime, dateTimeTz, decimal, float, increments, integer, longText, mediumText, smallIncrements, smallInteger, string, text, time, unsignedBigInteger, unsignedInteger, unsignedSmallInteger

I had hoped that by listing the ones that cannot be changed people might:
1. try to reduce the length of the list by making them changeable after all.
2. try to avoid making the list longer by not adding any more unchangeable column types.

mehrdaad:
@Kender2 do you know if dbal is going to support these types in future?

Kender2:
@mehrdaad No idea. I'm not involved in [doctrine/dbal](https://github.com/doctrine/dbal).

mehrdaad:
@Kender2 I think you can alter those columns with something like ``DB::statement('ALTER TABLE tablename ...')``  
even without dbal. right?

PR #2795, opened 2016-10-31
stopping the vagrant machine

BartHuis:
copied som einfo from vagrant, when first using vagrant i always used suspend, but when i did vagrant up after that the next day, it gave problems, i want to spare people thair time from making the same mistake ;)


taylorotwell:
I think this is better kept in the Vagrant docs for now.

PR #2796, opened 2016-10-31
dump autoload before migration in case of new models

BartHuis:
dump autoload before migration in case of new models

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -127,6 +127,10 @@ You may rollback & re-migrate a limited number of migrations by providing the `s
 
     php artisan migrate:refresh --step=5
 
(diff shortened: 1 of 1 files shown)


migrations.md
```diff
@@ -127,6 +127,10 @@ You may rollback & re-migrate a limited number of migrations by providing the `s
 
     php artisan migrate:refresh --step=5
 
+When you've made new models and you get an error on non excisting classes, run the command:
```
browner12: *existing*

taylorotwell:
Going to hold off on this for now.

migrations.md
BartHuis: existing?

PR #2797, opened 2016-10-31
Include Form Model Binding With Generic Controller Methods

frigginglorious:
The action method was pointing to a route for a different model, because they used the same Controller and Method. This was surprisingly frustrating until I realized I could specify the 'url' attribute directly. It wasn't mentioned anywhere, and I thought it might be useful to someone else!


taylorotwell:
Laravel doesn't ship with Form helpers.

PR #2806, opened 2016-11-04
Add auth()->id() function

TheShiftExchange:
Just documenting a missing feature


taylorotwell:
It might be better to put this in the authentication docs. I'll add it now.

PR #2808, opened 2016-11-04
Use single back slash for windows path

KluVerKamp:


taylorotwell:
It used to be like that and a Windows user told me to change it to this? Need more confirmation on this.

KluVerKamp:
hmm he might wanted to make sure that they are correctly rendered as a single back slash out of the md file? I would only use them if I am writing a path in a string.

Anyway!

PR #2813, opened 2016-11-08
Add note about @stop directive

mdavis1982:
Add a note about using `@stop` interchangeably with `@endsection`.

### DIFF
--- a/blade.md
+++ b/blade.md
@@ -74,7 +74,7 @@ When defining a child view, use the Blade `@extends` directive to specify which 
 
 In this example, the `sidebar` section is utilizing the `@@parent` directive to append (rather than overwriting) content to the layout's sidebar. The `@@parent` directive will be replaced by the content of the layout when the view is rendered.
 
-In addition to using the `@@endsection` directive to end a section, you can also use `@@stop` which has the same effect.
+Instead of using the `@@endsection` directive to end a section, you can also use `@@stop` which has the same effect.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Ah OK, for this one I think I don't show it because when I wrote 5.3 docs I decided to just typically show 1 way to do things instead of showing several different ways to do the same thing. It just makes the docs more streamlined and simple. I think the other things you mentioned on Twitter are more needed in the documentation though.

mdavis1982:
@taylorotwell sure, but it's confusing when you see people using it (@jeffreyway uses it a lot on his videos), and it's not in the documentation

taylorotwell:
Ah, wasn’t aware that it was showing up in videos a lot. OK will reconsider that

JeffreyWay:
Haha - Yeah I've used @stop for years. Just habit. I can switch over to @endsection.

mdavis1982:
Cool... Either works 😄 

I'm digging into Blade a lot in preparation for an upcoming project which is how I noticed that these directives weren't in the documentation. Lots of source diving done today!

tkaw220:
@mdavis1982 Mind to share which source code files you dig for hidden Blade gem?

mdavis1982:
@tkaw220 Sure... I've been spending a lot of time in here: https://github.com/laravel/framework/blob/5.3/src/Illuminate/View/Compilers/BladeCompiler.php

PR #2814, opened 2016-11-08
Add information about Blade @php directive

mdavis1982:


blade.md
```diff
@@ -99,6 +99,16 @@ Of course, you are not limited to displaying the contents of the variables passe
 
 > {note} Blade `{{ }}` statements are automatically sent through PHP's `htmlentities` function to prevent XSS attacks.
 
+### Embedding PHP Code Blocks
+In some situations, it's useful to embed PHP code into your views. In addition to allowing you to do that within a Blade echo statement, you can also use the Blade `@@php` directive:
```
lukepolo: why is there 2 `@@`

mdavis1982: I've removed it. Incidentally, there are 2 `@@` for the `@parent` - I presume that must be a reserved word in the parser or something.

lukepolo: gotcha, well it may need to stay in there then, guess we will find out!

PR #2815, opened 2016-11-08
Document new prepareForValidation method

jtgrimes:
The wording isn't perfect, but it's something.

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -9,6 +9,7 @@
 - [Form Request Validation](#form-request-validation)
     - [Creating Form Requests](#creating-form-requests)
     - [Authorizing Form Requests](#authorizing-form-requests)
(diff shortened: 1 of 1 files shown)


validation.md
```diff
@@ -261,6 +262,22 @@ If you plan to have authorization logic in another part of your application, sim
         return true;
     }
 
+<a name="modifying-data-before-validation"></a>
```
rscafi: I guess that it should be `<a name="executing-data-before-validation"></a>` so that the anchor works

mdavis1982: Shouldn't it be `executing-code-before-validation` as in the link above?

jtgrimes: Thanks. Will update this afternoon.

taylorotwell:
Can you send this to master branch?

PR #2821, opened 2016-11-11
Link to Resource Controllers

cweiske:


taylorotwell:
I'm not sure that we need that link right here.

cweiske:
If that link had been in there before, then I'd have saved about 2 hours today.

PR #2824, opened 2016-11-11
Added note on registering commands

bmitch:


bmitch:
A new user might try generating their command then immediately running `php artisan` and expect to see the new command listed. The "Registering Commands" section isn't until the very end which in my opinion is going to be the next step when creating a new command.

So thought a note right below would be helpful. Either that or move the "Registering Commands" section up.

PR #2827, opened 2016-11-14
Add information about deleteFileAfterSend()

rsands2801:


taylorotwell:
This feels edge case enough that I would prefer to just leave it out.

rsands2801:
@taylorotwell That's unfortunate as it is pretty much required feature if you use a load balanced server setup.  Tidy up after yourself and leave the server nodes clean.  That and the API documentation has it buried (not even google can find it)

PR #2837, opened 2016-11-16
Update authorization.md

mnrahimi:
I think it could be `note` not `not`.


taylorotwell:
The original was correct.

taylorotwell:
I will tweak the wording a bit though to make it not so awkward.

mnrahimi:
Yes, that is a great idea, because it is a little bit confusing now.
**Temp note:**  I sent an email to your `taylor@laravel.com` email address, from `mnr@karyal.com`.
The above note will be removed after your signal.

PR #2839, opened 2016-11-17
Added note about versioned files in css being replaced in production

imerr:
I had trouble finding any documentation about that behaviour and I probably won't be the only one with that issue

### DIFF
--- a/elixir.md
+++ b/elixir.md
@@ -321,6 +321,8 @@ After generating the versioned file, you may use Laravel's global `elixir` helpe
 
     <link rel="stylesheet" href="{{ elixir('css/all.css') }}">
 
+Versioned files are also automatically adjusted to the versioned link in stylesheets when running `gulp` with the `--production` flag.
+
 #### Versioning Multiple Files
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't know what this means.

imerr:
When you want to version images for example with elixir for cache busting and you reference those images in stylesheets (background images for example) it replaces the unversioned image path with the versioned one when you run gulp in production mode

That behaviour is documented nowhere as far as I could see

michaeldyrynda:
You should just use the `elixir` helper function in your views, which is documented in the paragraph above where you [added this change](https://laravel.com/docs/5.3/elixir#versioning-and-cache-busting).

> Once the files have been versioned, you may use the elixir helper function to generate links to the proper hashed files.

imerr:
I am not talking about views
I am talking about css files (/stylesheets)

michaeldyrynda:
Ah, right, I misread. If they're automatically adjusted in your stylesheets, this isn't a problem, you're just documenting the existing behaviour?

imerr:
exactly

PR #2856, opened 2016-11-23
Add note about delayed job behavior when using QUEUE_DRIVER=sync

timbroder:
This is obvious once you think about it, but not really called out anywhere. Should be helpful if someone is having an issue testing delayed jobs


taylorotwell:
The fact that it is synchronous is listed at the top of the page. Perhaps the note should be clarified there instead.

timbroder:
@taylorotwell can do. Should I make a new PR or modify this one?

taylorotwell:
New PR, Thanks

PR #2860, opened 2016-11-25
Note renumbering key's

fernandobandeira:
Refs laravel/framework#16552

People can use a workaround with ``` keyBy() ``` or ``` pluck() ```
Example https://github.com/laravel/framework/issues/15409#issuecomment-247083776

Not sure if we could mention that, usually they can just use another function anyway...

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -724,6 +724,8 @@ The `mapWithKeys` method iterates through the collection and passes each value t
         ]
     */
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
This explanation isn't very clear. It may be clearer to show an example so I can see what you mean. Re-open a PR with that if you have time.

PR #2863, opened 2016-11-27
[5.4] Automatic filename generation now uses a different method

vlakoff:
Follow-up to https://github.com/laravel/framework/pull/16193.

Previously MD5 of file content, now UUID 4 without dashes, but let's just say "random unique identifier". The user doesn't care about the details, and more importantly this can change later.


taylorotwell:
Better to just say UUID because developers will want to know the details. Can you re-submit with that?

PR #2864, opened 2016-11-27
[5.4] Automatic filename generation now uses a different method

vlakoff:
Resubmit of #2863 with requested change.

### DIFF
--- a/filesystem.md
+++ b/filesystem.md
@@ -159,13 +159,13 @@ If you would like Laravel to automatically manage streaming a given file to your
 
     use Illuminate\Http\File;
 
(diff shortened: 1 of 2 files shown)


browner12:
I believe we should be using 'a UUID' since it is pronounced like 'you' which starts with a consonant sound.  We would say 'a user', not 'an user'.

http://www.englishpage.com/articles/a-vs-an.htm

vlakoff:
@browner12 Correct. I'm finding numerous sources saying it has to be "**a** unique". Do you want to submit a PR?

PR #2871, opened 2016-11-30
Updated Authorization Readme File

atefth:
Added documentation to indicate the convention of using `@can`, `@cannot` and `@authorize` directives with custom error messages to override the default unauthorized access message.


taylorotwell:
That is not possible atm.

PR #2874, opened 2016-11-30
Fix typo

zlatevbg:


taylorotwell:
I removed that line entirely. It doesn't make sense right there. I think it was left over from old docs.

PR #2887, opened 2016-12-03
Add weeklyOn() method

andrey-helldar:


taylorotwell:
We should clarify which day of the week is 0. Also would be 2nd not 2th.

taylorotwell:
Feel free to re-submit when these changes have been made.

PR #2891, opened 2016-12-06
Add tip for policy filters

Kyslik:
clarify what happens when policy filter (before()) returns true, false,
null


Kyslik:
Maybe it needs better wording, but sure its important to have that information in documentation, I've burned my-self returning false... (twice)

PR #2893, opened 2016-12-06
Update database-testing.md

capir:
Error: Call to undefined method Illuminate\Database\Eloquent\Factory::state()

it must be the defineAs method in this case.

Thanks


taylorotwell:
Sounds like your Laravel is out of data. State is a valid function.

PR #2901, opened 2016-12-10
Require any version of valet

jeroennoten:
This makes sure that composer global update actually updates to any new (major) version. It's maybe too late for the upgrade to version 2 now, because everyone has a version constraint to Valet 1 now, but at least new people can just `composer global update` when Valet 3 or higher comes out.


taylorotwell:
Hmm... I don't necessarily want to recommend this because people will have no way to stick to a minor version if they aren't ready for some yet future unknown breaking change.

jeroennoten:
You're right. Was thinking that after posting this. We better change the upgrade instructions to `composer global require laraval/valet:2.*`.

PR #2904, opened 2016-12-12
Add HeidiSQL as alternative for GUI Database tool

xhs345:
Navicat is not free / only has a short trial period. Sequel Pro is Mac OS X only. HeidiSQL is OpenSource, runs on Windows as well and is easy to use


taylorotwell:
Removed the reference to specific programs entirely since we will just keep adding to the list.

xhs345:
Yes, fair enough :)

PR #2905, opened 2016-12-12
Update helpers.md

lg430:
swap $key, $value


taylorotwell:
This is wrong. You are probably not using 5.3.

PR #2908, opened 2016-12-13
Added note on .htaccess when upgrading from 5.3

DODMax:
There is no upgrade note suggesting to add the following to .htaccess when migrating to 5.3
RewriteCond %{HTTP:Authorization} .
RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]

It really took me some time to figure that out :-)


taylorotwell:
This was in the default .htaccess file in Laravel 5.2 as well.

PR #2910, opened 2016-12-13
mention the Ubuntu valet fork

dakira:
While I understand that the fork can't be merged back into the official valet, it has proven to be actively maintained and is always in sync with official Valet development.

I think it should at least be mentioned in the docs as an unsupported community fork.


taylorotwell:
I can't really mention that in the official docs as I have nothing to do with ensuring its quality, etc.

PR #2911, opened 2016-12-14
Added missing RedirectIfAuthenticated change when customizing path after authentication in Authentication Docs

anandsinghkunwar:


taylorotwell:
This doesn't make sense to be placed here.

anandsinghkunwar:
I felt we needed to change any redirects to 'home', am I not right in my thinking?

PR #2915, opened 2016-12-14
Add back missing @hasSection documentation

mtdavidson:
Previously removed in commit https://github.com/laravel/docs/commit/fec6ca03fab981d9a751b476e71cf00c0f15c0af#diff-5ec203a153377bbcca8d0af017818bcf but still valid and useful syntax.

### DIFF
--- a/blade.md
+++ b/blade.md
@@ -164,6 +164,16 @@ For convenience, Blade also provides an `@unless` directive:
     @unless (Auth::check())
         You are not signed in.
     @endunless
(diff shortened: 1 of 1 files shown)


taylorotwell:
You can just do `@yield('title', 'App Name')` in this case.

taylorotwell:
So we would need a better example probably.

taylorotwell:
Closing pending a better example.

PR #2921, opened 2016-12-15
Including @overwrite in Blade Docs, v2

charlielondon:
https://github.com/laravel/framework/issues/1058

### DIFF
--- a/blade.md
+++ b/blade.md
@@ -81,6 +81,8 @@ Blade views may be returned from routes using the global `view` helper:
         return view('child');
     });
 
+If a compiled template contains multiple section definitions of the same name that each expect different content, as is the case when using partials that use inheritance, you may terminate your `@@section` injection directives with `@@overwrite`.
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can you provide the use case you are using @overwrite for? I'm curious if it would be better to just use new @component syntax.

charlielondon:
Simplified version:
master: `views/form.blade`
parent partial: `views/partials/form_element.blade`
derived partial: `views/partials/text_input.blade`
derived partial 2: `views/partials/date_input.blade`

in `form.blade`
```<form>
    @include('text_input', ['name' => 'title'])
    @include('date_input', ['name' => 'date'])
    ...
</form>```

in `text_input.blade`
...
```@section('element')
    ...
@overwrite```

taylorotwell:
Yeah this needs to be just using components in Laravel 5.4 Using overwrite is a much more confusing way of building out this type of behavior.

PR #2925, opened 2016-12-19
fix routing definition for `Explicit Binding`

wumouse:
no need to add a type hint `App\User` in route definition for `Explicit Binding`.


taylorotwell:
It still drives the point home I think.

PR #2927, opened 2016-12-20
Add nullableMorphs in migrations

a-komarev:
I don't know how to better describe that fields will be nullable:

```
Adds unsigned INTEGER `taggable_id` NULL and STRING `taggable_type` NULL
```

Or:

```
Adds nullable unsigned INTEGER `taggable_id` and nullable STRING `taggable_type`
```


tillkruss:
You could even shorten it to.

```
Same as `morphs()`.
```

Same as `nullableTimestamps()` does.

a-komarev:
It's not the same. Because it's generating nullable fields.

PR #2929, opened 2016-12-20
Eloquent save method return value

ejunker:
After running some tests and looking at the code in Illuminate\Database\Eloquent\Model::save() it looks like it returns `true` not `false` if the model has not been changed since the last time it was retrieved or saved.


taylorotwell:
I just removed this section. I think it was reverted at some point.

PR #2931, opened 2016-12-22
Add info on Dnsmasq

gbs:
With Dnsmasq adding *.app domains to the hosts file is no longer needed, which makes a developer's life a little easier.


taylorotwell:
I don't want to try to provide official guidance for this in the docs.

PR #2941, opened 2016-12-30
Accessing Route Model Bindings in Middleware

viraj-khatavkar:
Documentation doesn't contain any information regarding the above use-case. I had to google it quite a few times for many of my apps. I think this would deserve a place in the docs.

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -271,6 +271,21 @@ Since we have bound all `{user}` parameters to the `App\User` model, a `User` in
 
 If a matching model instance is not found in the database, a 404 HTTP response will be automatically generated.
 
+### Accessing Route Model Bindings in Middleware
+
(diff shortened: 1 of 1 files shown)


lagbox:
You can access route params from the Request as property `$request->param`. The `__get` method calls to `route` as the fallback. One should not do it this way but you 'can'.

viraj-khatavkar:
@lagbox In middleware you cannot - I have tried it :)

lagbox:
Im sorry that you can't but it works as I described. 
I can demonstrate this works as expected from a fresh 5.3 install.

GrahamCampbell:
> In middleware you cannot - I have tried it :)

Yes you can, provided your middleware is run **after** the binding middleware.

lagbox:
Even if it was before the binding middleware you can still get the original parameter from the request via the dynamic property (assuming there are no inputs with the same name).

taylorotwell:
It depends on where your middleware is located in the stack.

PR #2950, opened 2017-01-06
Fix authentication redirectTo

vfsoraki:
Docs mention that `redirectTo` method may be used instead of `redirectTo` property, which is incorrect. `authenticated` method may be used instead.

### DIFF
--- a/authentication.md
+++ b/authentication.md
@@ -75,14 +75,16 @@ When a user is successfully authenticated, they will be redirected to the `/home
 
     protected $redirectTo = '/';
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
The docs are correct

vfsoraki:
@taylorotwell Look here http://stackoverflow.com/questions/41506102/laravel-redirect-after-login

Was I wrong or something? Code says something else.

vfsoraki:
I get it, fixed in 5.3.29.

But seems rather strange that I got 5.3.28. I created the project some hours ago, it should be latest version.

Issue #2949, opened 2017-01-05
redirectTo property and method discussed in Docs


nakie:
https://laravel.com/docs/5.3/authentication#included-authenticating

discusses overriding the redirectTo Property with a function of the same name for dynamic generation of redirectTo Route/URL. I wasn't able to get this to work and started researching its use.    I have only found reference to redirectPath() and don't see a redirectTo() mentioned anywhere else.  

is this simply a typo in the docs or am I missing something?

vfsoraki:
I asked the same question.

http://stackoverflow.com/questions/41506102/laravel-redirect-after-login

Seems like docs were wrong, I created a pull request.

nakie:
I understand now.  I was missing that the redirectTo function was introduced in 5.3.29.   

After reviewing vfsoraki's previous PR https://github.com/laravel/docs/pull/2950 and the Change logs https://github.com/laravel/framework/blob/5.3/CHANGELOG-5.3.md

I have created another PR https://github.com/laravel/docs/pull/2956 that will hopefully eliminate any confusion for others.

laurencei:
Hi there,

FYI - I'm going to close this issue.

The docs have been updated significantly since this was posted, and that section has been re-written.

If you believe there are still improvements that can be made, please feel free to submit a PR with the changes, or open a new ticket with the details of the problem.

Thanks.

PR #2956, opened 2017-01-07
Expand redirectTo Function Docs

nakie:
Expanding the documentation around the redirectTo function used in Auth controllers based on recent confusion.

### DIFF
--- a/authentication.md
+++ b/authentication.md
@@ -77,10 +77,17 @@ When a user is successfully authenticated, they will be redirected to the `/home
 
 If the redirect path needs custom generation logic you may define a `redirectTo` method instead of a `redirectTo` property:
 
-    protected function redirectTo()
-    {
-        //
-    }
+        /**
+         * Generate post-authentication redirect location
+         * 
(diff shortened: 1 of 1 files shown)


nakie:
in response to https://github.com/laravel/docs/issues/2949

authentication.md
```diff
@@ -77,10 +77,17 @@ When a user is successfully authenticated, they will be redirected to the `/home
...
 If the redirect path needs custom generation logic you may define a `redirectTo` method instead of a `redirectTo` property:
 
-    protected function redirectTo()
-    {
-        //
-    }
+        /**
+         * Generate post-authentication redirect location
+         * 
+         * @since 5.3.29 Introduced redirectTo function support in 5.3.29
```
GrahamCampbell: this is a misuse of that annotation

nakie: can you elaborate?  looked right to me based on the docs page https://phpdoc.org/docs/latest/references/phpdoc/tags/since.html

PR #2962, opened 2017-01-10

Skintillion:
Took a while to figure out why my morphMap wasn't working.


taylorotwell:
I would almost rather using strings in this case instead of ::class. Like we do in a lot of the rest of the Eloquent docs.

PR #2965, opened 2017-01-11
Details on how to map a custom domain

danielbachhuber:
Fixes https://github.com/laravel/valet/issues/215


taylorotwell:
I'm not sure the Laravel docs should be responsible for documenting how to use /etc/hosts in this way. This is also OS specific - won't work for Windows.

danielbachhuber:
@taylorotwell But this is for Laravel Valet, which is Mac only?

PR #2967, opened 2017-01-13
Add a note about not using closures in config files

HipsterJazzbo:
Package developers all-too-frequently resort to putting closures in their config files. While this is an anti-pattern, it appears to work fine until a user tries to do `php artisan config:cache`, at which point the closure is not serialized correctly and the app breaks.

This PR is a (probably futile) attempt to make package developers aware of this issue before they cause it :)

### DIFF
--- a/packages.md
+++ b/packages.md
@@ -73,6 +73,8 @@ Now, when users of your package execute Laravel's `vendor:publish` command, your
 
     $value = config('courier.option');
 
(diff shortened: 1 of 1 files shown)


HipsterJazzbo:
Side note: I've actually done this a few times, and as someone who does actually read the docs top to bottom every new version, I would have learned something from this note :)

PR #2979, opened 2017-01-17
Terminating callbacks documentation

nejtr0n:


taylorotwell:
This isn't the right place for this.

cbj4074:
So, it's closed without further explanation? Would it be too troublesome to point the submitter to the appropriate location in the documentation, rather than scrap what seems to be a worthwhile addition?

With all due respect, this is a case study in how best to discourage would-be contributors.

poppabear8883:
I agree with @cbj4074 ...

PR #2985, opened 2017-01-19
Update routing.md

MrMavin:
I think that using the 'as' array key option is more convenient than chaining the 'name' method to the end of route definition. So, it should be described in the new documentation since it's still working.


taylorotwell:
I only want to document one way of doing this.

PR #2989, opened 2017-01-19
Add info that ChromeDriver or eq. should run

pablopaul:
@taylorotwell I am unsure about good wording and still feel that this note could help a lot.


taylorotwell:
I will tweak the docs to mention this now with some different wording.

PR #2998, opened 2017-01-22
Laravel 5.3 Homestead instructions 

elstamey:
Updated yaml and hosts config for additional sites section

In order to clarify the instructions, I added an example project to the `Homestead.yml` sites configuration example.  I illustrated this because in forums, people were updating the `folders` section instead of `sites`.  I also added the instructions for `/etc/hosts` because that detail was omitted from instructions.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -210,7 +210,20 @@ To connect to your MySQL or Postgres database from your host machine's database 
 <a name="adding-additional-sites"></a>
 ### Adding Additional Sites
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'll look at the 5.4 one. Only needs to be sent there.

elstamey:
@taylorotwell why wouldn't the update be copied into other versions of the documentation?  Documentation for 5.3 comes up first in google searches, and it should move forward into newer versions.

PR #3009, opened 2017-01-23
Use methods concept instead of functions

DojoGeekRA:
`Methods` by definition are associated with an object, so given that the relationships are defined in the `Eloquent Model`, it makes more sense to called them `methods` instead of `functions`.


taylorotwell:
Can you send this to the 5.4 branch? Thanks!

DojoGeekRA:
@taylorotwell sure 👍

PR #3011, opened 2017-01-23
[5.4] Use methods concept instead of functions

DojoGeekRA:
`Methods` by definition are associated with an object, so given that the relationships are defined in the `Eloquent Model`, it makes more sense to called them `methods` instead of `functions`.


DojoGeekRA:
Done @taylorotwell , now it targets the right branch.

PR #3014, opened 2017-01-23
[5.4] Adding a precision to whereHas

jpmurray:
Hello, first time proposing changes to the doc, so I hope I'm doing things right!

I only added one line, under the `whereHas` example to explicitly say that you can also use nested statement with dot notation like you can do with the `has` method.

I'm making this edit because at first, reading the documentation, I didn't make the assumption that it was also possible because it wasn't written for `whereHas`. I just hope this small change can be beneficial to the understanding of the usage of the method.

I didn't write code example, I think that would be too much, but if it is required, please tell me, I'll write a small one.


taylorotwell:
We have an example of using nested relationships just a few lines above so I think it's fine for now.

PR #3020, opened 2017-01-24
Update upgrade.md

apex39:
Pasting `Laravel\Tinker\TinkerServiceProvider` causes `artisan`'s segmantation fault.

`Laravel\Tinker\TinkerServiceProvider::class` is what should be in the doc


taylorotwell:
I've updated this on the proper branch.

Issue #3044, opened 2017-01-28
[5.4] Passport documentation outdated


Quezler:
[Documentation](https://laravel.com/docs/5.4/passport#consuming-your-api-with-javascript) still speaks about vue-resource, while [the code](https://github.com/laravel/passport/pull/230/files) uses axios.

Quezler:
7f35fdfc9e8901b170f76bdc48595436fcf71d08
c289b3679f537bbee7473652e7013f02a392b0a6
bf8fa82f6782cb12dbb983cab9c49d0dcde9f7c5

PR #3045, opened 2017-01-28
Make passport.md compatible with axois

Quezler:


Quezler:
Part of #3044

brunen9:
@Quezler, good job! I only want to point that instead of `window.Laravel.csrfToken` we can use just `Laravel.csrfToken`.

Quezler:
Well sometimes (generally inside a vue file) it often has trouble recognizing some variables not prefixed with `window.`

And sometimes apps like PhpStorm even go as far as to mark it completely invalid, while it works in the background.

So that were the primary reasons, but then again, its just a pull request, so anything can change 😃

taylorotwell:
Those aren't even the same headers?

taylorotwell:
Axios will automatically send the CSRF token. There isn't any configuration that is needed in that regard.

Quezler:
Oh? on a fresh Laravel 5.4 password's 'consume your own javascript' refuses to work without this code?

Quezler:
(axios does send A header, but that is xsrf, with a c)

taylorotwell:
What I'm saying is the docs talk about setting a CSRF header but then you are setting a totally different header in your example.

taylorotwell:
Change the wording.

Quezler:
What parts should i change?

taylorotwell:
I'll do it.

taylorotwell:
I fixed a bunch of other places where Vue resource was still being used.

Quezler:
Correct me if i see it incorrectly, but i only see the changes i initially submitted, i don't see any changes from you?

brunen9:
> Axios will automatically send the CSRF token. There isn't any configuration that is needed in that regard.

@taylorotwell, but it doesn't. Comments in `bootstrap.js` says about it too, and it is really confusing, cause Axios only sends `X-XSRF-TOKEN`. There are some questions at laracasts and stackoverflow about this problem.

Quezler:
Could you provide links?

brunen9:
@Quezler 
[Comments in bootstrap.js](https://github.com/laravel/laravel/blob/master/resources/assets/js/bootstrap.js#L22)
[StackOverflow](http://stackoverflow.com/questions/40808405/how-to-add-token-when-using-axios-to-send-an-ajax-request)
[Laracasts](https://laracasts.com/discuss/channels/vue/laravel-54-and-axios)

Quezler:
The last link is exactly the problem i addressed in the PR ❤️

monomichael:
@taylorotwell I think you omitted the fact that people have to do the tweak **manually** at this moment. 

Axios doesn't automatically include `X-CSRF-TOKEN` in the header (it sends `X-XSRF-TOKEN` instead). See https://github.com/laravel/passport/issues/256

We have to either change the documentation or merge https://github.com/laravel/laravel/pull/4128. What do you guys think?

monomichael:
https://github.com/laravel/laravel/pull/4128

brunen9:
I vote for code change so it will just work like it was with v-resource.

Quezler:
What i personally do not like about axios is that the 400 range requires a custom function to allow up to the 500 range, else it will throw a huge error.

Issue #3047, opened 2017-01-28
Homestead installation instructions do not mention MongoDB


Shelob9:
There is a section in the Homestead install docs for installing MariaDB, but not for installing MongoDB. 

See: https://github.com/laravel/homestead/blob/aac1631cf1f2fa3b3124c92d2786ddeaedc59ab4/scripts/homestead.rb#L205-L217

PR #3048, opened 2017-01-28
Document installation of MongoDB

Shelob9:
#3047

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -7,6 +7,7 @@
     - [Launching The Vagrant Box](#launching-the-vagrant-box)
     - [Per Project Installation](#per-project-installation)
     - [Installing MariaDB](#installing-mariadb)
+    - [Installing MongoDB](#installing-mongodb)
 - [Daily Usage](#daily-usage)
     - [Accessing Homestead Globally](#accessing-homestead-globally)
     - [Connecting Via SSH](#connecting-via-ssh)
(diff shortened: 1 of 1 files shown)


taylorotwell:
I wasn't even aware Homestead supports Mongo officially. To me it is something that should be managed in the `after.sh` script.

svpernova09:
We do support MongoDB via https://github.com/laravel/homestead/pull/424

Seemed like it was one of those "Well it's easy enough why not". We should probably reopen/accept this PR.

vesper8:
I wish I had known that Homestead supported MongoDB out of the box, I went through some pains installing it and will definitely be using it by having homestead manage it in the future. I had to search the issues on the Laravel Homestead repo to find out that it was already in there.. never thought to look in the scripts section.

If the support is already in then it should be made known that it's an option.

I just destroyed my box and recreated it with mongodb:true

I also modified the install script so it installs 3.4.x instead of 3.2

It worked perfectly.

the only thing though is that the installation instructions from this unmerged PR mention that if you install mongodb it will REPLACE mysql which is not the case.. and good thing too because I definitely want to use both mongodb and mysql. I guess the docs were copy/pasted from the mariadb instructions which is why it says that

DFurnes:
Agreed! I think anyone using [jenssegers/laravel-mongodb](https://github.com/jenssegers/laravel-mongodb) would really appreciate this being better documented. I only stumbled upon it when peeking into the Homestead repository on a whim.

deadlybutter:
👍 Would be nice to have this on the official documentation

PR #3051, opened 2017-01-29
Migrating Laravel Elixir to Laravel Mix

jettoki:
Failing to remove 'laravel-elixir' from package.json causes very confusing errors with Laravel Mix that cost me several hours of debugging. I have seen other people dealing with this issue, so I think it would be a good idea to include in the upgrade guide.


taylorotwell:
I added this originally but decided to remove it. I feel like if we're going to mention Mix here (which I'm not sure we should since it is not required for upgrading) it should be a more full migration guide.

PR #3052, opened 2017-01-29
update slot documentation

browner12:
passing strings through the slots is incredibly easy and well documented, but I ran into an issue when trying to pass PHP constructs through, like arrays or objects.  I was able to pass it as the second argument to the slot, and it worked fine.  I'm curious if this is the intended operation, and if so I'd like to add some documentation on it.

### DIFF
--- a/blade.md
+++ b/blade.md
@@ -109,7 +109,7 @@ Sometimes it is helpful to define multiple slots for a component. Let's modify o
         {{ $slot }}
     </div>
 
-Now, we can inject content into the named slot using the `@slot` directive. Any content is not within a `@slot` directive will be passed to the component in the `$slot` variable:
+Now, we can inject content into the named slot using the `@slot` directive. Any content not within a `@slot` directive will be passed to the component in the `$slot` variable:
 
     @component('alert')
         @slot('title')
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not sure what this even does :smile:

browner12:
I'm currently using it like this in a project, and it works fine.  The first parameter in the slot becomes the variable name, and the second parameter is the value.

I wasn't sure if this was intended, or simply dumb luck, so I wanted to run this by you first to see if I should continue using it this way.  And if not, if there is another way to pass in arrays or objects.

taylorotwell:
But the slot in this examle has no content? are you sure you didn't mean to pass the parameters to the component?

browner12:
It actually works both ways.

```php
@component('alert', ['options' => ['option1' => 'blue', 'option2' => 'red']])
@endcomponent
```

works the same as 

```php
@component('alert')
    @slot('options', ['option1' => 'blue', 'option2' => 'red'])
@endcomponent
```

I've also noticed that I can pass data through the `@extends` directive, and I can't remember if this was possible previously.

```php
@extends('master', ['key' => 'value'])
```

The way the directives are compiled are a little bit of a mystery to me, so I'm not sure if this is intended to work this way or not, although TBH I kind of like it.

browner12:
@taylorotwell I just noticed this has subsequently been removed from the documentation, and now you are documenting passing the data through the second parameter of the `@component` directive. why was this switched? now the `@component` is being used the same way as the `@include` directive is.

Could we *also* include a line about being able to pass the data to the `@component` **or** the `@slot`?

PR #3053, opened 2017-01-30
The Alias method

lioannou:
Ok - so I think the wording might be wrong for what I've written - but I dont know how to write it better. Rather than just create an issue for it, I wanted to start a PR and someone can write a better one if needed.

`alias()` is needed for when you want to use Lumen, Notifications and Mail. It address this issue https://github.com/laravel/lumen-framework/issues/503 - which is resolved with this command.

There might be other/better uses that could be put in the documentation as well.

p.s. Although this solves a Lumen issue - I'm doing the PR to the Laravel docs as the Lumen service docs basically says to look here :)

### DIFF
--- a/container.md
+++ b/container.md
@@ -9,6 +9,7 @@
 - [Resolving](#resolving)
     - [The Make Method](#the-make-method)
     - [Automatic Injection](#automatic-injection)
+    - [Alias](#alias)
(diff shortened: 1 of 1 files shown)


taylorotwell:
I would rather just fix it at the Lumen level. I don't really like container aliases.

PR #3054, opened 2017-01-30
Wrong array key

gauravmak:
It should be resource name instead of route url part


taylorotwell:
Two of these three things are wrong. The title link is correct.

gauravmak:
@taylorotwell I am curious how would it come to know that developer wants to set custom parameter name for **show** route here. And what should be specified for setting one for _edit, update or destroy_?

PR #3056, opened 2017-01-30
BrowserKit tests will need the BrowserKitTest namespace for migrations and transactions

bencarter78:
Just came across an issue where the DatabaseMigrations trait wan't being used. I noticed that it needs to be using the BrowserKit version. It was undocumented so this PR notes the namespace change need.


taylorotwell:
Those traits don't exist in the repository. Update to the latest browser kit testing.

PR #3067, opened 2017-02-01
fix order of $key $value in first() of collections

ahuggins:
Seems pretty minor, but the order of $key and $value makes a difference. The current docs, with $value first, actually returns 4 instead of the 3 it suggests. Switching the order returns 3 as expected.


ahuggins:
The reason the current docs returns 4 is because arrays are 0 based indexes.

ahuggins:
Actually, this looks like a change between versions. I am using 5.3.20. So nevermind. Sorry for any confusion on this. I think I was looking at current docs, but using older version.

PR #3070, opened 2017-02-02
Testing: Add a note about new namespace for traits

eblount:
The testing traits namespace has changed, so that `Illuminate\Foundation\Testing\DatabaseTransactions` is now `Laravel\BrowserKitTesting\DatabaseTransactions`.

If the old namespace is used, the `Laravel\BrowserKitTesting\TestCase::setUpTraits()` method does not set up the traits.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -382,6 +382,10 @@ Tests written on Laravel 5.3 will extend the `BrowserKitTest` class while any ne
 
 Once you have created this class, make sure to update all of your tests to extend your new `BrowserKitTest` class. This will allow all of your tests written on Laravel 5.3 to continue running on Laravel 5.4. If you choose, you can slowly begin to port them over to the new [Laravel 5.4 test syntax](/docs/5.4/http-tests) or [Laravel Dusk](/docs/5.4/dusk).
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
That isn't the case. I think you just need to update your browser-kit-testing package installation.

eblount:
If you're sure, that's cool.

Just in case, here's my experience and debugging info:

Last night, I upgraded to 5.4. I followed the Testing guide, installed `BrowserKitTesting`, and created a `BrowserKitTest.php` that extends `BaseTestCase`.

I have a base Integration testing class, and it now had the following relevant lines:
```
     use Illuminate\Foundation\Testing\DatabaseTransactions;
     abstract class IntegrationTestCase extends BrowserKitTest
     {
          use DatabaseTransactions;
```

I ran my test suite full of BrowserKit tests, and no migrations were done.

The relevant lines in `BrowserKitTesting\TestCase.php` are here:
https://github.com/laravel/browser-kit-testing/blob/master/src/TestCase.php#L107

The `$uses` array contains the full, namespaced class name:
```
$uses = array_flip(class_uses_recursive(static::class));
```

So in my case, it contained:
```
$uses['Illuminate\Foundation\Testing\DatabaseTransactions'] = 'Illuminate\Foundation\Testing\DatabaseTransactions'
```

It then compares that with a class name _in its own namespace_:
```
if (isset($uses[DatabaseMigrations::class])) {
    $this->runDatabaseMigrations();
}
```

Within that namespace, the isset is looking for an entry for a class with the full namespace:
```
Laravel\BrowserKitTesting\DatabaseTransactions
```

Changing my import statement to the new namespace caused the class names to equal and `runDatabaseMigrations()` to be called, and the migrations to run successfully:
```
use Laravel\BrowserKitTesting\DatabaseTransactions;
```

PR #3085, opened 2017-02-06
Document Passport::actingAs()

themsaid:


taylorotwell:
Merged. I don't understand why GitHub doesn't auto close these sometimes.

PR #3087, opened 2017-02-06
Update homebrew installation instructions

Flimm:
Both PHP and Composer need to be installed before installing Valet on macOS.

### DIFF
--- a/valet.md
+++ b/valet.md
@@ -44,7 +44,7 @@ However, you may extend Valet with your own [custom drivers](#custom-valet-drive
 
 As you may know, Laravel offers [Homestead](/docs/{{version}}/homestead), another local Laravel development environment. Homestead and Valet differ in regards to their intended audience and their approach to local development. Homestead offers an entire Ubuntu virtual machine with automated Nginx configuration. Homestead is a wonderful choice if you want a fully virtualized Linux development environment or are on Windows / Linux.
 
-Valet only supports Mac, and requires you to install PHP and a database server directly onto your local machine. This is easily achieved by using [Homebrew](http://brew.sh/) with commands like `brew install php71` and `brew install mysql`. Valet provides a blazing fast local development environment with minimal resource consumption, so it's great for developers who only require PHP / MySQL and do not need a fully virtualized development environment.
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not sure we must recommend installing Composer via Homebrew.

Flimm:
This is confusing to me. If installing PHP through Homebrew is no big deal, then why wouldn't the lesser concern of installing Composer through Homebrew be problematic?

In any case, there needs to be a step that says "install Composer", because the rest of the instructions don't work without Composer. Are you going to open a new pull request, or shall I?

michaeldyrynda:
The docs don't tell you how to install PHP, these are assumed pre-requisites. I don't think it's unreasonable to make some assumptions about what a PHP developer will already know before they start with the framework.

Flimm:
You say the docs don't tell you how to install PHP, but they actually do. They are not assumed pre-requsites, they are listed there explicitly in the documentation.

Here's one place where the docs tell you how to install PHP:

> Valet only supports Mac, and requires you to install PHP and a database server directly onto your local machine. This is easily achieved by using [Homebrew](http://brew.sh/) with commands like `brew install homebrew/php/php71` and `brew install mysql`. Valet provides a blazing fast local development environment with minimal resource consumption, so it's great for developers who only require PHP / MySQL and do not need a fully virtualized development environment.
 

Here's another place where the docs tell you how to install PHP:

> - Install or update [Homebrew](http://brew.sh/) to the latest version using `brew update`.
> - Install PHP 7.1 using Homebrew via `brew install homebrew/php/php71`.
> - Install Valet with Composer via `composer global require laravel/valet`. Make sure the `~/.composer/vendor/bin` directory is in your system's "PATH".

All I want is to add one line explaining how to install Composer, since it doesn't make sense to explain how to install PHP and MySQL with Homebrew, but not Composer, which is required for the next step in the explicitly documented process.

If the docs are not meant to tell you how to install PHP, should I submit a pull request removing that information, or will you?

michaeldyrynda:
In the context of Valet, yes, it makes sense to provide instruction on installing PHP. It doesn't tell you how to do so in the Laravel install docs. 

The Homebrew install is also Mac-specific; what about Linux or Windows? The composer docs tell you how to install it. Should suffice.

Flimm:
This pull request is only about Valet, it's not about Laravel.

Valet is Mac-specfic.

Now that we've realised that this closure is based on a misunderstanding, can we re-open this pull request?

PR #3093, opened 2017-02-07
Mention adding the Tests namespace

bryannielsen:
I wasn't sure if this information should go under the Testing section or up by the composer.json changes so feel free to move or change wording.  I encountered this today on a 5.3 -> 5.4 upgrade and thought it would be worth calling out in the upgrade notes.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -11,7 +11,14 @@
 
 ### Updating Dependencies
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I've added a section on this to the testing section of the upgrade guide.

PR #3094, opened 2017-02-08
Documented skip symlinks option

robjbrain:
Added more information on the optional skip symlinks config for the local driver

### DIFF
--- a/filesystem.md
+++ b/filesystem.md
@@ -47,6 +47,16 @@ Of course, once a file has been stored and the symbolic link has been created, y
 When using the `local` driver, all file operations are relative to the `root` directory defined in your configuration file. By default, this value is set to the `storage/app` directory. Therefore, the following method would store a file in `storage/app/file.txt`:
 
     Storage::disk('local')->put('file.txt', 'Contents');
+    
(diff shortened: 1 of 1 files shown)


taylorotwell:
We would need more information about the specific error to give it more clarity. Thanks.

robjbrain:
The error is:

> NotSupportedException in NotSupportedException.php line 21:
> Links are not supported, encountered link at 

This is from Flysystem here: https://github.com/thephpleague/flysystem/blob/master/src/Adapter/Local.php#L429

I wasn't sure what the protocol was for documenting an error from another package hence why I left it vague.

At the very least the documentation could be changed to simply document the config option without referencing an error:

```
The local driver has an optional flag allowing you to skip symlinks by setting the `links` config to `skip`:

    'local' => [
        'driver' => 'local',
        'root'   => storage_path().'/app',

        // Optional Driver Settings...
        // 'links' => 'skip',
    ],
```

I've submitted a commit just documenting the config option, without referencing an error here: https://github.com/robjbrain/docs/commit/d99a18082e9f6bea8580b7dfa01be24514230c1d

PR #3099, opened 2017-02-09
Override plural by the range

cjmaxik:
Since https://github.com/laravel/framework/pull/17826 bring back the plurals, I find that Symfony's construction like `{0} There are none|There are one|There are many` is no longer work. However, it works in opposite direction now (like in PR). Since there is no mention how user can do that later, we can add this now.

Also it should works in other languages (like Russian) with more complex plural:
`'comments_count' => ':count комментарий|:count комментария|:count комментариев|{0} Нет комментариев',`
0 => Нет комментариев
1 => 1 комментарий
2 => 2 комментария
5 => 5 комментариев

### DIFF
--- a/localization.md
+++ b/localization.md
@@ -134,6 +134,10 @@ You may even create more complex pluralization rules which specify translation s
 
     'apples' => '{0} There are none|[1,19] There are some|[20,*] There are many',
 
+You may even override some ranges of plural rules by putting it after all the line:
+
+    'apples' => 'There are one|There are many|{0} There are none',
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't know what this means.

taylorotwell:
This is already demonstrated right above your example.

cjmaxik:
This is not the same. Example right above will not use internal plural (https://github.com/laravel/framework/pull/17826/files/b2d997dcfb73347670e6261f6d6cae3980aa9181#diff-4b78d7d05e662d3ea18c7e872456ec96R111)

PR #3100, opened 2017-02-09
Keep consistency in after*/before* validation rules

paulofreitas:
This PR aims to improve consistency in the `after_or_equal`/`before`/`before_or_equal` validation rules documentation making it clear that all these rules accepts a date string or another field to compare against.

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -593,7 +593,7 @@ Instead of passing a date string to be evaluated by `strtotime`, you may specify
 <a name="rule-after-or-equal"></a>
 #### after\_or\_equal:_date_
 
-The field under validation must be a value after or equal to the given date. For more information, see the [after](#rule-after) rule.
+The field under validation must be a value after or equal to the given date. Dates are evaluated as in the [`after`](#rule-after) rule.
 
 <a name="rule-alpha"></a>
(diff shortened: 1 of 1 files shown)


taylorotwell:
Let's just specify exactly what the behavior is instead of referring them to another rule. If you have time, another PR with that approach would be appreciated.

paulofreitas:
@taylorotwell Ah right, I tried to follow the same convention used on `between`, `max` and `min` validation rules, where they all link to the `size` rule for evaluation details.

Would it be fine documenting the `before` rule in detail as in `after` rule, leaving the others linking to their similar rules, thus avoiding repeating much?

Like this:

> **after:_date_**
> *Leave it as is.*
>
> **after_or_equal:_date_**
> *Leave it as is.*
>
> **before:_date_**
> *Detail the behavior and include examples as in `after` rule.*
>
> **before_or_equal:_date_**
> *Link to `before` as in `after_or_equal` rule.*

It seems to me to improve understanding while keeping the brevity of explanations.

Does it seems reasonable to you? :)

taylorotwell:
Even if you have to repeat yourself, just do it.

PR #3103, opened 2017-02-09
Added two missing methods and their replacements.

MarkVaughn:
During upgrading these methods were not mentioned and needed to dig around to figure out the replacement.


taylorotwell:
These notes should be under a routing section. The test one is not needed since it is noted the entire test component was re-written in the testing section. Thanks.

taylorotwell:
I've added these changes to routing section.

MarkVaughn:
Thanks!

PR #3108, opened 2017-02-11
[5.5] Recommend the null coalescing operator

GrahamCampbell:
This removes the docs for the `or` functionality blade provides, and recommends the use of PHP's built in null coalescing operator.

### DIFF
--- a/blade.md
+++ b/blade.md
@@ -148,13 +148,9 @@ Of course, you are not limited to displaying the contents of the variables passe
 
 #### Echoing Data If It Exists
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Thanks

ergonomicus:
But why?

browner12:
Laravel 5.5 will require PHP > 7.0, which has the null coalescing operator.  The Blade `or` syntax was just a way to create this functionality before it existed in PHP.

I believe Taylor or Graham mentioned it would be removed or deprecated in Laravel 5.5 as well.

ergonomicus:
Very true, but I just find it more pleasant to see `or` instead of `??`, that was my concern.

PR #3110, opened 2017-02-11
Make it clearer when implicit model binding occurs

Juddling:
Implicit model binding occurs when the parameter is type hinted


taylorotwell:
Wrong branch. I'll make the changes on 5.4.

PR #3112, opened 2017-02-12
Change route regular expression constraints parameter

foo99:
I changed it to a little realistic example. :smiley:


taylorotwell:
I think the original makes the point fine.

PR #3113, opened 2017-02-12
Add an example of use of `make` method to make the change more visible

mrliptontea:
Hi,

I missed this change when I was reading upgrade guide on Friday. Found out about it the hard way on the production today!

I hope that by providing an example, it'll be more clear, that `$container->make()` now accepts only one argument.


taylorotwell:
That's not the only way to solve this.

PR #3115, opened 2017-02-14
Update http-tests.md

dabuno:
Tests are now using namespaces. We need to access global scope for User::class


taylorotwell:
Merged.

Issue #2407, opened 2016-07-12
Document framework events


ejunker:
In the 5.1 documentation the Framework Events are listed at https://laravel.com/docs/5.1/events#framework-events

I could not find a similar list of events for 5.2. I do know that many/all of the framework events changed from string event names to objects but it would be nice to have a table of them like we did for 5.1

GrahamCampbell:
You can find the new event names at https://laravel.com/docs/5.2/upgrade#upgrade-5.2.0. I assume the list was dropped because the new classes are self-documenting so you no longer need to guess what the parameters are now. I'm not at all opposed to a PR to the docs to add something similar back in, infact, I'd quite like to see that. Maybe give Taylor a ping to check he's ok with this before sending a PR though. He might have different ideas. :)

ejunker:
@taylorotwell would you be interested in a PR to add a table of framework events like there was for 5.1? Which branch would this go on?
https://laravel.com/docs/5.1/events#framework-events

azamtav:
I definitely agree that the events should be added to the documentation. I starting writing unnecessary events to my project because I didn't know there were already built in ones for authentication. It took me a little while to actually find the events in the source code.

Also, I always wondered why the Model Events are listed under the Eloquent section instead of under the Events section. I tend to check the Events page first, and then I have to remember that Model Events is documented under the Eloquent section.

laurencei:
I'm going to close this ticket here. If anyone finds any events that are undocumented, please feel free to submit a PR with the event so it can be added to the docs.

PR #3117, opened 2017-02-14
Updated Events Documentation

azamtav:
- Added Framework Events back into the Events documentation
- Moved the Model Events from the Eloquent section into the Events
section

See Issue #2407


azamtav:
Copying my commentary from Issue #2407 here:

I definitely agree that the events should be added to the documentation. I starting writing unnecessary events to my project because I didn't know there were already built in ones for authentication. It took me a little while to actually find the events in the source code.

Also, I always wondered why the Model Events are listed under the Eloquent section instead of under the Events section. I tend to check the Events page first, and then I have to remember that Model Events is documented under the Eloquent section.

taylorotwell:
Model events should be with Eloquent.

PR #3123, opened 2017-02-17
Added valet information for common installation issue with macOS Sier…

itguy614:
…ra.  re: sudo valet install


taylorotwell:
I don't really like adding these little caveats to the docs each time some strange issues occur. They would fill up pretty fast :laugh:

itguy614:
Understandable.  I know this hit me twice in a month so thought having it in the doc's instead of digging up a past GitHub issue might be beneficial.  Take care!

PR #3126, opened 2017-02-18
Hint at change of default billable model

richartkeil:
If people use another model for billing than `App\User` they'll have problems after upgrading from 5.2 to 5.3 because Cashier looks for a configured model now.
Therefore they should rename the column `user_id` in the `subscriptions` table to the appropriate name.


taylorotwell:
This isn't really all you would have to do I don't think.

PR #3127, opened 2017-02-18
Add dropTimestamps description to migrations docs

jougene:


taylorotwell:
Honestly people can usually just use ->drop.

PR #3134, opened 2017-02-22
Instructions on how to perform one-time charges on new Stripe subscriptions

se1exin:
Sorry, I should have given this PR a bit more context.

The changes are to update the Cashier (billing) documentation, to both highlight the ability to create Stripe Invoice Items using the `tab` method, and to detail how this can then be used to create initial one-off charges when establishing a subscription, such as a sign up fee that only gets charged at the beginning of a subscription.

All feedback is welcome if you feel the addition isn't applicable or should be changed in any way

### DIFF
--- a/billing.md
+++ b/billing.md
@@ -224,6 +224,31 @@ If you would like to apply a coupon when creating the subscription, you may use 
          ->withCoupon('code')
          ->create($stripeToken);
 
+#### One-time Subscription Charges
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
In my opinion I would make this entirely different section on `tab` since it can be used outside of this context as well to add items to an invoice during the billing cycle. It's not just limited to adding things before creating a subscription.

PR #3139, opened 2017-02-23
Adding gc_collect_cycles tip

weotch:
I found that **without** explicitly calling gc_collect_cycles() at the end of my job's `handle()`, my worker would continue to consume memory until it died.  This could be due to the PHP configuration of my host but, nonetheless, this tip could save folks in a similar situation some time.


taylorotwell:
That is just masking some other memory leak in the application IMO.

PR #3143, opened 2017-02-24
Indicate when HTTP Test assertions were added

Indemnity83:
Some of the available assertions were added after the initial 5.4.0 release. There was no indication of this at all in the documentation which may cause confusion.


taylorotwell:
No plans on adding this.

PR #3146, opened 2017-02-25
Observers != Events

markhuot:
It appears, in my testing that if you're using the `$events` property to define "traditional" events then the Observable class will not fire. E.g.,

```php
class Document extends Model
{
    protected $events = [
        'saved' => DocumentSaved::class,
    ];

    public function boot() {
        static::observe(DocumentObserver::class);
    }
```

```php
class DocumentObserver
{

    public function saved(Document $document)
    {
        dd('does not fire');
    }
```

I uncovered this while simplifying some code. I was in the process of removing an event for an observable and found it it didn't fire until I removed the `$events` property. Not sure if this is intended but the documentation update should make it a little more clear for future people.

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -736,9 +736,11 @@ To get started, define an `$events` property on your Eloquent model that maps va
     }
 
 <a name="observers"></a>
-### Observers
+## Observers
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
This was more of a bug. I've fixed it and the fix will be available in the next release.

markhuot:
Great, thanks!

PR #3154, opened 2017-02-28
vendor/bin/homestead is a shell script - removing unnecessary PHP call

gtoran:
In the "Per project installation" chapter, the **Mac/Linux** section indicates a user should execute the command `php vendor/bin/homestead`, implying that homestead is a PHP script rather than a shell script. Executing it as php will simply output the script contents to the screen.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -169,7 +169,7 @@ Once Homestead has been installed, use the `make` command to generate the `Vagra
 
 Mac / Linux:
 
-    php vendor/bin/homestead make
+    vendor/bin/homestead make
 
 Windows:
 


taylorotwell:
You can call it with PHP.

taylorotwell:
Confirmed the current docs are correct.

gtoran:
Hi Taylor, thanks for getting back to me. I have to admit I was perplexed when I got the error (having been used to running the command with php), so that's why I dug in and found that my `vendor/bin/homestead` file was in fact a shell script (attached below for reference). I thought the method had changed, that's why I issued a pull request.

Upon reading your answer, I ran a fresh install and got the expected PHP script, which can be invoked either with or without `php` as the docs state & you mentioned.

I initialized this Laravel project last Friday on Homestead 1.0.1, having moved it to its own project Homestead environment today. The only difference is that laravel/homestead is 5.0.1 in my new test project (being 5.0 in the project where I experienced this). I cannot reproduce this again nor do I understand clearly why I got this version of the file.

Code:

```shell
#!/usr/bin/env sh

dir=$(d=${0%[/\\]*}; cd "$d"; cd '../laravel/homestead/bin' && pwd)

# See if we are running in Cygwin by checking for cygpath program
if command -v 'cygpath' >/dev/null 2>&1; then
	# Cygwin paths start with /cygdrive/ which will break windows PHP,
	# so we need to translate the dir path to windows format. However
	# we could be using cygwin PHP which does not require this, so we
	# test if the path to PHP starts with /cygdrive/ rather than /usr/bin
	if [[ $(which php) == /cygdrive/* ]]; then
		dir=$(cygpath -m "$dir");
	fi
fi

dir=$(echo $dir | sed 's/ /\ /g')
"${dir}/homestead" "$@"
```

Thanks and apologies for the non-issue.

PR #3155, opened 2017-02-28
Adding doc for Elixir directing user to Mix

votemike:
The search bar in the Laravel docs do not throw up any result when searching for Elixir. It took me a while to figure out that it had been renamed. I imagine other people have had the same problem.


taylorotwell:
No plans on adding this.

PR #3156, opened 2017-02-28
added note about mix.copy third parameter

dfreer:


taylorotwell:
Me and @JeffreyWay had a quick chat about this and we'd like to introduce a new function that does this instead of having to specify a `false` flag since that isn't very communicative of the intent.

JeffreyWay:
This is on master now. https://github.com/JeffreyWay/laravel-mix/commit/23a91bed6281bdbb54bb7eaebe55523b7f2c0265

PR #3166, opened 2017-03-03
Added note about the flatten argument of .copy()

LiamHammett:
The documentation isn't very clear about all of the arguments of the .copy() method, and its default behaviour is to flatten the directory structure, which is not necessarily expected.

Mix's own documentation does mention this in a footnote. https://github.com/JeffreyWay/laravel-mix/blob/master/docs/copying-files.md


taylorotwell:
We are changing this on Mix and will update the documentation then. /cc @JeffreyWay

PR #3177, opened 2017-03-07
Add additional password reset conditions section

lasseal:
This is to reflect the "Specifying Additional Conditions" section under
authentication which gives an example about blocking inactive users from
authenticating. By not also adding the condition on the
ForgotPasswordController inactive users can authenticate by simply
requesting and performing a password reset.

### DIFF
--- a/passwords.md
+++ b/passwords.md
@@ -95,3 +95,13 @@ You may easily modify the notification class used to send the password reset lin
         $this->notify(new ResetPasswordNotification($token));
     }
 
+#### Specifying Additional Conditions
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
This wouldn't actually work. You would need more logic than that.

lasseal:
Shoot, you're right of course. How I'm using it is to check the subdomain to see of the user has access. 

```
$response = $this->broker()->sendResetLink(
            $request->only('email') + ['city_id' => $request->city->id]
        );
```

I thought to generalise the example in the docs, but of course more logic would be needed for the example in the PR. I could submit another PR if this is of interest to have in the docs.

PR #3179, opened 2017-03-08

sobers-2002:
Show example of the quick select syntax for eager loading


taylorotwell:
Going to hold off on this for now because I feel it will be a source of problems with people forgetting the right columns.

PR #3185, opened 2017-03-09
Added kebab_case() to the string helpers

TJSoler:
Was this method not documented so it will not break the 6 methods per column on 3 columns?


taylorotwell:
That is not a reason we avoid documenting things.

TJSoler:
I guess I should have /s that comment. :flushed:

taylorotwell:
😆

PR #3189, opened 2017-03-10
Explained which headers to include when using other javascript frameworks

MrMashy:
Addition to "Consuming your API with Javascript"


MrMashy:
Make sure to view all subsets

taylorotwell:
There is already a note that you should do this.

MrMashy:
I must say it took me a while to figure out what exact headers to use and what the header value should even be. This let's users figure it out.

PR #3190, opened 2017-03-10
Fixed typo in upgrade doc

jrhenderson1988:
Changed "If you will still like to customize" to "If you would still like to customize"


taylorotwell:
Thanks

PR #3193, opened 2017-03-13
Added description for getMessage() in Error pages

ron4stoppable:
$exception variable has methods for displaying the message passed by the abort function, which can be useful in Custom HTTP Error Pages. The proposed changes mention this method with an example for blade template under - "Custom HTTP Error Pages" section.


ron4stoppable:
thanks @taylorotwell for accepting changes.

PR #3202, opened 2017-03-15
update to Echo install section to account for 1.3.0 change

philbenoit:
With the recent change in 1.3.0 there is a doc update at the top but it should also be added here. Some users will see the skip links at the top and navigate directly down the page, missing the initial install instructions.


taylorotwell:
It doesn't make sense to put things in the docs because users might "miss" other things in the docs. We would have to put little reminders everywhere. :smile:

philbenoit:
Understand your position but isnt that the purpose of documentation. To outline how things work. You have two example on the same page showing two different things. I will keep my eye out for other updates, seems Echo is in flux at the moment. 😄

PR #3203, opened 2017-03-16
Camel cased relationship names require a snake case column name when using withCount

ryrobbo:
I reported the below issue when debugging a possible bug in Laravel. Column names for the count property require snake case naming convention. Had this caveat been in the the documentation, I would have amended accordingly. 

https://github.com/laravel/framework/issues/18381

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -687,6 +687,8 @@ You may also alias the relationship count result, allowing multiple counts on th
     echo $posts[0]->comments_count;
 
     echo $posts[0]->pending_comments_count;
+    
+> {tip} If your relationship names are camel cased, you will need to use the snake case version when referencing the column name. For example, if you had a relationship on your model named `authorNotes`, your column name will be `author_notes_count`.
 
 <a name="eager-loading"></a>
 ## Eager Loading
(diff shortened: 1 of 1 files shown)


taylorotwell:
This seems to already be demonstrated in the documentation's examples.

ryrobbo:
Could you tell me where? I have been referencing this https://laravel.com/docs/5.4/eloquent-relationships#counting-related-models and there is no mention of camel casing not working.

devcircus:
It looks like the only example in the docs of using snake_case in this situation is when mentioning the ability to alias. Seen below:
```
$posts = Post::withCount([
    'comments',
    'comments AS pending_comments' => function ($query) {
        $query->where('approved', false);
    }
])->get();

echo $posts[0]->comments_count;

echo $posts[0]->pending_comments_count;
```
However, as @ryrobbo mentioned, if your relationship name has multiple words, as in "magazineSubscriptions", you will need to snake case. For example, 
```
echo $users[0]->magazine_subscriptions_count;
```
I agree that it would be helpful to note this in the documentation.
ping @themsaid

ryrobbo:
Thanks @devcircus 

I must have wasted a good hour re-reading the docs and sanity checking my Eloquent model. With 5.4 gathering momentum we could see a lot more devs running into this issue.

I am happy to modify my changes to something more suitable if required.

ryrobbo:
@themsaid @taylorotwell 

I appreciate you guys maybe busy but is this issue going to be ignored?

PR #3210, opened 2017-03-19
No semver compliance

slavicd:
It's important to know that if you have 5.1, you can't just run a command and your site will run 5.4. That takes painful work.


taylorotwell:
We really need a more full explanation of versioning on a separate page. I will write that. Adding to my todo list for next week.

PR #3212, opened 2017-03-20
Simplify examples

JosephSilber:


taylorotwell:
I think this line is a little too long for the docs. Can we slim it down?

PR #3213, opened 2017-03-20
Express why the event dispatcher will halt when your observer contain…

Caroga:
Express why the event dispatcher will halt when your observer contains any other return value than null;

I've stumbled upon this situation where stacking observers did not work. Turns out that the events dispatcher will halt if your observer returns something that is not `null`.
https://github.com/laravel/framework/blob/5.4/src/Illuminate/Events/Dispatcher.php#L204

This is not made clear in the documentation, only with a docblock in the example with a return type void. I've added a line to express the necessity of this example.

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -771,6 +771,8 @@ If you are listening for many events on a given model, you may use observers to 
         }
     }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think it more likely depends on the specific event being fired. Some halt and others don't, so this wouldn't exactly be correct I don't think.

Issue #3209, opened 2017-03-19
[Bug] Pivot class should be extended as UserRole, not as Role


mpyw:
# Wrong

```php
class Role extends Pivot
{
    public function users()
    {
        return $this->belongsToMany('App\User')->using('App\UserRole');
    }
}
```

# Correct

```php
class Role extends Model
{
    public function users()
    {
        return $this->belongsToMany('App\User')->using('App\UserRole');
    }
}
```

```php
class UserRole extends Pivot
{
}
```

connectkushal:
@mpyw its corrected in the master branch of docs and the error is in 5.4 branch.

Another thing wrong in 5.4 the namespace in the example code

```<?php

namespace App;

use Illuminate\Database\Relations\Pivot;
```

It should be `use Illuminate\Database\Eloquent\Relations\Pivot;`

mpyw:
Thanks

PR #3215, opened 2017-03-21
Corrected wrong example code in sync with master branch

connectkushal:
The master branch has updated the docs, making the same changes in 5.4

https://github.com/laravel/docs/blob/master/eloquent-relationships.md#many-to-many


connectkushal:
#3209 suggestions to further extend the example code.

PR #3218, opened 2017-03-22
Adding ::class

shealan:
Adding the ::class part of the service provider name as Laravel 5.4 fails without it.

### DIFF
--- a/billing.md
+++ b/billing.md
@@ -117,7 +117,7 @@ First, add the Cashier package for Braintree to your `composer.json` file and ru
 
 #### Service Provider
 
-Next, register the `Laravel\Cashier\CashierServiceProvider` [service provider](/docs/{{version}}/providers) in your `config/app.php` configuration file.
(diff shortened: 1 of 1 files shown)


lagbox:
`'Laravel\Cashier\CashierServiceProvider' == Laravel\Cashier\CashierServiceProvider::class`

browner12:
yah, no need for the `::class`.  all that does is return a string of the class name.

shealan:
Step 1: Install a copy of Laravel 5.4.
Step 2: Add Cashier.
Step 3: Add `Laravel\Cashier\CashierServiceProvider` into the config.
Step 4: Kills the project dead.

Tried on Homestead and Valet on multiple machines.

If you find otherwise please report back.

browner12:
do you have it in quotes?

check here 

https://github.com/laravel/laravel/blob/3516f4f6771bd1817d08b529549c4f4a449e672f/config/app.php

for an old version of the config file, where they still used strings.

taylorotwell:
I dunno, I think the current wording is correct. That is the class name of the service provider you want to add to the array. It's up to you if you want to use a string or use the ::class constant.

PR #3235, opened 2017-03-28
CORS with Axios

suhaibmalik:
Took me way too long to figure this one out. Might be cleaner as a subsection but I believe this aught to be mentioned here.


taylorotwell:
Laravel documentation should not be explaining deeper Axios features. That is for the Axios documentation to handle.

PR #3243, opened 2017-04-02
Update code in facades.md for simplicity

connectkushal:
`use Cache;` is too confusing without the full namespace, to be in the documentation. It looks like an error or distracts the users from the topic, unless there is an extra line added, talking about AliasLoader which will check the facade class from array of class aliases in config/app.php.


connectkushal:
its confusing without it, to be in the docs. Looks like an error. Or there should be an extra line stating the aliasclass loader will load the cache from app/config.php

PR #3245, opened 2017-04-03
[5.3] Added deleteFileAfterSend() back in the docs

concept-core:
Reason to add this back is because it is handy to know the functionality already sits in the framework.


taylorotwell:
I'll just merge this into the latest docs.

PR #3249, opened 2017-04-04
[5.4] Improve the environment configuration documentation

paulofreitas:
- Moves two general-topic paragraphs from *"Retrieving Environment Configuration"* sub-section into the main *"Environment Configuration"* section
- Adds a security note about committing `.env` files
- Adds a general note to clarify that `.env` files variables can be overridden by external environment variables
- Makes the *"Retrieving Environment Configuration"* sub-section linkable
- Adds a tip to the *"Determining The Current Environment"* sub-section to mention that the `APP_ENV` variable could be overridden at server-level configuration

These are issues I often see when answering questions about Laravel, so I think they're a good addition to make the environment configuration documentation even better. 👍 

I'm certainly not very skilled at writing in english, so feel free to improve wording mistakes I might made. :smile:

### DIFF
--- a/configuration.md
+++ b/configuration.md
@@ -20,7 +20,7 @@ It is often helpful to have different configuration values based on the environm
 
 To make this a cinch, Laravel utilizes the [DotEnv](https://github.com/vlucas/phpdotenv) PHP library by Vance Lucas. In a fresh Laravel installation, the root directory of your application will contain a `.env.example` file. If you install Laravel via Composer, this file will automatically be renamed to `.env`. Otherwise, you should rename the file manually.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I really don't like to put notes or tips back to back against each other. I just think it looks too busy. Is there another place we can show the tip?

paulofreitas:
@taylorotwell Sorry for not being able to respond in time, this was a really pertinent remark and I see here you have tackled this issue very well in the last commit! Thanks for merging and especially for making it even better! 😉 

The environment configuration topic had a good improvement! 🎉

PR #3251, opened 2017-04-05
Describe Blade or behaviour

thijsvdanker:
The `or` behaviour to use defaults in blade does not work when the first parameter is not a variable.
This is caused by the regex
```php
return preg_replace('/^(?=\$)(.+?)(?:\s+or\s+)(.+?)$/s', 'isset($1) ? $1 : $2', $value);
```

in [CompilesEchos.php](https://github.com/laravel/framework/blob/5.4/src/Illuminate/View/Compilers/Concerns/CompilesEchos.php#L103)
 and discussed in https://github.com/laravel/framework/issues/17754

It's behaviour is not exactly like the ternary statement earlier in the docs, so a little clarification on what does and what doesn't work might help someone.

### DIFF
--- a/blade.md
+++ b/blade.md
@@ -158,6 +158,8 @@ However, instead of writing a ternary statement, Blade provides you with the fol
 
 In this example, if the `$name` variable exists, its value will be displayed. However, if it does not exist, the word `Default` will be displayed.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm just removing that section from the docs entirely. PHP now has the ?? operator which can be used like:

```
{{ $name ?? 'default' }}
```

PR #3253, opened 2017-04-05
Clarify 2nd argument of firstOrCreate (5.3)

wmcmurray:
In 5.3, `firstOrCreate()` [has a second parameter](https://github.com/illuminate/database/blob/5.3/Eloquent/Builder.php#L240) while `firstOrNew()` don't yet.
Clarify that in the example.


taylorotwell:
I'll look at the 5.4 PR.

wmcmurray:
Ok but the 5.3 doc will still be a little bit ambiguous :)

PR #3254, opened 2017-04-05
Add contextual binding tip

jmsfwk:
Note that contextual binding is only available for constructors.


taylorotwell:
AFAIK we never demonstrate any other way to use them.

PR #3258, opened 2017-04-07
Document the Mailable callbacks functionality

hskrasek:
This PR aims to add documentation for the callback functionality that exists on Mailable objects.


taylorotwell:
You're supposed to use the `->withSwiftMessage(function ($message) {});`, not use `$callbacks` directly.

hskrasek:
@taylorotwell That makes sense. I can update the title and all that, and instead document the `withSwiftMessage` method instead

taylorotwell:
Thanks

PR #3265, opened 2017-04-11
Update query builder JSON docs

HipsterJazzbo:
Removed MariaDB 10.2.3 from the list of databases that support JSON columns. It does not, and likely won't until 10.3 at least. 10.2.3 *does* include JSON *functions*, but not a JSON column type.

To confirm, I installed a fresh install of MariaDB 10.2.5, created a new Laravel app, added a migration that included `$table->json('test')`, and confirmed that the migration failed.

### DIFF
--- a/queries.md
+++ b/queries.md
@@ -398,7 +398,7 @@ The query above will produce the following SQL:
 <a name="json-where-clauses"></a>
 ### JSON Where Clauses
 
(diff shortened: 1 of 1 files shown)


HipsterJazzbo:
Related: laravel/framework#13622

PR #3269, opened 2017-04-14
Redirect to named route in session flash example

svenluijten:
The previous example (`return redirect('dashboard')->with('status', 'Profile updated!');`) doesn't actually work. This is fixed by using the `->route()` method on the redirect class.

### DIFF
--- a/redirects.md
+++ b/redirects.md
@@ -76,7 +76,7 @@ Redirecting to a new URL and [flashing data to the session](/docs/{{version}}/se
     Route::post('user/profile', function () {
         // Update the user's profile...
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
The original code example does work though. :/ ... maybe it's an application problem?

svenluijten:
As far as I can see the `redirect()` helper only takes a URI as an argument, not the name of a route. Or am I misreading the example and should it instead be `/dashboard`?

eliurkis:
You should have an URL `dashboard`, in that way the original code example will work.

PR #3270, opened 2017-04-15
Document "flatten" parameter of mix.copy()

phanan:
The third (`flatten`) parameter of `mix.copy()` is weirdly set to `true` by default, causing [some confusion](https://laracasts.com/discuss/channels/elixir/copy-folder-structure-with-laravel-mix-laravel-54) to users. This PR aims to at least document this.


taylorotwell:
We actually have a `copyDirectory` method. I have documented it.

phanan:
Thanks. For at least `"laravel-mix": "^0.8.0"` though, `mix.copyDirectory` is not a function.

PR #3272, opened 2017-04-16
Update copy method usage.

jaesung2061:


taylorotwell:
Why are you changing this. copyDirectory is the correct method.

jaesung2061:
Sorry, turns out I didn't have the latest version.

PR #3275, opened 2017-04-16
Improvised container docs

introwit:
- Added a bit more info about what is dependency injection since understanding DI well will help people to understand the service container better

- Added a few lines of simple definition about what the container is actually doing for the developer

- Edited the explanation of the given example to make the most of it to understand where the container comes into action

### DIFF
--- a/container.md
+++ b/container.md
@@ -14,7 +14,9 @@
 <a name="introduction"></a>
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't think I'm going to include this at the moment. Thanks though.

PR #3280, opened 2017-04-21
fix errors

baooab:


taylorotwell:
This is not correct in some aspects.

PR #3282, opened 2017-04-23
fix(Broadcasting): typo in “leaving a channel”

Haroenv:
Title should be `Leaving a channel`, not `Leaving A Channel`


taylorotwell:
I capitalize every word / letter in those sections.

Haroenv:
Okay, that’s fine, it confused me, because I’m not used to seeing “title case”

PR #3287, opened 2017-04-25
Clarify what queue:work --sleep exactly does

NiclasHedam:
I've had a discussion with a fellow Laravel Developer, where we actually disagreed on the actual meaning of the --sleep option in the Queue Worker.
This should clarify the expected behaviour.

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -396,7 +396,7 @@ The `retry_after` configuration option and the `--timeout` CLI option are differ
 
 #### Worker Sleep Duration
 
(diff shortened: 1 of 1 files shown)


laurencei:
Is that not very similar to what is already there? 

> However, the `sleep` option determines how long the worker will "sleep" if there are no new jobs available.

NiclasHedam:
We disagreed on the meaning of sleep. It's ambiguous. It can mean that it will stop processing in sleep seconds on empty queue (correct) or that it will wait for new jobs in the empty queue for sleep seconds before exiting. Both are probable in the Laravel domain.

I think that it should be clearly stated that the worker will not wake up if a new job appears, which in my opinion would be the expected behavior.

PR #3288, opened 2017-04-26
Show how to use Mix with environment variables

adiachenko:
See [this PR](https://github.com/JeffreyWay/laravel-mix/pull/464).

### DIFF
--- a/mix.md
+++ b/mix.md
@@ -20,6 +20,7 @@
 - [Versioning / Cache Busting](#versioning-and-cache-busting)
 - [Browsersync Reloading](#browsersync-reloading)
 - [Notifications](#notifications)
+- [Environment variables](#environment-variables)
 
 <a name="introduction"></a>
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't know what "you should not access process directly" means.

adiachenko:
@taylorotwell It means that this won't work:

```
const x = process
console.log(x.env.MIX_VARIABLE)
```
But this does:
```
console.log(process.env.MIX_VARIABLE)
```

Not sure how to phrase it properly. Basically, Webpack's DefinePlugin  substitutes requested string with new value, it doesn't actually populates `process` object in a way you'd expect using Node.

adiachenko:
Ok, I edited PR to make instructions less open to interpretations while removing the note about `process` object to avoid causing unnecessary confusion (most people will never reach out to it in browser environment anyway).

PR #3290, opened 2017-04-28
Simplify Inconsistencies

ConnorVG:
`App\Flight` and `Flight` are used sort of erratically in this document.


ConnorVG:
It's actually annoyingly difficult to decide which direction to go with `App\Flight` vs `Flight` so... Screw it.

PR #3293, opened 2017-05-03
Include the simpler Closure technique for Mail::send()

orrd:
The Mailable way to send emails adds an extra layer of complexity to sending emails, involving creating a separate Mailable class (possibly having to create one for each kind of email your app sends). That's a significant amount of extra code for something that used to be simple. For a lot of situations, developers would prefer a way to simply send an email without creating a separate dedicated class just to send an email.

The original pre-5.3 way to use Mail::send() still works, it just isn't in the documentation any more. This pull restores the documentation of the simpler use of Mail::send(), while also still explaining Mailable for devs that need or prefer to create a separate class to handle sending an email.

### DIFF
--- a/mail.md
+++ b/mail.md
@@ -2,6 +2,9 @@
 
 - [Introduction](#introduction)
     - [Driver Prerequisites](#driver-prerequisites)
+- [Sending Mail](#sending-mail)
+    - [Attachments](#attachments)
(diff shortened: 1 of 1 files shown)


taylorotwell:
I would prefer to just document Mailables at this time.

PR #3296, opened 2017-05-04
Using echo with custom broadcastAs

mbuk:
When you use echo in conjunction with a customer broadcastAs event name you need to prepend the event name in your javascript with a '.'


mbuk:
See here for a discussion https://github.com/tlaverdure/laravel-echo-server/issues/116

PR #3297, opened 2017-05-05
fix crontab syntax

Lyfeway:


taylorotwell:
I think this looks confusing.

Lyfeway:
but Laravel user copy the line, it will never work out. And it's so hard to find out what is going on. Feasibility is more important than anything else.

PR #3298, opened 2017-05-05
fix Homestead.yaml directory name

bobflay:


browner12:
this is not true anymore. this was changed in a recent release so we no longer have the two separate directories (`~/.homestead` and `~/Homestead`) but instead we have one directory, which I believe can technically be called anything you want, but convention is `~/Homestead`.

PR #3304, opened 2017-05-09
Fix a little grammatical error.

amitmerchant1990:


taylorotwell:
Original is correct. Middleware is plural.

PR #3306, opened 2017-05-09
[5.5] Update HasManyThrough Documentation

phroggyy:
This updates the documentation for `HasManyThrough` relations to clarify how the different keys are used, and to document the addition of https://github.com/laravel/framework/pull/19114.

I'm not quite happy with the (in my opinion) too technical wording of the explaining paragraph, so I will gladly accept proposals for change.

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -368,18 +368,20 @@ Now that we have examined the table structure for the relationship, let's define
 
 The first argument passed to the `hasManyThrough` method is the name of the final model we wish to access, while the second argument is the name of the intermediate model.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Wait, so the last argument is for the intermediate table? That's kind of annoying because now the argument order goes... (model, model, intermediate, final, final, intermediate)... it should be model, model, intermediate, final, intermediate, final.

phroggyy:
Yes. We can revert the previous one, and I'll re-PR, or I can submit a patch if you want that changed. The reason for it was to minimise breaking changes. Right now, the only BC break is in the `HasManyThrough` class. Also, if I'm not mistaken, the way it currently goes is `Model, Model, intermediate, final, local, intermediate`. `localKey` does not refer to the final if memory serves me right

taylorotwell:
I think it's OK.

PR #3312, opened 2017-05-12
add suggestion for package table prefixes

browner12:
per conversation on larachat packages channel.

while by no means do we want to *force* the prefixing of table names in packages, it might be a good idea to *suggest* developers use them, to help prevent name collisions in the user's database.


taylorotwell:
If a package is not doing this raise it as an issue on their GitHub. Thanks.

browner12:
was hoping we could add a notes in the docs to *encourage* developers to do this.

PR #3329, opened 2017-05-15
Minor fix 5.0

driade:


driade:
I'd also propose to change "high,low" to "high,default", as it's much clearer that it's needed to state the default queue if you want the worker to pop the jobs. Seen on 5.3 branch.

PR #3339, opened 2017-05-22
Add an example for naming resource routes

trajchevska:
I had a hard time finding how to name routes when working with resources, so I think adding this piece of code in the docs would be valuable for others.


lagbox:
It already exists.

https://laravel.com/docs/5.4/controllers#restful-naming-resource-routes

taylorotwell:
Looks like we have this in another section as noted. Thanks!

PR #3341, opened 2017-05-23
Update dusk.md

ciptohadi-web-id:
to make attribute method usage clearer


taylorotwell:
I think the current wording is clearer.

PR #3350, opened 2017-05-28
remove unnecessary `php` command

hiroqn:
`vendor/bin/homestead` is a sh file
so php command is unnecessary, and php just print shell code

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -169,7 +169,7 @@ Once Homestead has been installed, use the `make` command to generate the `Vagra
 
 Mac / Linux:
 
-    php vendor/bin/homestead make
+    vendor/bin/homestead make
 
 Windows:
 


taylorotwell:
It is required for me.

hiroqn:
Really?

`vendor/bin/homestead` is 

```
#!/usr/bin/env sh

dir=$(d=${0%[/\\]*}; cd "$d"; cd '../laravel/homestead/bin' && pwd)

# See if we are running in Cygwin by checking for cygpath program
if command -v 'cygpath' >/dev/null 2>&1; then
	# Cygwin paths start with /cygdrive/ which will break windows PHP,
	# so we need to translate the dir path to windows format. However
	# we could be using cygwin PHP which does not require this, so we
	# test if the path to PHP starts with /cygdrive/ rather than /usr/bin
	if [[ $(which php) == /cygdrive/* ]]; then
		dir=$(cygpath -m "$dir");
	fi
fi

dir=$(echo $dir | sed 's/ /\ /g')
"${dir}/homestead" "$@"
```

Your `php` command  can execute this ? really?

i cannot execute this on Homestead php

`php -v`

```
PHP 7.1.3-3+deb.sury.org~xenial+1 (cli) (built: Mar 25 2017 14:00:03) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.1.3-3+deb.sury.org~xenial+1, Copyright (c) 1999-2017, by Zend Technologies
    with blackfire v1.16.0~linux-x64-non_zts71, https://blackfire.io, by Blackfireio Inc.
```

and, another environment, cannnot too

```
php -v
PHP 5.5.9-1ubuntu4.21 (cli) (built: Feb  9 2017 20:54:58) 
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2014 Zend Technologies
    with Zend OPcache v7.0.3, Copyright (c) 1999-2014, by Zend Technologies
```

PR #3358, opened 2017-05-31
Added an example for validating array-fields

trajchevska:
I think it would be nice to have an example for validating "array" fields together with the general validation examples. I had to look for external blog articles to find how to do this :)


taylorotwell:
I've added an example in the Validating Arrays section.

trajchevska:
How did I miss this!? Great, deleting the branch :)

PR #3361, opened 2017-06-01
Reminder to keep out of git or other vcs

alnutile:
Just a quick reminder to not let these get into your VCS


taylorotwell:
It says they shouldn't be in source control right above.

alnutile:
Duh sorry

PR #3363, opened 2017-06-03
Added makeWith method

saumya04:
Updated 'Service Container Docs' with `makeWith` method


taylorotwell:
This isn't a good example. It should be shown with an associative array I think.

PR #3367, opened 2017-06-04
Added Slim

fwartner:
Slim also works perfectly with valet.


svenluijten:
Remove the space between the `]` and `(`, Markdown is broken now.

fwartner:
Oh! Thanks! Will fix that as As I'm back on my MacBook :)

PR #3370, opened 2017-06-06
Add tip to show how to use the customized email view (components)

pixelpeter:
Based on my issue here: https://github.com/laravel/framework/issues/17970

I only included this as a {tip} because it doesn't seem to affect to many people.

If you want to have a more extensive explanation (with examples) just let me know


taylorotwell:
This should not be an issue.

jpgnz:
@taylorotwell but it is. Just look at the discussion in #17970, which is the top Google result for figuring out what's going on here.

Would be good to get this documented properly?

Kyslik:
Please take a look at this again.

PR #3375, opened 2017-06-10
Missing helper methods in Database Testing

michaelrtm:
Couldn't find `assertDatabaseMissing` anywhere in the docs, and only found it in the source code. Also stumbled across `assertSoftDeleted`


michaelrtm:
https://laravel.com/api/5.4/Illuminate/Foundation/Testing/Concerns/InteractsWithDatabase.html

These methods are also missing from API documentation.

PR #3376, opened 2017-06-10
dd note

decadence:
I've seen many times people were confused about why their `Auth` or cookies don't work.
It was because they use `dd` for debug and it prevents `\Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse` from running.

### DIFF
--- a/helpers.md
+++ b/helpers.md
@@ -898,6 +898,8 @@ The `dd` function dumps the given variables and ends execution of the script:
 
     dd($value1, $value2, $value3, ...);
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't know, this feels so obvious that killing the script will interrupt the flow of cookies I'm not sure I want to document it at this time.

decadence:
Of course it's obvious for you as a Laravel creator. But people who don't know inner flow of Request can easily fall into this trap. Especially newcomers.

m1guelpf:
@taylorotwell I think this should be added. It would save people from a stackoverflow dive, and adding it doesn't hurt anyone.

PR #3380, opened 2017-06-13
Add fluent routing to routing.md

connectkushal:
Referred these

- https://github.com/laravel/framework/blob/5.4/src/Illuminate/Routing/RouteRegistrar.php#L39
- https://laracasts.com/series/whats-new-in-laravel-5-4/episodes/5


taylorotwell:
We should just change the example to use this style then we don't really need a dedicated section for it.

connectkushal:
@taylorotwell added it as a tip instead and changed the examples #3382

PR #3381, opened 2017-06-13
Add reference to nullable(false) column modifier

pejalo:
I had to leave the docs in search for how to do this, and found the answer here:
https://stackoverflow.com/questions/14013832/make-column-not-nullable-in-a-laravel-migration


taylorotwell:
Columns are not nullable by default?

pejalo:
@taylorotwell Isn't this needed to reverse a migration that makes an already existing column nullable?

PR #3382, opened 2017-06-14
Create routing.md

connectkushal:
Add fluent routing to docs


vpratfr:
@taylorotwell I did not see the `uses` fonction anywhere in the updated docs. Any reason?

E.g.

`Route::get('foo')->uses ('Ctrl@fn');`

VS 

`Route::get('foo', 'Ctrl@fn');`

PR #3383, opened 2017-06-14
Options short names

decadence:
https://github.com/laravel/framework/issues/19594#issuecomment-308420986


taylorotwell:
This is just normal symfony behavior.

PR #3385, opened 2017-06-15
Update homestead release version to v5.4.0

kenhys:


taylorotwell:
We should probably find a way to stop listing the version here and instead list something more evergreen. @svpernova09

svpernova09:
We could add an update command that would check the newest version and update to it. So the docs would read something like:

```
 cd Homestead		  
// Clone the desired release...
./bin/homestead update
```

Issue #3397, opened 2017-06-22
Add manual install of Mix to the docs.


adampatterson:
I would like to see how to manually install Mix in the documentation.

I am coming from an existing project that started with version 5.2, my `package.js` has all of my React and old Babel configs.

I would like to read the docs and know how to manually install Mix.

Thanks!

decadence:
https://github.com/JeffreyWay/laravel-mix/blob/master/docs/installation.md

adampatterson:
Looks like what I was trying to find, And Mix its self doesn't seem to have its own site.

But I don't see that on https://laravel.com/docs/5.4/mix#installation

decadence:
Because Laravel docs doesn't care about non Laravel installation obviously.

adampatterson:
I cant tell if you're being sarcastic because the docs are on the Laravel site, and Mix is included with the install. Weather or not its an official part of Laravel is irrelevant. Someone took the time to document it on the site. As a user of Laravel, and now Mix. 

I am pointing out that there is a gap in the documentation.

decadence:
I'm absolutely serious. Laravel docs says how you can use Mix with Laravel and it doesn't care about using Mix without Laravel.

You can create PR to docs but I think it will be rejected. As I compromise you can suggest adding link to Mix repository at least (if it isn't there already).

adampatterson:
I guess my point then is that after upgrading from 5.2 and moving to 5.4 I can not use Mix with Laravel without finding the install instructions.

It's not 100% clear that Mix is not a core part of Laravel.

You helped me find what I was looking for and thanks for that. But consider how others come down this path.

I will consider adding the PR.

PR #3398, opened 2017-06-22
Add link to initial Laravel Mix repository

decadence:
I know it's easy to find it via Google but some people are confused how to start Mix on non Laravel project.
I suppose you don't want to insert full instuctions in Laravel docs so I propose compromise as link to @JeffreyWay repository.
https://github.com/laravel/docs/issues/3397


adampatterson:
Thanks!

PR #3404, opened 2017-06-26
documentation on default relationship models

browner12:
in reference to https://github.com/laravel/framework/pull/19733

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -120,6 +120,43 @@ If your parent model does not use `id` as its primary key, or you wish to join t
         return $this->belongsTo('App\User', 'foreign_key', 'other_key');
     }
 
+#### Default Models
+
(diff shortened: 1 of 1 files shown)


decadence:
What's about `HasOne`? Can you add note that `withDefault` is available for it? Or it is already somewhere in the docs?

decadence:
Yep, there is no any mention about `withDefault` in docs.

browner12:
@taylorotwell  thoughts on this? I didn't want to duplicate documenting the `withDefault` method under every relation it is available in.

What if I made a top level heading at the very bottom (under 'Touching Parent Timestamps') that could be a more general description of it, and which relations it was available for?  @sileence is working on it for some of the other relations right now, so then we could add them to the list when he's done.

sileence:
It'd make sense to document the feature and mention the relations that support it. By the way here is the PR for morphOne https://github.com/laravel/framework/pull/19788.

taylorotwell:
I don't really mind duplicating a basic example of it under both sections. I can tolerate duplication more in docs than code because it's more of a user assistance thing.

PR #3405, opened 2017-06-27
fix: syntax error in Eager Multiple Relationships usage

awonwon:
it caused show only one table's data which is the first parameter without array type

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -738,7 +738,7 @@ For this operation, only two queries will be executed:
 
 Sometimes you may need to eager load several different relationships in a single operation. To do so, just pass additional arguments to the `with` method:
 
-    $books = App\Book::with('author', 'publisher')->get();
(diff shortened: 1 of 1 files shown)


browner12:
both syntaxes are allowed, since the method uses `func_get_args()` if a string is passed as the first argument.  this pattern is common throughout the framework, where we normalize to an array, but allows parameters passed either way.

taylorotwell:
I'll go ahead and show the array syntax since that is somewhat less magical of a syntax.

awonwon:
I'm sorry. >_<   I found that it is my fault when I called `call_user_func('model:: with', ..)` without checking the quantity of the passed args before. like this: 
```
abstract class Repositroy{
  public function with($tables)
  {
    return call_user_func("{$this->modelClassName}::with", $tables);
  }
}
```

thank browner12 and taylorotwell for your explaining it very much. 😊

PR #3406, opened 2017-06-27
Corrected file location path

ravigehlot:
The 503 template and errors directory is not initially available under resources/views.


taylorotwell:
Fixed this a little differently. Thanks!

PR #3409, opened 2017-06-29
Add Undocumented WaitsForElements Methods

paulredmond:
I went through the `WaitsForElements` trait and tried to document features that haven't been documented yet.

References:

- laravel/dusk#261
- laravel/dusk#176
- https://twitter.com/taylorotwell/status/879090376568602624


paulredmond:
I did actually have a question about the feature from laravel/dusk#176:

I could be missing something.

When using `$browser->waitForLocation('/secret');` it seems like a `$browser->assertLocation('/secret')` would become redundant since the `waitForLocation` is actually going to fail unless it gets to the assertedLocation. Any thoughts on this?

PR #3410, opened 2017-06-30
Add documentation about PSR-11 implementation

deleugpn:
Since Laravel 5.5 will be compliant with PSR-11, it makes sense to specify it in the **container** section as well as to explain the difference of use between `resolve` and `get`.

### DIFF
--- a/container.md
+++ b/container.md
@@ -10,6 +10,7 @@
     - [The Make Method](#the-make-method)
     - [Automatic Injection](#automatic-injection)
 - [Container Events](#container-events)
+- [PSR-11](#psr-11)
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Why do people keep saying this is somehow different than calling App::make()? It's not.

deleugpn:
`App::make(Foo::class)` will automatically return an instance of `Foo` **IF** Foo is a concrete class. `App::get(Foo::class)` will throw a `NotFoundExceptionInterface`

taylorotwell:
OK. I'll rephrase the docs a bit then.

PR #3411, opened 2017-06-30
Added the actual line to add to the `providers` array

mikemike:
The wording of the _Service Provider_ paragraph is such that someone might think they should place `Laravel\Cashier\CashierServiceProvider` as an entry in the `providers` array rather than the full `Laravel\Cashier\CashierServiceProvider::class`.

This is quite important as not including it kills the app without a stacktrace of any kind.  `Laravel.log` does have clues for it, but it's an unusual outcome so I think docs should make it as clear as possible.

### DIFF
--- a/billing.md
+++ b/billing.md
@@ -117,7 +117,9 @@ First, add the Cashier package for Braintree to your dependencies:
 
 #### Service Provider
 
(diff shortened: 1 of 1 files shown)


danjdewhurst:
+1

PR #3412, opened 2017-06-30
Clarify allowed characters in alpha, alpha_num and alpha_dash

TimWolla:


taylorotwell:
I dunno, I feel like most people are pretty clear on what alphanumeric means?

TimWolla:
Personally I expected `[a-zA-Z0-9]` for `alpha_num` from my Western POV.

I suspect more people would be confused why characters like `ᐊ` (CANADIAN SYLLABICS A) would be accepted, because they don't look like letters they are used to.

TimWolla:
I just dug through the commit history to find these two changes to the validator.

1. https://github.com/laravel/framework/commit/7d64a976fe61a5013c77d5c646facf553ccc403f Change from `[a-z]` to `\pL`.
2. https://github.com/laravel/framework/pull/3837 Bugfix, adding the `Mark` category of Unicode.

Thus I feel that the expectations in general are not so clear as they might be to you and kindly ask you to reconsider this PR.

PR #3430, opened 2017-07-10
Fix assigning multiple route middleware

damian-pf:
Docs currently show multiple parameters when you actually need to pass in an array. I have updated the text and code sample to demonstrate this.

### DIFF
--- a/middleware.md
+++ b/middleware.md
@@ -127,11 +127,11 @@ Once the middleware has been defined in the HTTP kernel, you may use the `middle
         //
     })->middleware('auth');
 
-You may also assign multiple middleware to the route:
+You may also assign multiple middleware to the route by passing in an array:
 
     Route::get('/', function () {
         //
-    })->middleware('first', 'second');
+    })->middleware(['first', 'second']);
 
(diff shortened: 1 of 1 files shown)


lagbox:
It doesn't seem to need to be an array. 
https://github.com/laravel/framework/blob/5.4/src/Illuminate/Routing/Route.php#L725

damian-pf:
Hmm that is interesting. I am using the facade in the RouteServiceProvider file.

I have just done a quick test and it seems that using it this way does not call that method you referenced.

lagbox:
You aren't using the example from the docs that you are trying to update. You are doing something else that isn't this example.

damian-pf:
Valid point, where would be an appropriate placi to document this then? I am merely making a minor addition to the given service provider so it is not like this is a hidden feature. I feel that this should be documented, do you disagree?

lagbox:
https://laravel.com/docs/5.4/routing#route-group-middleware

damian-pf:
How do you suggest either linking to or replicating something similar on the Middleware page?

PR #3435, opened 2017-07-12
Tip: Global scope and addSelect

xxRockOnxx:


taylorotwell:
I don't like putting "tip" / "note" callouts back to back. Can this be worked in somewhere else?

PR #3440, opened 2017-07-15
more clarity on CreateFreshApiToken Middleware

msonowal:
As per the documentation it says to define the middleware in the web groups which is applied to all the routes by default.
If the application is using a multi auth guard setup and if the all the routes are defined in the web.php file than this adds the `laravel_token` to all the requests regardless of whatever the guards logged in. Which causes a security flaw by which any authenticated user can issue/access routes of the passport guard applied.
I was having this issue which is a intended behaviour but it was nowhere mentioned in the documentation, So I though it might help for the users who has multiple auth guard setups.
To Fix it It should be said that the middleware should only applied in those routes whose auth provider is same as the api guard provider otherwise it will get cookie even if it is not logged in to that api guard provider

### DIFF
--- a/passport.md
+++ b/passport.md
@@ -628,6 +628,8 @@ Typically, if you want to consume your API from your JavaScript application, you
         // Other middleware...
         \Laravel\Passport\Http\Middleware\CreateFreshApiToken::class,
     ],
+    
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can you give an actual example with code?

msonowal:
will share it asap thanks

taylorotwell:
Please re-open with more detail when you do.

PR #3443, opened 2017-07-16
currentRouteNamed discoverability

AyubM:
Add instructions for a common use case of checking of the path matches a named route.


taylorotwell:
I'm going to document the $request->route()->named() approach.

PR #3445, opened 2017-07-17
inProduction is not a function

tizianozonta:


taylorotwell:
I think you need to update Mix? /cc @JeffreyWay

JeffreyWay:
Yep

PR #3449, opened 2017-07-18
Add some infos about the .env file

lucasmichot:
Add some infos about the .env file


taylorotwell:
We already link to Dotenv earlier. I don't want to re-document all of its features, etc.

PR #3450, opened 2017-07-18
[5.5] Prefer ::class notation

lucasmichot:
Prefer ::class notation


taylorotwell:
Honestly I think it's OK without. Sometimes adding it is longer.

PR #3460, opened 2017-07-24
Update queues.md

billythekid:
Small note on persisting Eloquent models before queuing jobs that refer to them.


taylorotwell:
I think I'm going to hold off on this note for now. I'm not sure it's entirely necessary.

PR #3468, opened 2017-07-26
And doc for @auth and @guest

pierreneter:


taylorotwell:
Syntax is not correct. I will update the documentation myself before release.

PR #3471, opened 2017-07-28
Bootstrap SASS variables further reference link

renjithvk:


taylorotwell:
Hmm, I'm not sure we really need to call that out. It's on the Bootstrap documentation already.

PR #3472, opened 2017-07-28
Revision to docs\Upgrade.md

dspitzle:
This revises the Routing section of the 5.3 upgrade guide to address the move of routes to the routes folder, the split into web.php and api.php, and the associated change in Kernel.php.


taylorotwell:
That is not needed for upgrading.

PR #3481, opened 2017-08-03
Typehinting on the suggested billing migrations.

Alexgmin:


taylorotwell:
It's not really necessary.

PR #3484, opened 2017-08-04
Added information on how to debug commands

phpguru:
Provided a code sample and information on how the `-vvv` option passed to the command on the CLI is picked up by Laravel, and how it can be used to assist with debugging.


taylorotwell:
This is documented when you use "help" on the command.

phpguru:
`php artisan help ___` gives

    -v|vv|vvv, --verbose  Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug

I don't think it's very clear how to use the verbose option in your own commands, and I don't see why you would decline a PR that makes it explicit and improves the documentation by providing a simple example.

PR #3489, opened 2017-08-07
Fixed a typo

DCzajkowski:
I am pretty sure the time it takes to generate a hash **de**creases with the machine power increase ;)


DCzajkowski:
I may be horribly wrong. I don't know what the intent was 😅

DCzajkowski:
Maybe it should be rephrased:
> Bcrypt is a great choice for hashing passwords because its "work factor" is adjustable, which means that the **complexity of the** hash can be increased as hardware power increases.

gmsantos:
@DCzajkowski in this context increased means the algorithm cost to hash a password.

I like the rephrase since time increase could lead to misunderstandings

DCzajkowski:
Yeah, makes total sense!

PR #3493, opened 2017-08-08
[master] Minor fix to dynamic views variables change

paulofreitas:
We must use the `with` prefix to share variables dynamically. :grin:


taylorotwell:
Ah, thanks!

PR #3496, opened 2017-08-08
Fix docs for `Gate::resource` method.

mmghv:
By overriding the abilities array in the third parameter, The keys of this array will be concatenated with a period to the resource name in the first argument.

So keys should be `photo` & `image`, not `posts.photo` & `posts.image`.

### DIFF
--- a/authorization.md
+++ b/authorization.md
@@ -78,8 +78,8 @@ This is identical to manually defining the following Gate definitions:
 By default, the `view`, `create`, `update`, and `delete` abilities will be defined. You can override the default abilities by passing an array as third argument to the `resource` method. The key of the array defines the name of the ability while the value defines the method name:
 
     Gate::resource('posts', 'PostPolicy', [
-        'posts.photo' => 'updatePhoto',
-        'posts.image' => 'updateImage',
+        'photo' => 'updatePhoto',
+        'image' => 'updateImage',
     ]);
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
This was just recently changed to what it is now and I confirmed it.

taylorotwell:
@browner12 weren't you involved in this? Why does this keep getting changed back and forth?

browner12:
I'll take a look this evening.

thanks

browner12:
@taylorotwell What @mmghv is switching it to is the correct way.

Here is the relevant snippet of code:

```php
public function resource($name, $class, array $abilities = null)
{
    $abilities = $abilities ?: [
        'view'   => 'view',
        'create' => 'create',
        'update' => 'update',
        'delete' => 'delete',
    ];

    foreach ($abilities as $ability => $method) {
        $this->define($name.'.'.$ability, $class.'@'.$method);
    }

    return $this;
}
```

So no matter if we use the default `$abilities` or define our own, we **always** prepend the `$name`.

looks like my original PR was correct: 

https://github.com/laravel/docs/commit/c4214e20356f1ea64a67463575510b101ed274a8

then when you made some wording changes, you inadvertently switched it.

https://github.com/laravel/docs/commit/9fa49d7bd1b2549b10e4aa44359ccff778585ddb

PR #3500, opened 2017-08-09
Update pagination to use non-escaping style

stueynet:
I believe this is the correct way. When you use `{{ $users->links() }}` the HTML breaks pretty badly. Changing to `{!! $users->links() !!}` seems to fix it.


taylorotwell:
It is correct how it is I believe. Maybe you're on an older version of Laravel?

stueynet:
Weird. Using L5.4.

PR #3503, opened 2017-08-09
Reword a tip

DCzajkowski:
Changed "which means that the time it takes to generate a hash can be increased" to "which means that the complexity of a hash can be increased", as this is what really happens.


taylorotwell:
But that doesn't describe why it's better over time.

DCzajkowski:
But the time won't really increase, will it? It's just the complexity increase in the same amount of time.

PR #3509, opened 2017-08-11
Options inside clusters key

gmsantos:
Makes sense to set cluster options inside `clusters`

https://github.com/laravel/framework/blob/5.4/src/Illuminate/Redis/RedisManager.php#L99


taylorotwell:
I don't think this is correct.

PR #3510, opened 2017-08-11
[5.5] Add missing `assertSuccessful` method.

zaknes:
I feel that this is, for those who don't look through the main code base, a good method to have documented here.

I put the method first in the list because that is how it is in the `TestResponse` class.


taylorotwell:
I'm not sure that it takes any arguments?

zaknes:
Yeah it doesn't. When I changed the branches, it probably messed something up there. I'll fix it.

PR #3516, opened 2017-08-14
Add where clauses

GDdark:
where($column, $operator, $value, $boolean)
The $value can also be a closure.
The example added.


taylorotwell:
I don't think this is what I intended to demonstrate on this section though. Maybe there is a better place for this example?

PR #3517, opened 2017-08-14
Add leading slashes to class references

simonhamp:
To help new users who don't know why their code isn't working (see https://laracasts.com/discuss/channels/testing/unable-to-locate-factory-with-name-default)


taylorotwell:
I woudl rather them just understand importing, etc.

taylorotwell:
I just don't like the way this looks all over the code.

simonhamp:
Agreed, however this code is still wrong. Typically you would `use App\User;` at the top and then write `User::class`, not `App\User::class` as you'll still get a class not found.

PR #3518, opened 2017-08-14
Add example of using a macro to Collections

simonhamp:
Noticed that there's only [one reference to macros](https://laravel.com/docs/master/responses#response-macros) in the current docs. Hope this example shows the flexibility in the context of Collections.


taylorotwell:
I wonder if it makes more sense to just have a "macros" page under General Topics where we talk about them and how to use them.

simonhamp:
Yep definitely. Will put something together

PR #3525, opened 2017-08-17
Add Method Route::view on Available Router Methods Section

iqmal852:


taylorotwell:
This section is only for HTTP verbs.

PR #3533, opened 2017-08-18
Amended asset compilation 'copy' function to no longer suggest copyDirectory()

Cookizza:
Just a small change here on something I noticed while setting up Mix.

The docs were recommending the use of the `copyDirectory` function for not flattening a directory using `copy`

The correct recommendation however it to pass `false` as the third parameter of the `copy` function.

Hope this is alright 🍡


taylorotwell:
@JeffreyWay 

I believe we added copyDirectory for this reason.

JeffreyWay:
Yep.

PR #3538, opened 2017-08-20
Fixed wrong ending tag in template example

aik099:


brunogaspar:
Don't think it's wrong, as `@show` is meant to show the content of the section immediately without the need to call `@yield('sidebar')`, but it still allows the section itself to be overriden later if required.

aik099:
Ah. Since "@show" tag wasn't documented I thought it was a mistake.

Probably should document "@show" tag.
-- 
Sincerely,
Alexander Obuhovich
http://about.me/aik099

PR #3547, opened 2017-08-24
Fix @endsection

kylekatarnls:
I do not see `@show` in the directives list, so I presume, this should be `@endsection`, if so this mistake should be fixed in every laravel versions of the doc.

If it's not a mistake, can someone explain what exactly is `@show`?

### DIFF
--- a/blade.md
+++ b/blade.md
@@ -43,7 +43,7 @@ Two of the primary benefits of using Blade are _template inheritance_ and _secti
         <body>
             @section('sidebar')
                 This is the master sidebar.
-            @show
+            @endsection
 
             <div class="container">
                 @yield('content')


taylorotwell:
This comes up up almost every week. It is valid.

kylekatarnls:
Thanks for the quick answer.

If this comes up every week, it clearly means this should be clarified in the documentation. I will try to have some tests, look a bit at the code, then when I will properly understand the difference between those two, I will propose some explantation.

In my opinion:
- `@show` should have at least one dedicated line/paragraph in the documentation to explain what it is.
- As `@show` appears in the very first page of the blade section, you cannot expect prerequisites from the reader. So the explanation should come just after `@show` or `@endsection` appears in exemple code.

Thanks again,

brunogaspar:
@kylekatarnls See if https://github.com/laravel/docs/pull/3538#issuecomment-323580673 helps to clarify this a bit and maybe allows you to submit a PR to the docs :)

PR #3549, opened 2017-08-24
Explanations about @show and @endsection

kylekatarnls:
As discussed #3547 and #3538, `@show` has no explanation yet in the documentation and it's a confusing directive for beginners or developers coming from a template engine where the show is automated (I'm thinking of Pug where a block is automatically shown in a layout and replaced when you extend a layout).

Do not hesitate to fix this if my english is not correct or if I misunderstood the difference.


taylorotwell:
Thanks

kylekatarnls:
You're welcome!

PR #3553, opened 2017-08-25
Updated pluralization section to use Inf instead of *

robchava:


robchava:
Sorry, mistakenly was reading 5.4 docs while using 5.3.

PR #3557, opened 2017-08-25
Updated Supervisor configuration

palpalani:


taylorotwell:
There is no numprocs directive though?

PR #3569, opened 2017-08-28
[5.4] Fix passport middleware import

amazzoccone:


brunogaspar:
You just need to remove the `::class` part from the import, otherwise the example will be broken still.

taylorotwell:
Yeah, only the class needed to be removed. Fixed.

amazzoccone:
Oh, i have overlooked it! 👍

PR #3582, opened 2017-08-30
add misc upgrade notes

browner12:
this is the same note I added in the 5.4 upgrade. it helps users keep their `laravel/laravel` files up to date, and helps cover a couple of the more nuanced upgrades to the framework.


browner12:
also would you mind switching the default docs branch to 5.5?

PR #3588, opened 2017-08-31
heads up about the MessageSent event

catalinux:


taylorotwell:
This would go somewhere like an upgrade guide.

catalinux:
Actually it should be included also in 5.4 docs. That event is available, but is hidden in the changelog

PR #3591, opened 2017-08-31
assertJsonStructure method arguments fixed

pavemaksim:
https://github.com/laravel/framework/blob/5.5/src/Illuminate/Foundation/Testing/TestResponse.php#L352


taylorotwell:
That's not really intended to be used in user-land I don't think. Unless you have a use case?

pavemaksim:
@taylorotwell I find it weird when the list of arguments in docs differ from the source code. I have an option to pass `$responseData` to this function. Why should we hide it in docs? What if I want to swap the response data?

Use case? What if I want to assertJsonStructure for a part of JSON that I got from the response?

PR #3609, opened 2017-09-03
Update routing.md

enamul1:
fixing explicit route binding example code


decadence:
Actually that was right. How Eloquent knows which model to fetch?

enamul1:
I think it's not right. Because it already added in 
```
public function boot()
{
    parent::boot();

    Route::model('user', App\User::class);
}
```

decadence:
Oh, yeah. For explicit it seems right.
But anyway it helps for IDE autocomplete.

PR #3613, opened 2017-09-04
Should use distinct for array validation example

ellefsen:
In the example you are validating agains an array of people's emails, and not the user Eloquent model. If you are trying to make sure that all emails in the request are unique, wouldn't the distinct validation rule be more appropriate?


taylorotwell:
No

taylorotwell:
It probably needs both rules honestly.

PR #3626, opened 2017-09-09
Add [Redirects] Document Index

limxx:


taylorotwell:
Redirects are already documented under responses.

Issue #3526, opened 2017-08-17
Changing .app TLD references


inverse:
Since Google is in the process of [registering the .app TLD][0] should the documentation be changed to not use .app and use one of the RFC 2606 advised TLD's for testing?

See: https://tools.ietf.org/html/rfc2606#section-2

[0]: https://www.registry.google/about/domains.html

jakebathman:
So there's another issue with this change: Homestead uses `.app` as a configuration default, not just in the documentation. So we'd have to revise those defaults for Homestead here: https://github.com/laravel/homestead/blob/7007b45c56a09d2afc8c1019854a2ea26b771ff0/src/Settings/HomesteadSettings.php#L103-L108

For quick reference, here's all of the *.app local TLD references in the docs that would need revising:

**homestead.md:140:**homestead.app
**homestead.md:149:**homestead.app
**homestead.md:153:**http://homestead.app
**homestead.md:181:**http://homestead.app
**homestead.md:181:**homestead.app
**homestead.md:250:**homestead.app
**homestead.md:252:**another.app
**homestead.md:257:**homestead.app
**homestead.md:258:**another.app
**homestead.md:268:**symfony2.app
**homestead.md:280:**homestead.app
**homestead.md:294:**homestead.app
**homestead.md:328:**homestead.app
**homestead.md:328:**homestead.app
**homestead.md:328:**homestead.app
**homestead.md:330:**homestead.app
**homestead.md:334:**homestead.app
**homestead.md:346:**homestead.app
**localization.md:139:**messages.app
**pagination.md:137:**http://laravel.app
**valet.md:81:***.app

inverse:
@jakebathman do you feel this change makes sense though?

jakebathman:
Definitely there should be a change away from .app if it's going to be a valid TLD suffix.

I'm not a _huge_ fan of the RFC 2606 options, though. I think the `.dev` or `.local` may be better defaults (seems like a more well-used convention) but that could be discussed in internals.

In the mean time, you can probably submit a PR for the non-homestead docs.

inverse:
I would advise against using `.dev`. See: https://icannwiki.org/.dev

As for `.local` that could work, however looks like it's deemed more for private networks.

jakebathman:
Our shop's convention is `.local` for our own machines and `.dev` for "stage" environments, but I agree that the former is better than the latter.

jonnybarnes:
[...]

PR #3628, opened 2017-09-09
Change tld to localhost

inverse:
As per discussion in #3526 change `.app` tld to `.localhost`.

Still need to change the default configuration in homestead. See: https://github.com/laravel/homestead/blob/7007b45c56a09d2afc8c1019854a2ea26b771ff0/src/Settings/HomesteadSettings.php#L103-L108

### DIFF
--- a/valet.md
+++ b/valet.md
@@ -79,7 +79,7 @@ Valet will automatically start its daemon each time your machine boots. There is
 
 By default, Valet serves your projects using the `.dev` TLD. If you'd like to use another domain, you can do so using the `valet domain tld-name` command.
 
-For example, if you'd like to use `.app` instead of `.dev`, run `valet domain app` and Valet will start serving your projects at `*.app` automatically.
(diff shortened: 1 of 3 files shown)


valet.md
```diff
@@ -79,7 +79,7 @@ Valet will automatically start its daemon each time your machine boots. There is
 
 By default, Valet serves your projects using the `.dev` TLD. If you'd like to use another domain, you can do so using the `valet domain tld-name` command.
 
-For example, if you'd like to use `.app` instead of `.dev`, run `valet domain app` and Valet will start serving your projects at `*.app` automatically.
+For example, if you'd like to use `.localhost` instead of `.dev`, run `valet domain app` and Valet will start serving your projects at `*.localhost` automatically.
```
brunogaspar: Missed one here `valet domain app`.

But i think the example should be the other way around, but that's something for other topic.

inverse: Thanks :+1:

taylorotwell:
I don't really like `localhost`. Would prefer `dev` or something.

drbyte:
I don't like `.localhost` either, but am warming up to `.test` which is safe.

PR #3633, opened 2017-09-12
Added Homestead aliases documentation

poisa:


taylorotwell:
I'm not sure this is very clear. Edit which file? Where is it located, etc.?

PR #3638, opened 2017-09-14
Doc specification of columns in with method

gabriel-caruso:
Recently I saw [a tweet](https://twitter.com/stidges/status/908076318964817923) that shows how we can [specify columns](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Database/Eloquent/Builder.php#L1075) directly when eager loading a relation:
djoif5awaae_eyo
This PR documents this.

PS: This is my first PR for docs, so sorry about any mistakes! Also, this feature is available since [Laravel 5.3](https://github.com/laravel/framework/blob/5.3/src/Illuminate/Database/Eloquent/Builder.php#L1130), so should we document it for those versions?

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -813,6 +813,16 @@ In this example, Eloquent will only eager load posts where the post's `title` co
         $query->orderBy('created_at', 'desc');
     }])->get();
 
+You may use [select](/docs/{{version}}/queries#selects) method to specify a custom select clause for the query:
+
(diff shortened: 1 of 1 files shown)


sausin:
I would think that the docs need to specify that if this is done with a child, the parent id needs to be specified.

i.e.
```php
Author::with('comments:id,text')->get()
```

will not work but
```php
Author::with('comments:id,author_id,text')->get()
```

will work (checked in 5.5).

Was PR'd a while back #3179

PR #3660, opened 2017-09-21
in crontab should be 5 asterisks

feduska:


feduska:
in crontab should be 5 asterisks

decadence:
No. Laravel extends it, 6th is for year.

taylorotwell:
Laravel doesn't extend, some Cron systems just support it.

PR #3663, opened 2017-09-22
fix After Hooks word

baooab:
I think After **Hooks** To Form Requests & After Validation **Hook** To Validators should have the same behavior, not one "before", another "after"。


taylorotwell:
Original is correct.

PR #3664, opened 2017-09-22
Remove console check for package commands

NFarrington:
In the documentation for 5.3-5.5 the example only registers commands in a package if `runningInConsole()`. This would prevent a developer from being able to programatically execute the command from a web (non-console) request, and therefore needlessly removes some artisan functionality from the command.

The 'default' should be to always register the command so as to not remove any functionality and confuse the situation.

### DIFF
--- a/packages.md
+++ b/packages.md
@@ -247,12 +247,10 @@ To register your package's Artisan commands with Laravel, you may use the `comma
      */
     public function boot()
     {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Eh, I think it's fairly typical to only register in the console. It's definitely more efficient the way it is documented.

PR #3673, opened 2017-09-24
Update Homestead to latest version

fairbraxton:
Really not absolutely needed, but just for convenience sake as most people follow the documentation rather than always looking for the latest version.


taylorotwell:
It's silly to keep updating this for every version number. @svpernova09

svpernova09:
Agreed. Did we ever come up with a better idea?

svpernova09:
The Homestead repo could adopt more of formal Git Flow process where master should be stable. PRs are shared against a develop branch.

PR #3676, opened 2017-09-25
Add documentation on real-time facades

jakebathman:
This PR adds a small section into *Facades* that shows real-time facade functionality.

Please let me know if you have any questions. Thanks!


taylorotwell:
Interesting didn't realize these weren't documented.

bloggervista:
It would be great if it shows why would anyone need realtime facades...and its function.

jakebathman:
I tried to show a small example without making it too large, but there are more detailed examples dealing with dependency injection during instantiation that can also be included.

PR #3678, opened 2017-09-26
remove --no-bin-links description

m3esma:
If we use this in windows, when executing the `npm run dev` command we get `cross-env is not recognized as an internal or external command` error.
`npm install` alone is enough and works well.


m3esma:
`npm install` alone is enough and works well.

bloggervista:
yes i guess it was for older version for npm. These days npm works fine on windows.

PR #3684, opened 2017-09-28
To Prevent from Nginx White Screen

bloggervista:
or use ` fastcgi_param  SCRIPT_FILENAME    $document_root$fastcgi_script_name;` to fix to issue
 
Way to reproduce such issue.
* Add nginx deployment script in Amazon Ec2

### DIFF
--- a/deployment.md
+++ b/deployment.md
@@ -45,10 +45,8 @@ If you are deploying your application to a server that is running Nginx, you may
         error_page 404 /index.php;
 
         location ~ \.php$ {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Each system can be a bit different in this regard. We would need more information on if this is also the case on EVERY cloud provider.

bloggervista:
Yes checked many docs it was included or fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name; was used

PR #3686, opened 2017-09-30
Added comment about Windows installation

arasmit2453:
Horizon can't be installed on Windows without the composer option to ignore platform requests, and its ok to ignore them since horizon will be used in production environments running Ubuntu or any other Linux distribution


taylorotwell:
No plans to add this as it may not work properly.

PR #3687, opened 2017-10-01
Removed PHP 5.6 as an option. PHP 5.6 support was removed in Laravel 5.5

PeteH32:
PHP 5.6 support was removed in Laravel 5.5
See https://laravel-news.com/laravel-5-5-php-7-0

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -364,12 +364,12 @@ After running the command, you will see an Ngrok screen appear which contains th
 
 > {note} This feature is only compatible with Nginx.
 
-Homestead 6 introduced support for multiple versions of PHP on the same virtual machine. You may specify which version of PHP to use for a given site within your `Homestead.yaml` file. The available PHP versions are: "5.6", "7.0", and "7.1":
+Homestead 6 introduced support for multiple versions of PHP on the same virtual machine. You may specify which version of PHP to use for a given site within your `Homestead.yaml` file. The available PHP versions are: "7.0", and "7.1" (support for PHP 5.6 was removed in Laravel 5.5):
 
     sites:
         - map: homestead.app
           to: /home/vagrant/Code/Laravel/public
-          php: "5.6"
+          php: "7.0"
 
 In addition, you may use any of the supported PHP versions via the CLI:
 


PeteH32:
This week I downloaded Homestead following instructions here:

- https://laravel.com/docs/5.5/homestead

Running with default settings worked fine. But when I modified Homestead.yaml to set site PHP version to "5.6", the site no longer loaded.  Laravel framework dropped support for PHP 5.6.

- https://laravel-news.com/laravel-5-5-php-7-0

michaeldyrynda:
There might still be people with old projects on 5.6 not yet upgraded to later versions of PHP, which is why the multi-version support was introduced. Makes it easier for those using a single VM for multiple projects across different framework and PHP versions

PeteH32:
Hi @michaeldyrynda .  Thanks for taking a look. I was not sure best way to address this.

I wanted to double-check that you are aware the current stable versions of Homestead and Laravel fail with syntax errors in the source code for laravel/framework if the user tries to use PHP 5.6?   (The Laravel source code now uses PHP syntax introduced in PHP 7.x and will not run when setting 'php: "5.6"' in Homestead.)

Perhaps this documentation section ("Multiple PHP Versions") should include note that it will fail if you set PHP to 5.6, with reference to how to run older versions of Laravel. Anyone trying to run PHP 5.6 would have to downgrade to Laravel 5.5 (I'm not sure how to do that in Homestead).

browner12:
Why would you set a Laravel 5.5+ project to use PHP 5.6? 

Homestead creates separate Nginx configuration files for every mapped site, and specifies which PHP version to use for each.  If your project is Laravel 5.5+, use the default. If you project is 5.4 or less, you can switch to PHP 5.6 if you'd like.

Each project can use it's own defined version of PHP.

PeteH32:
So to run with PHP 5.6, you have to know how to tell either Homestead or "laravel new" command to use an older version of Laravel (must be Laravel 5.4 or less).  I cannot find anything in the docs on how to do that. I checked docs for Homestead, Laravel, and Google.  Sorry, but I am new to Laravel (first week!).

browner12:
Homestead has 3 version of PHP installed. 5.6, 7.0, and 7.1.  You can use **all** of them on a per site basis. If you're creating a new project just use Laravel 5.5, and the default PHP 7

PeteH32:
Our projects must use PHP 5.6. I am new to the project so I am creating everything from scratch with current version of Homestead. 

It sounds like current version of Homestead cannot be used with PHP 5.6 anymore?

browner12:
Yes it can. It says it right in the docs. https://laravel.com/docs/5.5/homestead#multiple-php-versions

So add your project to your code folder and specify the PHP version.

PeteH32:
That is what I tried. It fails to run. Runs fine if you not not set "php" version, but if you set it to "5.6" it will fail.  (It used to work fine, until Laravel 5.5 was released. Now it fails if you set PHP to "5.6".)

If you follow those instructions, your site will fail to load in web browser with this error:

```
( ! ) Parse error: syntax error, unexpected '?' in 
/home/vagrant/site01/vendor/laravel/framework/src/Illuminate/Foundation/helpers.php on line 233
Call Stack
#	Time	Memory	Function	Location
1	0.0008	238272	{main}( )	.../index.php:0
2	0.0034	238704	require( '/home/vagrant/site01/vendor/autoload.php' )	.../index.php:24
3	0.0043	240912	ComposerAutoloaderInita1a24e0307fdeec144ac12692272eb2a::getLoader( )	.../autoload.php:7
4	0.0566	786672	composerRequirea1a24e0307fdeec144ac12692272eb2a( )	.../autoload_real.php:56

```

The source code for laravel/framework now uses PHP 7 syntax, and will fail when you set Homestead to use PHP 5.6.  Here is the git commit that changed the source code for laravel/framework:

- https://github.com/laravel/framework/commit/5f4cd45d8137f1cadb9e0e4ac7270a3b637f5781

browner12:
you are trying to run a Laravel 5.5 project on PHP 5.6.  This **will not work**.  Install an older version of Laravel if you want to run PHP 5.6.

browner12:
please move this over to Slack or StackOverflow if you still have questions.

PeteH32:
Good idea. And I found an answer on StackOverflow before I posted the question!

- [stackoverflow.com: Install specific version using laravel installer](https://stackoverflow.com/a/35163627)

The "laravel new" command has no options for getting older versions of Laravel. So you must use the "composer create-project" command which has an option to specify older versions:

- composer create-project laravel/laravel=5.4.* site02

That is the only way I could get the current version of Homestead to run a site with PHP 5.6.

PR #3700, opened 2017-10-01
[5.5] Document Collection::macro() method

paulofreitas:
Documentation for the `Collection::macro()` method introduced in laravel/framework#11019.

People often use the documentation as their main source of information and, because of that, they often doesn't make use of features that are not documented. This was undocumented since Laravel 5.1.25 and I think that it would improve the Laravel documentation even more.

Because creating and extending collections are a major topic I thought it was relevant to create their own section. I also included the `macro()` method in the available methods section so it would be complete and easier to memorize all of them.

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -2,6 +2,7 @@
 
 - [Introduction](#introduction)
     - [Creating Collections](#creating-collections)
+    - [Extending Collections](#extending-collections)
 - [Available Methods](#available-methods)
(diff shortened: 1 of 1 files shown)


taylorotwell:
Conflicts. I'm going to close all of these PRs because you're going to get conflicts. Send it all as a *single* PR.

PR #3707, opened 2017-10-01
Note on using PHP 5.6 with current version of Homestead

PeteH32:
With current version of Homestead, if want to create a site that uses PHP 5.6, you cannot create new projects using "laravel new" command. Simply use "composer create-project" command instead.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -371,6 +371,10 @@ Homestead 6 introduced support for multiple versions of PHP on the same virtual 
           to: /home/vagrant/Code/Laravel/public
           php: "5.6"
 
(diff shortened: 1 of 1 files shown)


browner12:
this does not belong here. it is irrelevant to homestead.

you are looking at this problem the wrong way. You don't use a version of Laravel to run a version of PHP. It is the opposite. You pick a version of PHP based on your version of Laravel.

PeteH32:
I see what you're describing. Makes sense.  This was my first time trying Laravel, so I appreciate you explaining. Thanks!

PR #3710, opened 2017-10-02
[5.5] add note about mapping every project to a folder

browner12:
mapping every project to a folder is best practice for performance reasons.

@svpernova09 this is in relation to our Slack discussion. does this cover this gist of what we were talking about?


svpernova09:
Yeah, this is good. I think this gets our point across. 

@taylorotwell This is meant to dissuade people from mounting their entire code folders into Vagrant.

PR #3721, opened 2017-10-04
Updating the Billing Documentation for the new swap parameter

MichaelHoughton:
This PR takes adds documentation into the Billing section for the newly introduced second parameter for the swap() method.

Live PR Documentation: https://github.com/MichaelHoughton/docs/blob/feature/billing-docs-updated-for-swap-method-change/billing.md

Revert to: https://github.com/laravel/cashier/pull/446 for the specific change.


taylorotwell:
It's unclear why you would actually do this in many cases.

PR #3724, opened 2017-10-04
Added a note about retrieving cookies via a facade

DCzajkowski:
Good to know, you don't need to use a `Request` object, to retrieve cookies.


taylorotwell:
Combine your PRs for this into one.

PR #3725, opened 2017-10-05
Added Job's display name

vertisan:
Very useful, when we have a many jobs and we need to see more user-friendly title


taylorotwell:
This isn't really intended to be customized.

PR #3738, opened 2017-10-08
Added section about object_get

TobiasFroberg:
Added some info about `object_get`


taylorotwell:
Really should just always use data_get, IMO.

PR #3742, opened 2017-10-09
Update filesystem.md

ragingdave:
Add mention that laravel specific plugins outside of officially supported (in framework) plugins are available in the flysystem docs.
Depends on https://github.com/thephpleague/flysystem/pull/849


taylorotwell:
Will hold off on this for now.

PR #3743, opened 2017-10-09
📚 Updating docs for Homestead v6.3.0

svpernova09:
* Update checkout version (ugh)
* Update `.app` TLD to new default `.localhost`
* Mention port 4040 for ngrok UI


taylorotwell:
Doesn't Chrome not recognize `localhost` as a valid TLD?

svpernova09:
Current versions of Chrome, Safari, and Firefox on MacOS all support `.localhost`.

PR #3748, opened 2017-10-11
Remove one cron asterisk (*)

a3020:
A cron expression consists of max 5 fields, see https://en.wikipedia.org/wiki/Cron


michaeldyrynda:
Also see https://en.wikipedia.org/wiki/Cron#CRON_expression.

https://github.com/laravel/docs/pulls?utf8=✓&q=is%3Apr%20cron

PR #3749, opened 2017-10-11
mix.inProduction() is not a function

reindeer77:
`npm run production fails` 
with
`TypeError: mix.inProduction is not a function`


taylorotwell:
@JeffreyWay ?

taylorotwell:
Your Mix is out of date.

PR #3752, opened 2017-10-12
https instead of http for links

ahinkle:


ahinkle:
Looks like someone beat me before I noticed it!

PR #3771, opened 2017-10-17
Update homestead default TLD to *.test

DojoGeekRA:
This was updated in `v6.5.0` because of https://github.com/laravel/homestead/issues/695


taylorotwell:
.test is a weird name for a local application IMO.

jakebathman:
I really dislike `.test` as well, but I don't know if there's a way around it since `.app` is taken and `.localhost` has these issues on Chrome 😒 

We had a relevant discussion on this [here](https://github.com/laravel/docs/issues/3526) last month, and I just wish `.dev` would work forever.

DojoGeekRA:
@taylorotwell this has already taken place in the homestead repository and has been tagged as released, which means that anyone fetching the latest release will get confused by the generators creating a different TLD than the documentation.

I agree that maybe a further discussion must take place, but in the meantime, this is what homestead is working with.

Issue #2869, opened 2016-11-29
Document $keyType parameter on model


kevinburke:
I am using a UUID as a primary key in an Eloquent model, but getting funky results. I filed https://github.com/laravel/framework/issues/16576, and was told I should be setting `$keyType = 'string'` in my Eloquent model.

Unfortunately this parameter is undocumented - it's not present on the https://laravel.com/docs/5.3/eloquent page.

RRorg:
Bumping this up. Only found the `$keyType` attribute by chance.

decadence:
Do a PR, nobody cares about issues on this repo.

RRorg:
oh well, will try to get around this some time soon ¯\_(ツ)_/¯

laurencei:
Closing as this issue has been fixed.

PR #3777, opened 2017-10-18
[5.5] Complete documentation on how to use non-numerical keys

RRorg:
The usage of non-numerical keys is half-explained in the documentation, but the critical property `$keyType` responsible for type casting is left unexplained, causing confusion. 

see issue #2869


taylorotwell:
Fixed.

PR #3778, opened 2017-10-18
[5.5] Correct alphabetical order issue in available column types

michaeljberry:
Ensuring alphabetical order of column types

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -246,11 +246,11 @@ Command  | Description
 `$table->text('description');`  |  TEXT equivalent for the database.
 `$table->time('sunrise');`  |  TIME equivalent for the database.
 `$table->timeTz('sunrise');`  |  TIME (with timezone) equivalent for the database.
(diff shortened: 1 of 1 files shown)


DojoGeekRA:
@michaeljberry then `$table->timeTz(...)` should be placed after `$table->timestampsTz(...)` too, right?

michaeljberry:
@DojoGeekRA Good question! I thought about that too, but in context I thought leaving the following close to their counterparts made the most sense:

`$table->time()` => `$table->timeTz()`
`$table->timestamp()` => `$table->timestampTz()`
`$table->timestamps()` => `$table->timestampsTz()`

I haven't actually used these column types yet, but I was trying to imagine if I did need to use them and thought:

"Which order would be easiest to help me find them?" 

If these weren't together, how simple would it be to find the Timezone equivalents? That was the thinking I had. Granted they are all close together for now it wouldn't be as big of an issue...

That's why I only addressed `$table->tinyInteger()`, because that was clearly out of place.

PR #3785, opened 2017-10-19
document the $with property

james2doyle:
Document the `$with` property for always eager loading relationships on a model


taylorotwell:
I'm not really a fan of documenting this as a feature. Firstly because, the behavior is not necessarily intentional, it just happens to work sort of by accident. It's not something I set out to support. Secondly, I'm not sure it's even a really good idea to do this rather than requesting explicitly what you need for a given query when you actually do the query.

james2doyle:
Ok I understand. I haven't actually used it this way myself but I could see
a use case for some simple apps with simple APIs.

PR #3789, opened 2017-10-21
Explain autonetwork feature.

ryankilf:
The auto-network feature was added yesterday. This explains how to use it.


taylorotwell:
Bad formatting. Never describes what or where the "ip setting" is, etc.

PR #3793, opened 2017-10-22
Currently it isn't clear that you can pass a boolean to the nullable modifier to either add or remove the ability to insert NULL into that column.

jrgilman:
I added an example of how you can remove a nullable right under the existing example of how to change an existing column to nullable. Additionally, I updated the modifier table's description of the nullable modifier function to describe the above functionality.


taylorotwell:
Just show the argument with the optional value. I don't think you need this long explanation.

PR #3798, opened 2017-10-23
Note about route() helper method usage with Sub-Domain Routing

intrepidws:


taylorotwell:
I'm not sure this is really needed?

PR #3807, opened 2017-10-25
Fix mistake in Redis queue config

luizguilhermefr:
Seems like occurred a mistake on `redis` configuration for queues. The correct file for configuration is `config/queue.php` and it is not a database connection.


gmsantos:
I don't think so. Still need to make a redis connection configuration on config/database.php and then use it on config/queue.php

PR #3809, opened 2017-10-25
Update mix versioning

mmghv:
Update mix versioning docs according to `laravel-mix` `v1.0.0` which now uses query-string versioning rather than hashed filenames.

### DIFF
--- a/mix.md
+++ b/mix.md
@@ -292,12 +292,12 @@ When copying a directory, the `copy` method will flatten the directory's structu
 
 Many developers suffix their compiled assets with a timestamp or unique token to force browsers to load the fresh assets instead of serving stale copies of the code. Mix can handle this for you using the `version` method.
 
-The `version` method will automatically append a unique hash to the filenames of all compiled files, allowing for more convenient cache busting:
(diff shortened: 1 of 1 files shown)


taylorotwell:
@JeffreyWay ?

taylorotwell:
I don't see how the current wording is really invalid and both approaches are supported in next release.

mmghv:
The current docs is misleading :
> The version method will automatically append a unique hash to the filenames of all compiled files

It **doesn't** append the hash to the filename of the compiled files anymore according to `laravel-mix` [v1.0.0 release notes](https://github.com/JeffreyWay/laravel-mix/releases/tag/v1.0.0) :

>File versioning has been simplified. When using mix.version(), no longer will we version the file name, itself. Instead, we'll generate an md5 of the file's contents, and apply it as a query string to the generated mix-manifest.json file. If using a service like Cloudflare, please ensure that you've enabled querystring-based cache-busting.

mmghv:
I was following the docs and got confused because assets files didn't have a hash in their names .. I though something wrong with my project and I kept trying and coming back to the docs to see if there's something missing .. I searched the web and found and issue pointing to the `laravel-mix` release notes.

PR #3812, opened 2017-10-25
request()->routeIs() update

joshralph93:
Updated docs to give example of using wildcard parameters in request()->routeIs(...)


taylorotwell:
Changed code example without surrounding wording.

PR #3816, opened 2017-10-26
Added client scopes info to passport docs

spamoom:
Providing scopes to the client_credentials middleware was not documented - added a code snippet as an example.


taylorotwell:
Why not put this with the rest of the client credential docs?

spamoom:
I considered that - but the first place I looked was the scope section.

PR #3817, opened 2017-10-26
Fix documentation error

hastinbe:
Collection's sort() method uses php's uasort/asort, which maintains index
association, not usort.


taylorotwell:
Please don't change the rest of the sentence.

PR #3818, opened 2017-10-26
Collections use uasort not usort

bkilshaw:
Docs say that collections sort() method uses usort when supplied with a callback but it uses uasort:

https://github.com/laravel/framework/blob/5.5/src/Illuminate/Support/Collection.php#L1347


taylorotwell:
This has a bunch of other commits on it?

PR #3832, opened 2017-10-31
Update notifications.md

travoltron:
Includes info about size limits for sent SMS messages through Nexmo.


taylorotwell:
This can probably be found in the Nexmo documentation.

browner12:
my only concern with a change like this is now we become responsible for keeping it up to date. What if Nexmo changes their limits?  I would suggest, instead, linking to Nexmo's docs.

PR #3834, opened 2017-10-31
Updated notifications.md

travoltron:
Changed to link directly to Nexmo docs regarding character limits.


taylorotwell:
I don't think we need to call this out really. People can look at Nexmo's site if they are curious.

PR #3836, opened 2017-10-31
Add a note about X-CSRF-TOKEN

norvek:
I have found that the the documentation about `Consuming Your API With JavaScript` misses the requirement about the `X-CSRF-TOKEN` HTML meta tag which is required by axios in [bootstrap.js](https://github.com/laravel/laravel/blob/master/resources/assets/js/bootstrap.js#L32) so i added a note about it.


taylorotwell:
I've added wording about this in a slightly different way. Thanks.

PR #3844, opened 2017-11-06
small grammar change - changing is to are in intro paragraph of sessions docs

phegman:


taylorotwell:
Support is. Not Support are.

phegman:
Rereading now, that makes sense. Love working with Laravel, thanks for all the hard work!

PR #3845, opened 2017-11-06
Update in unique validation rule

kupendra:
Line 889 => unique:table,column,except,idColumn changed to unique:table,column,except_id,Column


taylorotwell:
I don't know what "Column" means.

kupendra:
@taylorotwell can you pleas explain what is the meaning of `idColumn` in `unique:table,column,except,idColumn`

Issue #3483, opened 2017-08-04
Minor issue: incorrect path in install docs for Ubuntu


benyanke:
In line 42 of [`installation.md`](https://github.com/laravel/docs/blob/5.4/installation.md#via-laravel-installer) , it mentions that the `$HOME/.composer/vendor/bin` should be added to the user's `$PATH`, so that they can call `laravel new` directly.

However, on Ubuntu 16.04.2 using PHP 7.0.x, that's found at `$HOME/.config/composer/vendor/bin`. Does anyone know the source of the .composer/vendor/bin path?

Assuming it's simply from another OS's default location, such as MacOS, I wonder if the docs could be updated to read something like this?

> Make sure to place composer's system-wide vendor bin directory in your $PATH so the laravel executable can be located by your system. It can be different on each system, but some common locations include:
>  * MacOS: `$HOME/.composer/vendor/bin`
>  * Ubuntu: `$HOME/.config/composer/vendor/bin`

In addition, perhaps we could also mention this command, which would return it (sans `$HOME`, of course, but it would be close). Perhaps by mentioning something like this at the end of the section:

>  If you're having trouble finding this directory, composer can sometimes assist you in finding it: `composer config --list --global | grep "\[home\]" | awk '{print $2}'`

I know this was something that tripped me up when I first started with Laravel, so I'd love to get this a bit clearer and easier.

pwaring:
This bug caught me out too. It's not unique to Ubuntu, it also affects other Linux distributions - I've confirmed this on the current stable versions of Debian and Fedora. However, Gentoo uses the same path as macOS.

benyanke:
Above: submitted 5.4 and 5.5 branch PRs fixing the issue.

laurencei:
Closing as this issue appears resolved now.

PR #3848, opened 2017-11-06
Wrong Composer Path Fixed

benyanke:
Mirror to PR #3847 for the most recent branch

Fixes issue #3483


benyanke:
FYI @taylorotwell 

As per @pwaring, not all linux distros use this path, it might be better to add a 'most' to that.

[Source](https://github.com/laravel/docs/issues/3483#issuecomment-341987982)

PR #3849, opened 2017-11-07
Show people that you can check if a section exists

stefanoruth:
A nice little feature, and a example of the use case


taylorotwell:
It's cleaner to just pass a default value to `@yield` for that use case.

PR #3850, opened 2017-11-08
Eloquent cursor

uruloke:
Added more information to what cursor actually does.

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -217,6 +217,8 @@ The `cursor` method allows you to iterate through your database records using a 
         //
     }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not sure this is needed.

uruloke:
@taylorotwell The reason for adding information about how it works is because I have seen people asking what's the difference between chunk and cursor, and the answer which is given is that cursor uses less memory, but they rarely know how it actually work.
As this is just leveraging a SQL function, I think it somehow needs to say this, so we can easily understand when to use it and when not to.

PR #3862, opened 2017-11-12
Array instead of string for multiple route middleware

marksalmon:
Route->middleware() accepts array when assigning multiple middlewares.


taylorotwell:
Both work.

PR #3868, opened 2017-11-15
Added array for when assigning multiple Middleware

JohnnyWalkerDesign:
Seems to have been a typo. It appears you need to wrap multiple Middleware assignments in an array. If you don't, the second assignment is seen as a parameter for the first.


paulofreitas:
It does work. This was added in laravel/framework#21621, available since Laravel 5.5.15. :wink:

PR #3874, opened 2017-11-20
- Scheduling docs Homestead Configuring Cron Schedules linking

sasivarnakumar:
People referring the docs during development might not know that Homestead has an easy option for enabling scheduling. 
They might end up configuring CRON jobs in their Homestead box instead. This commit / doc block helps users know that there is an easy way in Homestead for enabling scheduling.


taylorotwell:
It's already one the page.

PR #3884, opened 2017-11-27
Add note about auto-loading commands

luiz-brandao:
Auto-loading of Artisan commands was not working for me when I upgraded from 5.4 to 5.5. I found that some code was missing from my `App\Console\Kernel::commands`.  This PR adds a note about that in the Upgrade guide.


browner12:
this is why you need to follow the "Miscellaneous" directions at the bottom of the upgrade guides.

PR #3887, opened 2017-11-28
Add note about mass assignment during seeding

luiz-brandao:
Today I learned, after some struggle, that mass assignment protection is disabled during seeding. It might be useful to add a note somewhere in the docs about it.

I'm submitting another PR, adding the same note in the `Eloquent` section, so you can chose the best place to add the note in case you judge it pertinent.

### DIFF
--- a/seeding.md
+++ b/seeding.md
@@ -44,6 +44,8 @@ As an example, let's modify the default `DatabaseSeeder` class and add a databas
         }
     }
 
+> {note} Mass Assignment protection is disabled during seeding.
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
What struggle did you have?

luiz-brandao:
I was trying to reuse some code from a controller to quickly pre-populate the database with existing testing data. The code relied on the mass assignment protection to "filter" the data from the request. I was confused until I found some similar discussion about the subject.

PR #3895, opened 2017-12-04
5.4

CharlieBrownCharacter:
[First pull request here. Be gentle with me]

In the docs about the DB::transaction (see: https://github.com/laravel/docs/blob/5.5/database.md) it says:

```
DB::transaction(function () {
    DB::table('users')->update(['votes' => 1]);

    DB::table('posts')->delete();
});
```

It doesn't specify in the docs that `DB::transaction` may return a value. In my case I'm using it in the `create` method to register a user and I had to:
```
$user = null;
        $user = DB::transaction(function () use ($data, $user) {
            $user = User::create([
                'name'          => ucfirst(strtolower($data['name'])),
                'last_name'     => ucfirst(strtolower($data['last_name'])),
                'email'         => $data['email'],
                'password'      => bcrypt($data['password']),
                'concelho_id'   => $data['concelho'],
            ]);
            // Some other operations related to $user->id
            return $user;
        });

        return $user;
```
If I hadn't assigned it to an outside variable like `$user` it would throw `Type error: Argument 1 passed to Illuminate\Auth\SessionGuard::login() must implement interface Illuminate\Contracts\Auth\Authenticatable, null given`. This is specific to my case but I think it should state on the docs to not so experienced programmers (including me) that `DB::transaction` returns a value


browner12:
it appears your changes are not related to your description at all.

PR #3901, opened 2017-12-06
Nullable type example

enderandpeter:
- PHP 7.1 allows nullable types so that `null` will be accepted for
    parameters with a declared type. We should mention this so that the
    user knows their options.


taylorotwell:
I don't think we need to mention this here. Thanks.

PR #3904, opened 2017-12-07
Update Valet TLD

sagalbot:
Updated references to `.dev` domains to match the new default of `.test` set in laravel/valet#436

Had previously PR'd this (#3903) but pointed against `5.5`. Made more sense to point to master.


taylorotwell:
It actually makes the most sense against 5.5 I think?

PR #3907, opened 2017-12-10
Make it clear which migration to use in Dusk

alnutile:
The docs show examples, which is nice, but it might be easy, as a new dusk user, to assume the faster `use RefreshDatabase;` could work as well.
But maybe by making it clear here we can prevent some time lost to this since it will not work. I finally figured it out thanks to https://laracasts.com/discuss/channels/testing/dusk-problem-databasetransactions


taylorotwell:
I don't like to put callouts back to back. We would need to work it into the documentation flow somehow. Thanks.

alnutile:
Would it makes sense for me to follow up with the solution as another PR @taylorotwell

PR #3908, opened 2017-12-10
update policy behavior

mputkowski:
`before` method (if exist) will be called independently from method and ability name.

Reference: [#21989](https://github.com/laravel/framework/pull/21989)

### DIFF
--- a/authorization.md
+++ b/authorization.md
@@ -229,8 +229,6 @@ For certain users, you may wish to authorize all actions within a given policy. 
 
 If you would like to deny all authorizations for a user you should return `false` from the `before` method. If `null` is returned, the authorization will fall through to the policy method.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
@themsaid ... We keep changing the wording around this. What is the actual correct documentation?

themsaid:
The current documentation is correct, the changes in this PR only concerning this case:

```
Gate::define('update.post', 'PostPolicy@update');
```

Only if you explicitly set a method name, we won't check if method exists or not an will call the `before()` method assuming you know what you're doing since you **explicitly** said `PostPolicy@updatePost`.

BUT if you say `$user->can('updatePost', $post)` and you don't have an `updatePost()` method in your policy we won't call the before method, we'll directly try to find an ability called `updatePost` and ignore the policy completely.

mputkowski:
Should I mention about second case (if ability isn't defined) or it's ok? Currently there is incomplete information and this could be misleading.

PR #3911, opened 2017-12-12
add additional description for "Binding Primitives"

Sergey-Panchenko:


taylorotwell:
We never say otherwise.

PR #3912, opened 2017-12-13
Fixed determinant a to an

bosz:
Instead of of as a sms, i think it is supposed to be as an sms.


browner12:
I would tend to agree.

http://www.quickanddirtytips.com/education/grammar/a-versus-an

bosz:
Thanks @browner12 . Seams the same error spans across all the version of the documentation. 

So i should possibly just do a quick scan and make all changes on the various branches and submit the patches.

PR #3913, opened 2017-12-13
Try #2: Clearly document RefreshDatabase will not work

alnutile:
and you need to use DatabaseMigrations. 
Our team depends on these docs, and my concern is that more people on the team will lose time to this very very hard to track down issue.

### DIFF
--- a/dusk.md
+++ b/dusk.md
@@ -9,6 +9,7 @@
     - [Environment Handling](#environment-handling)
     - [Creating Browsers](#creating-browsers)
     - [Authentication](#authentication)
+    - [Database Migrations](#migrations)
 - [Interacting With Elements](#interacting-with-elements)
     - [Dusk Selectors](#dusk-selectors)
     - [Clicking Links](#clicking-links)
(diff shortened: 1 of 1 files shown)


browner12:
FYI, when writing large blocks of code you can put them in between

\```php

//PHP code in here

\```

so you don't have to start every line with 4 spaces.

In regards to this PR, `RefreshDatabase` and `DatabaseMigrations` are not part of Dusk.  They are part of the core testing facilities provided by the framework.  Therefore, at a minimum you should move these changes to `database-testing.md`.

alnutile:
@browner12 it was an exact copy of the block above?
So not sure why this would matter?

But thanks for the feedback.

browner12:
oops, my bad.  looks like Laravel isn't using that syntax

alnutile:
Maybe in some places? 
My real concern is my team relies on these docs. We rarely diverge from them. So when they are missing details like this that can cost people hours of their time it is something that I get a bit concerned with.

Though to be clear Laravel is still some of the best docs I have used for a framework.

browner12:
the details for `RefreshDatabase` are over on the `database-testing.md` page.  `RefreshDatabase` was created as a replacement for both `DatabaseMigrations` and `DatabaseTransactions`.

can you give an example test that isn't working with `RefreshDatabase`? All this trait does is decide if it should run `artisan migrate` or `artisan migrate:fresh` depending on if you're using an in-memory database or not.

alnutile:
@browner12 here is a pass fail gif
https://www.dropbox.com/s/kmdvo4dgd0rn9uu/passing_and_not.gif?dl=0
Sorry I did not give more context BUT if you look at my last pr https://github.com/laravel/docs/pull/3907#issuecomment-350551479 Taylor was not saying it was not true just did not like the docs.

alnutile:
@taylorotwell thanks!

PR #3918, opened 2017-12-14
Added a few references to Distinct rule, near Unique documentation

dwyer14:
Unique is well-documented with thorough examples, but it's easy to miss Distinct. I feel the two are related and might be used together somewhat frequently. Just added a couple of mentions near Unique documentation for array input.

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -888,7 +888,7 @@ The field under validation must be a valid timezone identifier according to the 
 <a name="rule-unique"></a>
 #### unique:_table_,_column_,_except_,_idColumn_
 
-The field under validation must be unique in a given database table. If the `column` option is not specified, the field name will be used.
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hmm, I'm not sure I see them as related?

dwyer14:
Fair enough, maybe it's more of an edge case. Here's my situation.

**TL;DR** - Anytime your front-end allows for multiple record uploads (see image) that include a unique DB column, there's a good chance you should use both `unique` and `distinct`, to get a proper validation respone.

- Admin can add multiple users at once. 
- Input array validation fields look like: `users.*.email`, `users.*.name`, etc. 
- Email is a required field, and has unique index in the database.
- `Unique` rule checks DB for duplicate email entry.
- If email isn't in DB, but input array has duplicate values, it passes `unique` validation. 
- When the second "instance" of duplicate email is attempted, Laravel returns 500 instead of 422, because the first instance has already been added.

If that's more of an edge case, totally get it. Just thought it might save somebody, who's in the same situation I was, from hunting around too much.

image

PR #3919, opened 2017-12-15
Tip on using composer with a specific php version

redgluten:
Right now running a composer command with a php version without specifying it will lead to dependencies version mismatch, this change adds a little note to prevent such issues.


taylorotwell:
I'm not sure this needs to be stated.

PR #3930, opened 2017-12-22
Added how to undo the cache

stealthAngel:
It happened to me and it was hard to find a way to undo the process. I'll place the fix here so if someone faces the same problem as me,  he can find the solution easily.


Kyslik:
What about using `php artisan config:clear` instead?

stealthAngel:
That didn't work for me. I did it in the pull before but closed it cause it doesn't work.

stealthAngel:
Then can we change it to 

"If you would like to revert the cache, execute the `cache:clear` command"

PR #3933, opened 2017-12-24
Rename resource name from `User` to `UserResource`

danielebarbaro:
Hi, i've just rename resource name from `User` to `UserResource` >> `php artisan make:resource UserResource`.

I've rename it to avoid collision on model name so 
```
use App\Http\Resources\User as UserResource;
```
simply becomes to 
```
use App\Http\Resources\UserResource;
```
Sometimes some IDE misunderstanding which class have to import


danielpclin:
In Laravel 5.5 if you create a collection for example UserCollection, when you `return new UserCollection()` from controller, it will throw file not found error because Laravel is looking for `\app\Resources\User.php` not `\app\Resources\UserResource.php` thus this commit and https://github.com/laravel/docs/pull/3934/commits/0638352c3f5e49df4340605f9184848c5143325e should be reversed in Laravel 5.5 docs

PR #3938, opened 2017-12-28
Explicitly show that the code for order shipping needs to be written

amadeann:
Reading the code example it is not clear that the code for shipping the order needs to be written.

3 out of 4 assertions in the example have a comment right above them, describing what the assertion does.
The comment above the first assertion does not refer to the assertion itself, but instructs the reader to write the logic for shipping the order.

This may be quite misleading, as the reader may assume that `assertSent` method is responsible for performing the order shipping.


taylorotwell:
There are many other examples without this type of comment.

PR #3939, opened 2017-12-28
Document Redirect::away()

guidocella:
I just had a redirect to PayPal 404 only on mobile devices which has hell to debug until I remembered `away` exists so I think it's worth documenting.

### DIFF
--- a/responses.md
+++ b/responses.md
@@ -7,6 +7,7 @@
 - [Redirects](#redirects)
     - [Redirecting To Named Routes](#redirecting-named-routes)
     - [Redirecting To Controller Actions](#redirecting-controller-actions)
+    - [Redirecting To External Domains](#redirecting-external-domains)
     - [Redirecting With Flashed Session Data](#redirecting-with-flashed-session-data)
 - [Other Response Types](#other-response-types)
     - [View Responses](#view-responses)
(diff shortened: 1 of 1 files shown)


taylorotwell:
How is this different than just `return redirect('https://some-website.com')`.

guidocella:
Not sure - I only know the redirect started working on mobile devices (or at least on Chrome for mobile) with `away`. With just `redirect` it redirected to the current URL with [object%20Object] as the last part of the URL (mydomain.com/checkout/[object%20Object]) which is really weird, and it's hard to figure out why since `redirect` works on desktops, even the comparison between the `RedirectResponse`s is `true` so it's really baffling...

According to [https://medium.com/@zwacky/laravel-redirect-to-vs-redirect-away-dd875579951f](https://medium.com/@zwacky/laravel-redirect-to-vs-redirect-away-dd875579951f) `away` won't:
    Determine if the URL is secure
    rawurlencode() the URL
    trim() the URL

The method was introduced in laravel/framework#2111 where it was needed to redirect to a URL with special characters, however in my case I was already calling `http_build_query` (`return redirect('https://www.paypal.com/cgi-bin/webscr?' . http_build_query([ ... ])`) or redirecting to a URL that doesn't need to be encoded (`return redirect("https://www.paypal.com/cgi-bin/webscr?cmd=_express-checkout&useraction=commit&token=EC-4TM64573H55171134")`) so I'm not sure why `away` was needed and whether it is due to a Chrome for mobile or PayPal bug, but I'm definitely using it from now on rather than risking that customers are unable to pay :).

[Here is another issue when not using away.](https://stackoverflow.com/questions/28642753/redirect-to-external-url-with-return-in-laravel/28643019)

guidocella:
I'm sorry, I figured out the problem - I was passing a google conversion function to a submit event (`$('form').submit(goog_report_conversion)`) so the event object became its first parameter, but the first parameter of that function is the URL to redirect to from within a callback.
I guess the error was only on happening Chrome for mobile because the submit handlers or the requests to Google executed in a different order, and maybe the requests to Google got delayed right when I switched to `Redirect::away()` (some customers could pay so it wasn't always happening)...

So it should be fine to just use `redirect` as long you don't omit the protocol and the URL doesn't need to be encoded or trimmed, in case you now want to remove it from the docs.

PR #3940, opened 2017-12-28
Use PHP 5.5 syntax

freekmurze:
Because it's almost 2018 😀

### DIFF
--- a/views.md
+++ b/views.md
@@ -124,7 +124,7 @@ For this example, let's register the view composers within a [service provider](
         {
             // Using class based composers...
             View::composer(
-                'profile', 'App\Http\ViewComposers\ProfileComposer'
+                'profile', App\Http\ViewComposers\ProfileComposer::class
             );
 
             // Using Closure based composers...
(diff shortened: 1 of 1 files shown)


carusogabriel:
Don't let @sisve see it, he hates CS PRs :joy:

views.md
```diff
@@ -124,7 +124,7 @@ For this example, let's register the view composers within a [service provider](
         {
             // Using class based composers...
             View::composer(
-                'profile', 'App\Http\ViewComposers\ProfileComposer'
+                'profile', App\Http\ViewComposers\ProfileComposer::class
```
carusogabriel: We don't need a backslash here, as it is namespaced? :thinking:

views.md
```diff
@@ -198,7 +198,7 @@ You may attach a view composer to multiple views at once by passing an array of
 
     View::composer(
         ['profile', 'dashboard'],
-        'App\Http\ViewComposers\MyViewComposer'
+        App\Http\ViewComposers\MyViewComposer::class
```
carusogabriel: We don't need a backslash here, as it is namespaced? :thinking:

taylorotwell:
I avoid this because it's namespaced. Would require import or backslash.

PR #3942, opened 2017-12-29
[5.3] Illuminate\Auth\Middleware\Authenticate renamed middleware

### LINKED ISSUE #3001: 5.2 to 5.3 does not mention change in routeMiddleware 'auth'
Following the process of migration from 5.2 to 5.3 does not mean that it should be changed in ```Kernel.php```

```php
protected $routeMiddleware = [
     'auth' => \App\Http\Middleware\Authenticate::class,
    ...
];
```
to
```php
protected $routeMiddleware = [
     'auth' => \Illuminate\Auth\Middleware\Authenticate::class,
    ...
 ];
```
Failure to perform the mentioned change results in null when using Passport to obtain the user corresponding to the token

carusogabriel:
This resolves #3001. [I used Github Compare feature to assert that](https://github.com/laravel/laravel/compare/5.2...5.3#diff-b3af398d5ac2670049a1f17f4683c03eR48).

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -379,6 +379,10 @@ Once this route middleware has been registered, you should add it to the `api` m
         'bindings',
     ],
 
+#### Authenticate Middleware
+
+The `App\Http\Middleware\Authenticate` middleware has been renamed to `Illuminate\Auth\Middleware\Authenticate`.
+
 ### Notifications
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
This isn't correct.

PR #3944, opened 2017-12-29
[5.5] Add Middleware Resource Routes

### LINKED ISSUE #3892: Expose all posible option values for resource controller method
I suggest we should expose all posible available keys in the `$options` array for the `Route::resource` method.

So far, in the docs, we can figure those out:
only, except, name and parameters

Is there other supported options that aren't documented? Can I add "middleware" in the options? Can anyone confirm the available options?

[Resources Controllers](https://github.com/laravel/docs/blob/5.5/controllers.md#resource-controllers)

carusogabriel:
This resolves #3892 by exposing the last missing group of options in `Resource Routes`: `Middleware` :wrench:

### DIFF
--- a/controllers.md
+++ b/controllers.md
@@ -8,6 +8,7 @@
 - [Controller Middleware](#controller-middleware)
 - [Resource Controllers](#resource-controllers)
     - [Partial Resource Routes](#restful-partial-resource-routes)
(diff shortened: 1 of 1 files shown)


taylorotwell:
This isn't correct information. Resource routes don't have the web middleware by default. That's only because your routes file is being loaded inside a route group with the web middleware.

PR #3950, opened 2018-01-01
Update collections.md

cPlayIt:
http://php.net/manual/en/function.list.php 
'Like array(), this is not really a function, but a language construct... '


taylorotwell:
You changed other things in this PR.

cPlayIt:
Changed $even and $odd variables because their values are mismatched.

PR #3953, opened 2018-01-04
env() - config:cache behaviour clarification

IlCallo:
[Reference](https://github.com/laravel/framework/issues/21727)

Docs is not exactly clear on which are the consequences of using the `env` helper other than within configuration file when launching `config:cache` command.
`env()` description doesn't even put a notice about this behaviour.

This PR adds some warnings and explanations


sisve:
> because after that command all `env()` calls will always return null

No, that is not at all the behavior. `env()` will still read environment variables, it's the population of environment variables from the .env file that isn't done. All other existing environment variables can technically still be read by `env()`.

IlCallo:
Taylor wrote it better than I did, check his commit after the merge

PR #3955, opened 2018-01-04
Passport: refine frontend quickstart example

LaughingCabbage:
Add a simple test so the user can immediately use the Vue components and verify passport is working without further digging into Vue documentation.


taylorotwell:
I'm not sure this is 100% necessary. Thanks.

PR #3957, opened 2018-01-05
documenting use of nested directory structure for translation files

amadeann:
Laravel localization doesn't follow the usual "dot notation" for files in nested directories. When using the dot notation with nested directories translation key is returned. 

Such edit to docs was suggested already in July 2016 ([#14212](https://github.com/laravel/framework/issues/14212)).

### DIFF
--- a/localization.md
+++ b/localization.md
@@ -106,6 +106,16 @@ Of course if you are using the [Blade templating engine](/docs/{{version}}/blade
 
 If the specified translation string does not exist, the `__` function will simply return the translation string key. So, using the example above, the `__` function would return `messages.welcome` if the translation string does not exist.
 
+If you are using a nested directory structure, you should specify the path to the translation file using forward slashes:
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
This behavior is not intentional and thus may not be supported in the future so would rather not document it.

timo75:
> This behavior is not intentional

You mean localization files should not be placed in nested directories? Or are you referring to slash vs dot notation when linking them?

StefanoTesla:
What in laravel 10?? We can use subfolder for translations or not??

vsjoshtait:
With reference to https://github.com/laravel/ideas/issues/2629, the fact that this behaviour currently works with `/` notation means that a decision should be made either way.

It would be trivial to officially support dot notation for nested folders. Doing so will allow developers to keep translation files smaller and have a closer parity between front-end components and their strings, not to mention greater flexibility.

PR #3962, opened 2018-01-10
Update homestead.md

octoxan:
Added quotes around site type in the example, as it's required to not error out. Added warning about trying to add apache to an existing nginx box.


browner12:
please see the comments in the Homestead issue (https://github.com/laravel/homestead/issues/790).

Apache and Nginx should work together on the same box.  `flip` command is super useful since you can only have one web server running at once.

PR #3975, opened 2018-01-17
Require dotenv in order to use environment variables

susanBuck:
In order to use environment variables in `webpack.mix.js` you have to require `dotenv` (ref: <https://github.com/JeffreyWay/laravel-mix/issues/1155>).

Documentation should reflect this (or `webpack.mix.js` that ships w/ Laravel should just include it?)

### DIFF
--- a/mix.md
+++ b/mix.md
@@ -327,7 +327,12 @@ You may pass either a string (proxy) or object (BrowserSync settings) to this me
 <a name="environment-variables"></a>
 ## Environment Variables
 
-You may inject environment variables into Mix by prefixing a key in your `.env` file with `MIX_`:
+To inject environment variables into Mix first require the `dotenv` package (which is installed by Mix) at the top of `webpack.mix.js`:
+
+    let mix = require('laravel-mix');
+    require('dotenv').config();
+
+Then prefix keys in your `.env` file with `MIX_`:
 
     MIX_SENTRY_DSN_PUBLIC=http://example.com
 


JeffreyWay:
This is a little misleading. You only have to pull in that package if you want to reference env vars directly within the webpack.mix.js file (before we have compiled your code). 

Anywhere else in your bundle, and it'll happen automatically.

susanBuck:
The [Laravel docs on Mix](https://laravel.com/docs/5.5/mix) tells us this:

>> The webpack.mix.js file is your entry point for all asset compilation. Think of it as a light configuration wrapper around Webpack.

Thus, unless told otherwise, it should be safe to assume that any of the example code provided on the docs page that follows that statement is able to be run in `webpack.mix.js`, no? 

I believe every other example code provided does in fact work as written in `webpack.mix.js` - this is the only exception. Shouldn't that be explained?

susanBuck:
@taylorotwell I think that's good solution.

JeffreyWay:
No, it’s ugly. 

I’ll see if we can make it work automatically in the configuration file.

JeffreyWay:
@taylorotwell This can be closed. I just pushed an update to Mix that'll make the config available even in the configuration file.

https://github.com/JeffreyWay/laravel-mix/commit/a7ce542f9cbbf37014283d06dc06658f8cf84b92

@susanBuck Thanks!

PR #3977, opened 2018-01-19
noobs need more hand-holding

zackaryvelo:
This proposed change is just to make it more obvious for new developers how to get Homestead up and running.


browner12:
seems like overkill to me.

if accepted could use some word-smithing.

PR #3981, opened 2018-01-20
Document routing groups name method

marcnewton:
Added missing documentation for route name method prefixes on groups in routing docs.

Tested rule:

```php
Route::namespace('Admin')->prefix('admin')->name('admin.')->group(function () {

    Route::resource('users','UserController');

});
```

Confirmed results:

* `namespace()` sets Controller namespace to \Admin\UserController

* `prefix()` sets request URi to /admin/users

* `name()` sets route name accessor to `route('admin.users.index')`

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -13,6 +13,7 @@
     - [Namespaces](#route-group-namespaces)
     - [Sub-Domain Routing](#route-group-sub-domain-routing)
     - [Route Prefixes](#route-group-prefixes)
+    - [Route Name Prefixes](#route-group-name-prefixes)
 - [Route Model Binding](#route-model-binding)
     - [Implicit Binding](#implicit-binding)
     - [Explicit Binding](#explicit-binding)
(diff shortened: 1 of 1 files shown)


marcnewton:
Reason for this PR is because I spent an entire day looking into solving the issue around changes to the `prefix` back from an update from 5.2 to 5.3 and did not particularly everyone's answer to it to use the array config `['as' => 'admin']`.
All I could find was 2 year old discussions and arguments about the change, that array solution, a lot of annoyed people and not one single mention of using the `name` method on a route group nor an example of it in the docs, hence the PR.

I strongly believe adding this usage case to the docs will help a lot of people.

routing.md
```diff
@@ -13,6 +13,7 @@
     - [Namespaces](#route-group-namespaces)
     - [Sub-Domain Routing](#route-group-sub-domain-routing)
     - [Route Prefixes](#route-group-prefixes)
```
marcnewton: Would it be better to rename **Route Prefixes** to **Route URi Prefixes** for a bit more clarity?

PR #3993, opened 2018-01-26
Update structure.md

davidangel:
Very minor grammar fix.


browner12:
this has always been a tricky one, because some people pronounce it "Sequel Lite" and some people pronounce it "Ess Q L Lite"

PR #4006, opened 2018-01-31
note about `hash` alias change when injecting the hasher

robclancy:
Words are hard.


taylorotwell:
Fixed this so you can still type-hint this and get the default hasher.

PR #4009, opened 2018-02-01
Mistyped php version

Batmandakh:
Fixed the mistyped php version to 7.1.13

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -11,7 +11,7 @@
 
 ### PHP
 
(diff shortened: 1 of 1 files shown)


Batmandakh:
I've confused of php 7.1.3. Because there were no 7.1.3 version. So, I thought, this could be mistyped.

taylorotwell:
http://www.php.net/ChangeLog-7.php#7.1.3

PR #4011, opened 2018-02-02
Clarifying folder examples from homestead.md

dtb:
Trying to make it a little clearer what the examples in `Homestead.yml` mean.


browner12:
hmm.... i'm not sure on this.  Part of me wants to avoid becoming too verbose in our documentation.  However, I realize I have been a part of the project for far too long to realize what it's like coming in as a newer user.

dtb:
Yeah, I don't feel too strongly about this one. It was slightly confusing, but also not that hard to figure out.

PR #4013, opened 2018-02-02
Update authorization.md

sileence:


taylorotwell:
No description on your PR.

sileence:
Sorry, I just found this confusing/redundant:

> When defining policy methods that will not receive a model instance, such as a `create` method, it will not receive a model instance

PR #4014, opened 2018-02-02
Update authorization.md

sileence:
I thought an extra clarification here could be useful for people not that familiar with exceptions?


taylorotwell:
I don't understand what this means.

PR #4018, opened 2018-02-04
Add missing word

enderandpeter:
A word like "create" appears to be missing in the sentence "The `url` helper may be used **to arbitrary** URLs for your application."


lagbox:
or 'generate' ... good catch

enderandpeter:
Yeah, that would be totally fine 😃

PR #4019, opened 2018-02-04
Add essential notes about Hyper-V

enderandpeter:
I think it would be to the documentation's great benefit to add a few more provider specific notes about Hyper-V. There are a handful of tips that are not very obvious to someone who just wants to get it up and running and I feel it is incumbent upon us to add only a little crucial information.

I wholly invite and greatly encourage help in paring this down or otherwise refining these recommendations.

Things would have gone much better for me initially if I had a more typical kind of network connection. I had a wifi adapter with a bridged network connection so things were working a little differently. Here are some of the resources I found that helped guide me towards a setup that finally worked:

- [Vagrant + Hyper-V (Windows 10 anniversary)](https://quotidian-ennui.github.io/blog/2016/08/17/vagrant-windows10-hyperv/)
- [Learning to Use Vagrant on Windows 10](https://blogs.technet.microsoft.com/virtualization/2017/07/06/vagrant-and-hyper-v-tips-and-tricks/)
- [Vagrant Hyper-V Configuration](https://www.vagrantup.com/docs/hyperv/configuration.html)

Thank you for your consideration.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -485,3 +485,31 @@ If symbolic are not working properly on your Windows machine, you may need to ad
     config.vm.provider "virtualbox" do |v|
         v.customize ["setextradata", :id, "VBoxInternal2/SharedFoldersEnableSymlinksCreate/v-root", "1"]
     end
+
+<a name="provider-specific-hyperv"></a>
+### Hyper-V
+
+#### Networking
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't really like merging vagueish information (these options might help, etc) into the documentation. It's not clear why or when those options would actually be needed.

PR #4024, opened 2018-02-05
Add information on generating a private key

TatiannaWS:
Every time I've tried to install Homestead, I get the mentioned error the first time I try to run `vagrant up`. I then have to spend extra time trying to figure out how to create/add a private key. I'm not sure if they way I added the information is the best way and I'm sure it could use some polishing, but I think the information could help others like me not get confused.


browner12:
i'm not necessarily against the change, but the wording needs a little work. Telling a user to blindly hit 'Enter' 3 times is not great practice.

taylorotwell:
Will hold off for now.

PR #4029, opened 2018-02-07
Add note about updating fideloper/proxy

svenluijten:
Laravel 5.6 requires `fideloper/proxy` v `~4.0`, but this was not documented in the upgrade guide before. It is now 👍

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -15,7 +15,7 @@ Laravel 5.6 requires PHP 7.1.3 or higher.
 
 ### Updating Dependencies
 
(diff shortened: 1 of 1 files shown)


svenluijten:
Here's the commit where that happened: https://github.com/laravel/laravel/commit/6779a2759ec806e619ccae603955df120d0c4808#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780

browner12:
does it ***require*** it, or just a nicety to have?

svenluijten:
It requires it. But we'll have to wait for @fideloper to tag 4.0 as stable though, currently it still fails unless you set the `minimum-stability` to `dev` in `composer.json`.

PR #4034, opened 2018-02-08
Uses new square bracket syntax for array destructing assignment

vishnub1626:
This changes the list() for the new square bracket syntax for array destructuring assignment.


browner12:
cool, didn't know we could do this.

PR #4042, opened 2018-02-08
Update upgrade guide on TrustedProxies

nasrulhazim:
You need to remove `config/trustedproxy.php` in order to fix the issue on `Undefined class constant 'HEADER_CLIENT_IP'`.

Reference: [Stackoverflow](https://stackoverflow.com/a/48595052/4248425)

screen shot 2018-02-09 at 5 37 25 am

This is due to upgrade from Symfony 3 to Symfony 4.


taylorotwell:
That file is not part of the Laravel installation process.

PR #4044, opened 2018-02-09
Add note about specifying an index name

jk3us:
It wasn't clear that you can specify a name when creating an index.  This is important for migrations, so you can use the same name when dropping it in the `down()` method.


taylorotwell:
Wrong branch. Please send to 5.6.

PR #4047, opened 2018-02-10
[5.6] Inclue cache config warning for deployment

laurencei:
Add a warning about using `env()` during `cache:config` for the deployment section.

This is already mentioned in the configuration page, but it could/is easily missed by people who only read the deployment section.


laurencei:
p.s. @taylorotwell - can you set the `docs` default branch to `5.6` on this repo? It's still set to `5.5`.

PR #4048, opened 2018-02-10
Add an asterisk (*) for Year

goodjack:


decadence:
Year asterisk is deleted from 5.6

Issue #3841, opened 2017-11-02
Request Lifecycle - confusing documentation for console requests


amadeann:
Request lifecycle documentation specifies that all request to Laravel app first hit the public/index.php script:

> The entry point for all requests to a Laravel application is the public/index.php file

and then bootstrap the app with bootstrap/app.php script:

> The index.php file loads the Composer generated autoloader definition, and then retrieves an instance of the Laravel application from bootstrap/app.php script.

When you keep reading you find out that:

> Next, the incoming request is sent to either the HTTP kernel or the console kernel, depending on the type of request that is entering the application.

As I understand it, the request first hitting the public/index.php script will always be sent to HTTP kernel. Only requests hitting the artisan script will be sent to console kernel. 

I think it would make sense to either point out in the 'First Things' section that some requests are going through artisan script or take out the part of requests being sent to HTTP or console kernel in the 'HTTP / Console Kernels' section.

There is a thread on SO as well with someone else getting confused with console requests handling ([link](https://stackoverflow.com/questions/43761213/how-to-use-console-kernel-is-laravel-php)).

Or am I getting the docs completely wrong?

laurencei:
Thanks. I submitted a PR for this, but Taylor has decided to leave it as is.

PR #4053, opened 2018-02-11
[5.6] Update lifecycle.md

laurencei:
The lifecycle documentation can be a little confusing. It currently says

> The entry point for all requests to a Laravel application is the `public/index.php` file

but we know this is not true for console requests. This is a first pass to clarify the two entry points to any Laravel application.

Closes https://github.com/laravel/docs/issues/3841


taylorotwell:
Has various wording issues I would need to clean up.

PR #4059, opened 2018-02-11
Add section for ignoring csrf when running tests

bobbybouwmann:
This question comes from the Laracasts forum and therefore I decided to shoot in a PR for this:  https://laracasts.com/discuss/channels/testing/is-only-csrf-middleware-ignored-for-testing-did-this-change-at-some-point

There is no place to find any information on why the csrf token is disabled when running tests. At least I couldn't find the issue. 

It should still be documented. Right now it's the only middleware that's ignored during tests, so that why I added it to the `csrf.md` file.

Let me know if I need to phrase this differently or if there is a why reason somewhere. Thanks in advanced.


rosswintle:
Thanks for this.  I started the Laracasts discussion.  Great to get a PR off the back of it.

I'd like to see this documented on the Testing page too.  When I get some time to focus I'll see if I can add this to the PR too.

Thanks again!

taylorotwell:
Doesn't follow the formatting of the rest of the Laravel documentation.

rosswintle:
I don't understand. Could you suggest how we can change this to make it match the formatting?

Or can we leave you with an action to document this feature so that other people can discover it without having to dive into the code to figure it out?

Thanks

laurencei:
@rosswintle @bobbybouwmann ,

As a single sentance, perhaps it would be better as a "tip"?

Otherwise it probably needs to be fleshed out a little bit further. The original post here mentions including "why" - but thats not in the PR itself.

In regards to formating - it should be

```diff
- ## TESTING
+ ## Testing
```

bobbybouwmann:
@laurencei I have no clue why this was added, but at least it should be mentioned in the documentation so you know what happens when running tests.

If it's only the heading than why would the PR be closed...

laurencei:
@bobbybouwmann it was added in 2015, because otherwise every time you run a unit test, you would need to manually incliuded the CSRF token in the post request, which is tedious.

I've done another PR - not sure if Taylor will like this format - but we'll get there: https://github.com/laravel/docs/pull/4083

PR #4061, opened 2018-02-11
Update notifications.md

MrAtiebatie:
I have been debugging for hours why my code was not sending an SMS to me. Happens to be that when you use a custom phone attribute using `routeNotificationsForNexmo` Laravel doesn't throw an exception when the attribute doesn't exists on your model. Changing this to the above example, PHP throws a warning. For debugging purposes this is a better solution...


taylorotwell:
Not necessary on most applications.

Issue #3049, opened 2017-01-28
Inaccuracy/incomplete info about Laravel Passport token lifetimes


brunen9:
[Here](https://github.com/laravel/docs/blob/5.4/passport.md#token-lifetimes) docs says:

> By default, Passport issues long-lived access tokens that never need to be refreshed

But it is not right since Nov 2016, and at present default TTL is only 1 year.

laurencei:
Thanks @brunen9,

I've submitted a PR for this to be fixed.

PR #4063, opened 2018-02-12
Fix Passport token expiry

laurencei:
Passport tokens no longer last forever. They expire after 1 year.

I found the PR where this change was made (although it doesnt say why): https://github.com/laravel/passport/commit/5a69318d810b4c99373d163b220c9808ad868261

Closes https://github.com/laravel/docs/issues/3049


DojoGeekRA:
@laurencei the relevant PR is https://github.com/laravel/passport/pull/185 😃

Issue #2851, opened 2016-11-21
Omission in Authentication: Path Customization section


batiste93:
Here : https://laravel.com/docs/5.3/authentication#included-authenticating

At the path customization section, the docs fail to mention to edit : `Http/Middleware/RedirectIfAuthenticated.php`
to change to the post-auth redirect location

laurencei:
Thanks @batiste93 - I've submitted a PR with this info for Taylor to consider.

PR #4068, opened 2018-02-12
Update authentication.md

laurencei:
The docs currently forget to include changing the auth redirect URI in `app/Http/Middleware/RedirectIfAuthenticated.php`.

I'm not sure the best way to include it in that area. i.e. as a "tip", or just another paragraph?

Either way - it needs to be listed as another place to change the `/home` URI if you modify it.

Solves https://github.com/laravel/docs/issues/2851


taylorotwell:
Reformatted to not use tip.

Issue #3590, opened 2017-08-31
More guidance for Horizon dashboard authentication


kamui545:
In the documentation, we got the following section:

# Dashboard Authentication

Horizon exposes a dashboard at /horizon. By default, you will only be able to access this dashboard in the local environment. To define a more specific access policy for the dashboard, you should use the Horizon::auth method. The auth method accepts a callback which should return true or false, indicating whether the user should have access to the Horizon dashboard:

```php
Horizon::auth(function ($request) {
    // return true / false;
});
```

--

What do you think about adding some advices on **where** exactly you would **recommend** to drop this snippet? Personally I've put it in `AppServiceProvider::boot()` but maybe there's a more suitable place and more guidance may be appreciated for newcomers who could be lost?

chrislentz:
Has anyone found anymore info on this question?

browner12:
there is no "right" answer, but your `AppServiceProvider` is a decent solution. You could also make your own custom Service Provider.

I think adding a suggestion is a good idea.

acidjazz:
@browner12 Placing the Auth facade and running `Auth::check()` results in a false even when a user is logged in in `boot()`, you sure this is good place to put this?

browner12:
the session is probably not populated yet at this point, so that's why it doesn't see a user yet. What you actually want to do is register a callback with `Horizon:auth()` to be checked later on in the lifecycle.  Try this:

```php
/**
 * Bootstrap any application services.
 *
 * @return void
 */
public function boot()
{
    //horizon authentication
    Horizon::auth(function(Request $request){
        return $this->app->environment('local') || ($request->user() && $request->user()->id === 1);
    });
}
```

This will register the closure, and when it is called, the `Request` will be passed in.

leadranger:
I put it in my web.php routes file, seems to work fine.

sebastiansulinski:
If you check default `HorizonServiceProvider` you'll see the method, which registers horizon's routes

```php
protected function registerRoutes()
{
    Route::group([
        'prefix' => config('horizon.uri', 'horizon'),
        'namespace' => 'Laravel\Horizon\Http\Controllers',
        'middleware' => config('horizon.middleware', 'web'),
    ], function () {
        $this->loadRoutesFrom(__DIR__.'/../routes/web.php');
    });
}
```

[...]

PR #4070, opened 2018-02-12
Add tip for where to place Horizon auth

laurencei:
The open issue here https://github.com/laravel/docs/issues/3590 is a request on suggesting where to place the Horizon auth example.

I appreciate you probably left it vague to encourage people to place it where it is most suitable for their application - but the open issue has 20x +1's - which is _by far_ the most of any issue on the docs page ever, so it is clearly something that people want guidence on.

I've included a tip as a starting point...


taylorotwell:
I worked it into the text flow. I don't like to over-do the tips for everything.

PR #4073, opened 2018-02-12
Add Heroku CI notes for Dusk

tillkruss:
Took me forever to get Dusk working on Heroku CI.


dzuelke:
Nice. How did you end up with that double `nohup` solution? I wonder if we can do anything on the Heroku side (different job control mode or whatever) that makes it possible to skip such workarounds.

The problem was that the `php artisan dusk` call would finish, but the process never exit (or never register as finished by Heroku CI), right?

tillkruss:
@dzuelke: Yes, the tests would never finish without `nohup`.

CircleCI and Travis CI actually allow to start background tasks, that would definitively make it easier. See https://laravel.com/docs/5.6/dusk#continuous-integration

PR #4075, opened 2018-02-12
Add better solution to the 767 bytes index limitation

pedzed:
Also explaining why 191 is the limit for mb4.


migrations.md
```diff
@@ -391,7 +391,31 @@ Command  |  Description
 
 #### Index Lengths & MySQL / MariaDB
 
-Laravel uses the `utf8mb4` character set by default, which includes support for storing "emojis" in the database. If you are running a version of MySQL older than the 5.7.7 release or MariaDB older than the 10.2.2 release, you may need to manually configure the default string length generated by migrations in order for MySQL to create indexes for them. You may configure this by calling the `Schema::defaultStringLength` method within your `AppServiceProvider`:
+Laravel uses the `utf8mb4` character set by default, which includes support for storing "emojis" in the database. If you are running a version of MySQL older than the 5.7.7 release or MariaDB older than the 10.2.2 release, you could either decrease the index size at the expense of the string size, or decrease the string size at the expense of the index size.
+
+##### Decreasing Index Size
+You can decrease the index size as follows:
+
+    $table->index([DB::raw('email(20)')]);
```
pedzed: These indexes can be up to `191` as well, perhaps something to add?

pedzed:
Is there a way to set the default index size like you can do `Schema::defaultStringLength()`?

taylorotwell:
I think this is maybe something that is a better fit a blog post.

pedzed:
@taylorotwell This is one of the most confusing Laravel errors because it simply does not work out of the box for many MySQL users. I do not think it should be just dismissed and shoved to bloggers. I mean, that's what the documentation should be for, right? To guide people into getting things working.

If I slim it down, would you accept the PR?

PR #4076, opened 2018-02-12
Added an example of storing a mailable.

daveismyname:
The example show a use case of calling the mailable render method.


taylorotwell:
Re-written and committed.

PR #4078, opened 2018-02-12
[5.2] Show 5.2 does not support PHP7.2

laurencei:
If we are not accepting PRs to fix Laravel 5.2 to give it support for PHP7.2, then we should be clear in the docs.

As per PR that was closed https://github.com/laravel/framework/pull/23136


laurencei:
wrong branch

Issue #3077, opened 2017-02-04
Gate example differs with recommendation


bkilshaw:
Not a huge issue, but under authorisation:

> Gates are most applicable to actions which are not related to any model or resource, such as viewing an administrator dashboard. In contrast, policies should be used when you wish to authorize an action for a particular model or resource.

It recommends you use a Policy for actions related to models or resources, and Gates for other permissions. However, the code example for Gates creates a permission to update a post, which should be a policy.

```
    Gate::define('update-post', function ($user, $post) {
        return $user->id == $post->user_id;
    });
```

The code example can be misleading for someone trying to figure out where they should be using Gates vs Policies.

laurencei:
I've submitted a PR for this to be changed. It's up to Taylor if it is accepted or not.

PR #4084, opened 2018-02-14
Update authorization.md

laurencei:
This attempts to solve https://github.com/laravel/docs/issues/3077

Essentially the docs currently say 

> Gates are most applicable to actions which are not related to any model or resource, such as viewing an administrator dashboard.

But then it goes on to give an example of using a model, which is almost the opposite of the statement above.

This PR changes the example to use the admin dashboard idea that is talked about, so it "flows" better as you read it.

### DIFF
--- a/authorization.md
+++ b/authorization.md
@@ -43,8 +43,8 @@ Gates are Closures that determine if a user is authorized to perform a given act
     {
         $this->registerPolicies();
 
-        Gate::define('update-post', function ($user, $post) {
-            return $user->id == $post->user_id;
+        Gate::define('view-admin-dashboard', function ($user) {
+            return $user->is_admin;
         });
     }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I still think the original example is a little better example of authorization use-cases.

bkilshaw:
@taylorotwell Yes but the documentation states the best practice is to do one thing, and the code example says otherwise:

From the Introduction section:
> Gates are most applicable to actions which are not related to any model or resource, such as viewing an administrator dashboard. In contrast, policies should be used when you wish to authorize an action for a particular model or resource.

And then the first code example under the Gates section is:

`    
    Gate::define('update-post', function ($user, $post) {
        return $user->id == $post->user_id;
    });
`

So the docs state that the best practice to authorize model methods is to use a Policy, yet the code example uses a Gate.

It would seem that either the code example needs to be updated as per this PR, or the introduction paragraph needs to be updated to either reflect new best practices (using Gates for models) or that paragraph needs to be removed outright if there is no best practice regarding this subject.

PR #4085, opened 2018-02-14
DB Read's and Write's Hosts should be an array

itsromack:
Configuration for the database hosts for the read and/or write resource should be an array


laurencei:
The original docs are correct as they are. I use that exact config without
issue.

PR #4087, opened 2018-02-14
Rename HipChat link to Stride

Ismaile7:
Before the PR, when clicking on HipChat (https://www.hipchat.com) in https://laravel.com/docs/5.6/queues#failed-job-events, we got redirected to https://www.stride.com, so the PR is here to rename the link and use Stride URL directly.
(another commit has been added afterwards to remove an unused HipChat anchor link in envoy documentation, here it is : https://laravel.com/docs/5.6/envoy#hipchat-notifications)

### DIFF
--- a/envoy.md
+++ b/envoy.md
@@ -159,7 +159,6 @@ If you would like to be prompted for confirmation before running a given task on
     @endtask
 
 <a name="notifications"></a>
-<a name="hipchat-notifications"></a>
 ## Notifications
 
 <a name="slack"></a>
@@ -177,4 +176,3 @@ You may provide one of the following as the channel argument:
 - To send the notification to a channel: `#channel`
 - To send the notification to a user: `@user`
 </div>
-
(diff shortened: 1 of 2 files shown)


laurencei:
Given most other things in Laravel now have a Slack default (i.e. notifications, logs etc) - perhaps Slack makes more sense now?

Ismaile7:
You are right, it would be more consistent with the rest. The PR initial goal is to not use HipChat naming, referencing Slack might be even better in term of consistency. Should I submit another PR for this ?

laurencei:
It's up to Taylor. It's just an idea of mine.

Wait to see what direction he wants to take.

Ismaile7:
Alright, I submitted another commit to remove an unused HipChat anchor link in Envoy documentation page. This might lead to another discussion regarding the support for HipChat in laravel/envoy. It is still there but not documented anymore.

Issue #3594, opened 2017-08-31
Adding After Hooks To Form Requests documentation is confusing


gms8994:
The documentation provided for [Adding After Hooks To Form Requests](https://github.com/laravel/docs/blob/4c81b4fdbde81e6aafb84a68381ab918ad1fe077/validation.md#adding-after-hooks-to-form-requests) is confusing, to say the least. It's an "after" hook that runs "before" validation rules are evaluated?

Does this mean that it's running the closure after the data validation, but you place it in your code somewhere before the validation is begun?

johnsky007:
I agree with you absolutely, te documentation is very confusing in this case and I did get the point!

AdolphYu:
为什么不是在rules 执行之后执行呢？不符合先基础验证后业务验证这个先后顺序

laurencei:
Ok - so to clarify - you can call the validator BEFORE it runs, and attach AFTER rules for when it is complete.

See my PR that Taylor closed for more info.

If anyone has a better way to write this, please feel free to submit a PR :)

PR #4088, opened 2018-02-14
Update validation.md

laurencei:
I've read this paragraph like 10 times now - and I am 95% sure this should be `after`, not `before`?

Further down the page it even says

> The validator also allows you to attach callbacks to be run **after** validation is completed

Solves https://github.com/laravel/docs/issues/3594

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -215,7 +215,7 @@ If validation fails, a redirect response will be generated to send the user back
 
 #### Adding After Hooks To Form Requests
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Nope. Original docs are correct. You are confusing what the `after` method does on the Validator. You can call it BEFORE validation runs in order to attach callbacks that should be run AFTER validation runs.

laurencei:
Arh... ok - now I get it...

Issue #3915, opened 2017-12-14
Eloquent Events missing example


garyross:
In the[ Eloquent > Events documentation](https://laravel.com/docs/5.5/eloquent#events) the following code is given:

```
protected $dispatchesEvents = [
        'saved' => UserSaved::class,
        'deleted' => UserDeleted::class,
    ];
```

At this point the documentation should have an example `UserSaved::class`. Instead, it sends you to the https://laravel.com/docs/5.5/events page which is a very detailed explanation of Laravel Events. Using artisan as so:
`php artisan make:event UserSaved `
 creates

```
namespace App\Events;

use Illuminate\Broadcasting\Channel;
..
use Illuminate\Contracts\Broadcasting\ShouldBroadcast;

class UserSaved
{
    use Dispatchable, InteractsWithSockets, SerializesModels;

 
    public function __construct()
    {
        //
    }

    public function broadcastOn()
    {
        return new PrivateChannel('channel-name');
    }
}
```

but the purpose of the `broadcastOn` doesn't appear anywhere.

laurencei:
Hi @garyross - what exactly do you think can be improved here?

From what I can see - you can create a `UserSaved` event. Then in the events section, you can see the options about how to respond to that event?

garyross:
Hi @laurencei 
Thanks for your response.
Well, crucially the docs don't show that the UserSaved event receives an instance of the model itself through the constructor, and without that knowledge it's not easy to see how to make the UserSaved class respond to the event:

```
class UserSaved{
    public function __construct($user)      // not in docs as far as I can see
    {
        $this->user = $user;
    }
}
```

The docs, which in general are excellent, here seem to be very helpful provided you are familiar with events as opposed to the user who's coming to it from scratch. A simple example of a UserSaved event doing something practical (such as saving a hashId) would help a lot for people coming to this page with no knowledge of events.

laurencei:
I think it is covere by this part: https://laravel.com/docs/5.6/events#defining-events

> An event class is a data container which holds the information related to the event. For example, let's assume our generated OrderShipped event receives an Eloquent ORM object:

Does that clarify the situation? Or do you have an idea how it could be made clearer perhaps?

garyross:
Hi @laurencei,
I think the problem is the disconnect between the example in https://laravel.com/docs/5.5/eloquent#events page, and the example in the https://laravel.com/docs/5.5/events#defining-events page.

[...]

PR #4097, opened 2018-02-18
Update eloquent examples

laurencei:
If you read the Eloquent page from the start, the first 80% of examples use the example of `App\Flight` to explain each concept.

However, towards the end, the last few examples suddenly introduce a new `App\User` concept that has not been used before. Anyone following the examples from start to finish can no longer use the `App\Flight` examples they have built so far.

This PR aims to continue the single use of `App\Flight` throughout all examples. This means people dont get introduced to an additional concept mid way through the docs, and it keeps the examples consistently themed throughout.

Partially solves https://github.com/laravel/docs/issues/3915 (more to come depending on the outcome of this PR).

_(The intention is if this PR is accepted, to do another PR for the Events pages, and switch from OrderShipped examples to Flight examples, as each doc is interlinked, so keeping a consistent theme throughout makes it easier for everyone to follow from scratch and see how the concepts are linked. At the moment the themes are disjointed and not easy to follow across pages)._

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -547,7 +547,7 @@ Writing a global scope is simple. Define a class that implements the `Illuminate
     use Illuminate\Database\Eloquent\Model;
     use Illuminate\Database\Eloquent\Builder;
 
-    class AgeScope implements Scope
+    class FlightTimeScope implements Scope
     {
         /**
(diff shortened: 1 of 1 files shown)


taylorotwell:
I really don't see this as confusing though?

Issue #3173, opened 2017-03-06
Broadcasting: It's not safe to use socket.io library from Laravel Echo Server directly


kkomelin:
See [the corresponding issue](https://github.com/tlaverdure/laravel-echo-server/issues/132) in the Laravel Echo Server issue queue.

Therefore, I believe some corrections to the ["Driver Prerequisites - Socket.IO"](https://github.com/laravel/docs/blob/5.4/broadcasting.md#socketio) section of the broadcasting.md are necessary.

barryvdh:
I actually recently submitted this change to the docs (https://github.com/laravel/docs/pull/3004)
And coincidentally the socket went down for like a week or so, last week: https://github.com/socketio/socket.io/issues/2883
Which would have resulted in the same problem, but for everyone using the CDN instead of just people where the socket is not running.

It could perhaps be described better, but usually your socket should be always running, and checking if `io` is defined is an easy way to see if it *is* running. Otherwise you would get continuous polling to the SocketIO server, without being aware that connect failed/socket is down.

kkomelin:
Thanks for your comment @barryvdh. 

I think users should be warned about consequences of the situation when Socket.io server (Laravel Echo Server) stops for some reasons. It will result in a complete failure of all JS code which follows the code:
```
    window.Echo = new Echo({
        broadcaster: 'socket.io',
        host: window.location.hostname + ':6001'
    });
```
and since this code is in the bootstrap.js (at the top), everything will fails.

> It could perhaps be described better, but usually your socket should be always running, and checking if io is defined is an easy way to see if it is running.

Firstly, you can't hope that your socket.io server will be always available.
Secondly, a realtime component is usually a minor part of an app and it should not become [a single point of failure](https://en.wikipedia.org/wiki/Single_point_of_failure) for the entire Laravel app.
Thirdly, from my experience, if a piece of code is included to the official documentation or to the default Laravel installation, users tend to trust the code. 
From my point of view, the checking for `io` existence should be included in the documentation and the default Laravel installation because it's a serious issue.

In [my article](https://komelin.com/articles/realtime-apps-laravel-echo-tips-and-tricks#laravel-echo-server-fails) I suggested two solutions that can work together or separately:

1) Copy Socket.io lib to your public/js folder (e.g. via Webpack) and add it to your main page template:
```
<script src="{{ asset('js/socket.io.min.js') }}"></script>
```
2) Always check if `io` object exists before using Echo:
```
if(typeof io !== 'undefined') {
[...]

PR #4098, opened 2018-02-18
[WIP] Better Socket.io instructions

laurencei:
This is an attempt to solve https://github.com/laravel/docs/issues/3173

Essentially there are some concerns that the default example in the docs for Socket.IO has the potential to cause a complete failure of all your JS scripts if the socket.io is not available for any reason (i.e. an outage). So if the server is not running (for any reason) then the JS is not available, and the whole thing falls apart, but brings down the rest of your unrelated JS scripting with it.

The advice from @barryvdh and @kkomelin is to modify the example to make it more robust to failures.

I've put **[WIP]** as I want to get confirmation from @barryvdh and/or @kkomelin that this PR solves their concerns first

### DIFF
--- a/broadcasting.md
+++ b/broadcasting.md
@@ -95,9 +95,15 @@ If you are going to pair the Redis broadcaster with a Socket.IO server, you will
 
     <script src="//{{ Request::getHost() }}:6001/socket.io/socket.io.js"></script>
 
-Next, you will need to instantiate Echo with the `socket.io` connector and a `host`.
+Add the following to your NPM dependancy:
+
+    npm add socket.io-client
+
+Next, you will need to instantiate Echo with the `socket.io` connector and a `host`. 
 
     import Echo from "laravel-echo"
+
+    window.io = require('socket.io-client');
 
     window.Echo = new Echo({
         broadcaster: 'socket.io',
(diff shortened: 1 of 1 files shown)


broadcasting.md
```diff
@@ -95,10 +95,16 @@ If you are going to pair the Redis broadcaster with a Socket.IO server, you will
 
     <script src="//{{ Request::getHost() }}:6001/socket.io/socket.io.js"></script>
 
-Next, you will need to instantiate Echo with the `socket.io` connector and a `host`.
+Add the following to your NPM dependancy:
```
kkomelin: @laurencei "dependency" please

broadcasting.md
```diff
@@ -95,10 +95,16 @@ If you are going to pair the Redis broadcaster with a Socket.IO server, you will
 
     <script src="//{{ Request::getHost() }}:6001/socket.io/socket.io.js"></script>
```
kkomelin: We don't need this if we add this "window.io = require('socket.io-client');"

laurencei:
@kkomelin - ok, changes made. How is that?

taylorotwell:
Please re-submit when you have personally tested it. Have you?

PR #4105, opened 2018-02-20
Fix PHP version formatting

highsolutions:
Earlier:
- PHP `>= 5.5.9` & `< 7.2.0`
Now:
- PHP >= 5.5.9 & < 7.2.0

This how it looks now:
- PHP &gt;= 5.5.9 & `

so we don't have info about PHP versions, but also lost info about another requirements

### DIFF
--- a/installation.md
+++ b/installation.md
@@ -16,7 +16,7 @@ The Laravel framework has a few system requirements. Of course, all of these req
 However, if you are not using Homestead, you will need to make sure your server meets the following requirements:
 
 <div class="content-list" markdown="1">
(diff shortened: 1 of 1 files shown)


laurencei:
No - this wont fix it, as it what is was before I did the PR the other day to try to resolve the issue. 

Something about the `<` symbol is breaking everything. Let me look into it.

laurencei:
Thanks for the PR. This will now be solved by https://github.com/laravel/docs/pull/4108 - which uses the HTML symbol for `<` to not break the parser.

Thanks for raising the issue.

PR #4106, opened 2018-02-20
Fix installation requirements

lewisfff:
Noticed it is also an issue for 5.3


laurencei:
Thanks for the PR. This will now be solved by https://github.com/laravel/docs/pull/4109

Unfortuantely the `<` symbol will still break in this instance.

PR #4108, opened 2018-02-20
[5.2] Definitive fix for markdown issue

laurencei:
I've personally tested this on a cloned Laravel.com website with the docs.

I dont know what it is about a single `<` - but it totally breaks the markdown parser, even if wrapped in code blocks. (good to be mindful of in any future PRs we see).

The use of the HTML symbol code for `<` solves it and displays correctly (on my local server).


taylorotwell:
Just used "between" instead of this back and forth fixing.

PR #4124, opened 2018-02-26
Add update instructions for fideloper/proxy

srmklive:
I had to upgrade my Laravel 5.5 application to 5.6 a few days back by following the upgrade instructions as listed [here](https://laravel.com/docs/5.6/upgrade#upgrade-5.6.0). I ran into upgrade errors when i updated the `fideloper/proxy` to `~4.0`. I couldn't properly run `composer update` due to this.

I then had to create a fresh 5.6 application, and copy the new `TrustProxies` middleware & the `config/trustedproxy.php` from 5.6 app to the 5.5 app which i was trying to upgrade. After performing this step, i was able to successfully upgrade my 5.5 app to 5.6.

I hope this helps out anyone performing the manual upgrade step from 5.5 to 5.6

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -15,7 +15,7 @@ Laravel 5.6 requires PHP 7.1.3 or higher.
 
 ### Updating Dependencies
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
This information is already in the upgrade guide.

PR #4126, opened 2018-02-26
"Trusting All Proxies": "**" is deprecated, use "*"

calebporzio:
I was source diving fideloper's TrustedProxies package and found this:
```
        // Trust any IP address that calls us
        // `**` for backwards compatibility, but is depreciated
        if ($trustedIps === '*' || $trustedIps === '**') {
            return $this->setTrustedProxyIpAddressesToTheCallingIp($request);
        }
```
[see file on github](https://github.com/fideloper/TrustedProxy/blob/master/src/TrustProxies.php)

It looks like the package docs still make a distinction between the two, so probably don't need this - but figured I'd put it out there.

### DIFF
--- a/requests.md
+++ b/requests.md
@@ -400,11 +400,11 @@ To solve this, you may use the `App\Http\Middleware\TrustProxies` middleware tha
 
 #### Trusting All Proxies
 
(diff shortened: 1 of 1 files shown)


laurencei:
ping @fideloper - can you please advise?

fideloper:
👍

PR #4127, opened 2018-02-26
Add section explaining bootable traits

ImLiam:
Add an explanation and example of how a trait for a model can register its own boot method as this functionality is currently undocumented.

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -21,6 +21,7 @@
     - [Local Scopes](#local-scopes)
 - [Events](#events)
     - [Observers](#observers)
+- [Traits](#traits)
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't mind documenting the feature but I don't want to show it registering event listeners, as I prefer to steer people towards using class based event listeners on Eloquent models. If you have another example of what that might be used for feel free to re-submit.

sasokovacic:
@taylorotwell 
Is there a reason you don't want to show it registering event listeners? Can you please explain? I'm building a Laravel package and would like to listen on eloquent events. 
Should

PR #4129, opened 2018-02-27
Add section explaining bootable traits with scopes

ImLiam:
Continuing from #4127, but adding a global scope instead of event.


taylorotwell:
Let me think about this entire feature a bit and decide what would be the best place to document it.

PR #4130, opened 2018-02-27
Add SFTP driver configuration

mputkowski:
Reference: [#23308](https://github.com/laravel/framework/pull/23308)

### DIFF
--- a/filesystem.md
+++ b/filesystem.md
@@ -54,8 +54,9 @@ When using the `local` driver, all file operations are relative to the `root` di
 
 #### Composer Packages
 
-Before using the S3 or Rackspace drivers, you will need to install the appropriate package via Composer:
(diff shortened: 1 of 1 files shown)


decadence:
Does it have `path` option?

mputkowski:
@decadence it doesn't have. You can check all options [here](https://github.com/thephpleague/flysystem-sftp/blob/master/src/SftpAdapter.php#L54),

decadence:
Actually it does. It's called `root`.

mputkowski:
@decadence so why are you asking? You asked specifically of `path`, the answer is obvious...

decadence:
I asked about option for specify path on server thinking it would have name path or whatever. Then I found root option for this.

PR #4131, opened 2018-02-28
[5.6] Update list format

mirorauhala:
Makes navigation and reading Available Assertions faster. No more `ctrl-f` or `cmd-f` to find the right assertion.


taylorotwell:
All of these sentences should use ":" instead of having a "." at the end... just like the collection docs. Thanks.

image

mirorauhala:
Switched!

PR #4139, opened 2018-03-02
valet installation additional composer step

aurawindsurfing:
I found this step missing in my initial install of valet, it is not an obvious thing for beginners to understand what a $PATH is.


taylorotwell:
This will depend on OS.

aurawindsurfing:
Agreed. This is something that however put me off valet at first. Thanks anyway!

PR #4155, opened 2018-03-08
Update eloquent-resources

schmidtflo:
Adding "Adding Authorization Data" to provide information about using policies in the resource response page


schmidtflo:
Sorry, please don't merge. As i know now, the before method is not called this way. Does anyone know how to implement it right?

taylorotwell:
I'm not sure this needs to be documented? You can add whatever you want to the array. That seems given.

PR #4164, opened 2018-03-13
crontab addition

NathnC:
thought it makes it more clear on where to put it, I was a bit of a fool and was simply entering it thinking it was some command >:(


taylorotwell:
That's not the only way to do it.

PR #4166, opened 2018-03-13
DB Read's and Write's Hosts should be an array

itsromack:
Configuration for the database hosts for the read and/or write resource should be an array


decadence:
@itsromack can you explain why?

connectkushal:
@itsromack can you explain why?

itsromack:
@decadence @connectkushal due to the possibility of multiple database replicas and slaves

connectkushal:
thnx

PR #4168, opened 2018-03-15
phpunit command executable file location

eissasoubhi:
Specify the phpunit command executable file location when running tests from the terminal


taylorotwell:
This may not always be necessary.

PR #4169, opened 2018-03-15
Update migrations.md

stevesweets:
Would be useful to notify of possible unexpected behaviour here.

https://ma.ttias.be/laravel-mysql-auto-adding-update-current_timestamp-timestamp-fields/

https://github.com/laravel/framework/issues/22900

https://github.com/doctrine/dbal/issues/2558


rjksn:
It sure as heck would be useful.

PR #4170, opened 2018-03-15
The middleware terminate method descripton is misleading.

DevparkAnders:
From the current version of docs you can think that something will be done on server side, after the browser will really receive the response from server. THIS IS NOT TRUE - if you will put e.g. sleep(10) in the middleware terminate method - you will see in the browser that you will have to wait those additional 10 second for the response from server to be received.

I've checked this using php artisan serve and also by running it on Apache 2.4 with php 7.2 installed.


taylorotwell:
Has typos.

PR #4182, opened 2018-03-19
Update bootstrap/autoload.php

JLMorales:
This pull request attempts to call attention to the fact that the bootstrap/cache/compiled.php file will never again exist, but the bootstrap/autoload.php file will still try to require it.

Because this change is in regards to some code in `bootstrap/autoload.php` checking the existence of a file (`compiled.php`) that the upgrade guide already instructs us to remove, it feels pertinent that it be mentioned to make sure no one inadvertently skips the removal of this bit also. Since we already have the users checking inside the `bootstrap` folder already, might as well have them do that too as part of the upgrade flow and save them from having to check that file with the comparison tool.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -15,7 +15,28 @@ Update your `laravel/framework` dependency to `5.4.*` in your `composer.json` fi
 
 #### Removing Compiled Services File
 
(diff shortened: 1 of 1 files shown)


browner12:
This should be covered by the "Miscellaneous" section at the bottom of the Upgrade Guides.

PR #4184, opened 2018-03-19
Update billing.md

ColinLaws:
Clarifying the relationship between users and subscriptions for the database migration example by adding an auto-incrementing primary key to the users table, and changing the related user_id column in the subscriptions table to an unsigned integer to match the users table PK data type. Also added a FK from Subscriptions to Users.


ColinLaws:
I can see that in the Billable trait, the relationship is automatically added to the model, so maybe not add the relationship to the table? I do still feel like adding a PK to the users table, and a corresponding unsigned integer FK to the subscriptions table might be nice.

taylorotwell:
Wrong branch. Not sure which version you intend this for.

ColinLaws:
Sorry about that Taylor, I guess I tried working too hastily on that one. I'll review my PR and resubmit, sorry for the inconvenience.

PR #4187, opened 2018-03-20
make sense of the grouped wheres example

jarektkaczyk:
same as https://github.com/laravel/docs/pull/4180 for 5.3


jarektkaczyk:
@taylorotwell help me understand, so we save our time in the future - you don't want PRs for relevant parts of the olders versions of the docs?

PR #4188, opened 2018-03-20
Updated billing.md to contain proper table relationships

ColinLaws:
I request to add a PK to the users table in the database migrations section and a corresponding FK to the subscriptions table.


taylorotwell:
This isn't creating a table. It's updating an existing one.

ColinLaws:
@taylorotwell Do you still want the foreign key to be updated to an unsigned integer to match the 'increments' type that people are likely to use on their users table?

PR #4191, opened 2018-03-21
Modified subscription table structure on billing.md migration example

ColinLaws:
Changed user_id column to uint, added foreign key to migration example.


ColinLaws:
@taylorotwell  Woops, there are two sections in there with migration examples for Stripe and for Braintree, my commit only had the migration example change for the Stripe section.

PR #4195, opened 2018-03-24
Update helpers.md

troyporter:
It wasnt immediately clear to me that str_is could be used without an asterisk.  So i didnt know there was a helper that could test two strings for case-sensitive equivalence.

Maybe i'm just slow :)


taylorotwell:
This wouldn't be the most performant method of doing this.

troyporter:
I see what you mean...theres a few things going on under the hood.  Point taken.

PR #4199, opened 2018-03-25
Add $request->validated() documentation

manniL:


taylorotwell:
I think this would probably belong with the form request documentation.

PR #4206, opened 2018-03-31
Add return $this in set mutator

LaravelMonster:
While chaining methods, we end up with unable to call method save on null (when $this is not returned from mutator).

\App\Model::find(1)->setAttribute('first_name','laravel')->save();

### DIFF
--- a/eloquent-mutators.md
+++ b/eloquent-mutators.md
@@ -83,6 +83,8 @@ To define a mutator, define a `setFooAttribute` method on your model where `Foo`
         public function setFirstNameAttribute($value)
         {
             $this->attributes['first_name'] = strtolower($value);
+            
+            return $this;
         }
     }
 


michaeldyrynda:
The idea of mutators is they're called implicitly using PHP's `__set` magic method. I'm not sure why you're calling the methods directly; that's not documented behaviour at all.

LaravelMonster:
Does it mean we should not use setAttribute directly?

michaeldyrynda:
Correct; Laravel will delegate to the `set<Attribute>Attribute` method when you call it directly on the model via PHP's `__set` magic method.

```php
// This is the same as your example above
$model = App\Model::find(1);
$model->first_name = 'laravel'; // Calls the model's setFirstNameAttribute() method
$model->save();
```

This is the documented usage of Laravel's mutator.

PR #4208, opened 2018-04-01
random doesn't always return Collection

decadence:
https://github.com/laravel/framework/pull/23759

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -1377,7 +1377,7 @@ You may optionally pass an integer to `random` to specify how many items you wou
 
     // [2, 4, 5] - (retrieved randomly)
 
-However if Collection has less items than you requested method will fail with `InvalidArgumentException`.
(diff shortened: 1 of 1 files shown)


collections.md
```diff
@@ -1377,6 +1377,8 @@ You may optionally pass an integer to `random` to specify how many items you wou
 
     // [2, 4, 5] - (retrieved randomly)
 
+However if Collection has less items than you requested method will fail with `InvalidArgumentException`.
```
browner12: Just a couple grammar changes.

> However, if **the** Collection has **fewer** items than you requested, **the** method will fail with `InvalidArgumentException.

https://www.grammarly.com/blog/fewer-vs-less/

decadence: As you say :) I am not native english speaker.

browner12: no problem, glad to help!

PR #4215, opened 2018-04-04
[5.6] Update authorization.md

sileence:
I found this part regarding Gate resource confusing, as it says you may "add or override" but after doing some tests and checking the source code I found the third argument does not add but instead *overrides* the default options.


taylorotwell:
I feel like we discussed this several times in previous PRs. Did you see them?

sileence:
I found this one: https://github.com/laravel/docs/pull/3446/files

Which sounds fine to me, but currently the docs says you may _override or add_. I find the _or add_ confusing as the third argument will override the array of default rules.

PR #4217, opened 2018-04-04
middleware.md wording

CalvinWilliams1012:
The sentence on line 28 said that the following code would check if the following age was greater than 200 while the code was checking if it was less than or equal to. Fixed the typo.

### DIFF
--- a/middleware.md
+++ b/middleware.md
@@ -25,7 +25,7 @@ To create a new middleware, use the `make:middleware` Artisan command:
 
     php artisan make:middleware CheckAge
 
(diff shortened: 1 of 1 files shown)


enderandpeter:
I would not consider this a typo. It's just that "greater than" is the negation of "less than or equal to", so the sentence is correct that in this example the user will be allowed access if `age` is greater than 200.

CalvinWilliams1012:
enderandpeter is correct, I just misread it. Trying to learn laravel and done too much doc reading. Sorry!

PR #4225, opened 2018-04-06
fixed a typo.

hasib32:
fixed typo for statuses. should be status


hasib32:
Found a typo  in the passport documentation. https://laravel.com/docs/5.6/passport#token-scopes
should be `status` instead of `statuses`

PR #4227, opened 2018-04-06
homestead.md: clarify installation procedure

vladdu:
I found the "Installing Homestead" section a bit confusing, like it was missing a bit. I added what I think would make it a bit clearer.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -88,7 +88,9 @@ If this command fails, make sure your Vagrant installation is up to date.
 
 #### Installing Homestead
 
-You may install Homestead by cloning the repository. Consider cloning the repository into a `Homestead` folder within your "home" directory, as the Homestead box will serve as the host to all of your Laravel projects:
+You may install Homestead by either cloning the repository or downloading and unpacking a release archive from GitHub. The former allows smoother updates. Consider using a `Homestead` folder within your "home" directory to host Homestead, as the Homestead box will serve as the host to all of your Laravel projects:
+
+If cloning the repository, follow these steps: 
 
     git clone https://github.com/laravel/homestead.git ~/Homestead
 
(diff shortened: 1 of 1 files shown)


enderandpeter:
I definitely appreciate your incentive to clarify the docs. However, it seems that in this case your suggestions would add a bit too much extraneous information.

For one, the URL makes it clear that the repo is on GitHub, and most people who get this far have some idea of the GitHub's features, so the user can go to the GitHub page if they want and download a zip file, but it would be a lot more convenient for them to use git to download and update the project, or checkout any version they want. So I don't see how adding that information makes the provided instructions clearer.

vladdu:
In that paragraph, there is already a reference to GitHub, for retrieving the latest version number.

The original text said "to install, you may clone the repo", which to me suggests that there is an alternative, but none was specified. Because of that, I wasn't sure if this installation is optional or another way to install the vagrant box.

Would it be better to just say "download  an unpack the latest archive from github releases"?

enderandpeter:
Downloading and unpacking the latest archive from GitHub releases is exactly what you are doing when you clone the repo from that URL. Yes, you may clone the repo or get those files some other way, but the instructions currently provided are succinct and explanatory enough to make clear to the reader that they can either clone the repo or go to the GitHub page or utilize GitHub's API or however they wish to get these files from GitHub.

vladdu:
Ok, you decide if it's better or not. I would still like to suggest to change "You may install Homestead by..." to "Install Homestead by...", because it's not an optional step.

enderandpeter:
I would say it is an optional step in the sense you are saying because you are right that you can either use git or unpack from a zip file. The point in the end is to get the files on your system, so yes one should get them there any way they prefer. Using a git client to manage these files is going to be the most common and recommended way for most development workflows but if someone wants to use GitHub's other features for obtaining the files, they are free to do so.

PR #4230, opened 2018-04-09
Warn about multi-dimensional arrays/collections

JanisE:
Arrays are not compared recursively. Furthermore, if the collection data contains an array, a misleading error "Array to string conversion" is output.

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -389,7 +389,7 @@ If you do not want to stop executing the script, use the [`dump`](#method-dump) 
 <a name="method-diff"></a>
 #### `diff()` {#collection-method}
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not sure this is the right wording. They are supported, just not compared ... it's just using array_diff under the hood.

JanisE:
@taylorotwell, they are not just not compared, PHP throws an exception if a collection contains an array.

Whatever the wording, currently the docs are missing info about arrays not being allowed in collections when using `diff` and `diffAssoc` methods.

A user reading the docs or using collections should not know what the implementation under the hood is.

PR #4241, opened 2018-04-16
Document the query builder chunkById method

jago86:
I think this useful method should be included in the docs.


taylorotwell:
Language doesn't fit rest of docs.

jago86:
Can you provide any details or guidelines to language fits the rest of docs?

PR #4244, opened 2018-04-18
add view:cache Artisan command

chapeupreto:


taylorotwell:
I don't want to document this quite yet.

PR #4259, opened 2018-04-27
Added the missing but quite crucial laravel filesystem driver caching docs.

malc0mn:
I noticed that a very crucial section about caching is missing from the filesystem documentation. After diving into the code I was pleased to see that the CachedAdapter from the excellent Flysystem bundle is supported by Laravel but is simply not documented.

As this is a vital part of Flysystem for *any* application or website, I decided to write up the docs myself, so there you have it :smile:


malc0mn:
Just for my own clarification: how will this get 'backported' to the laravel 5.5 docs? It'll obviously save us both some time when I create only the _neessary_ pull requests :smiley: .

PR #4271, opened 2018-05-02
L5.5 minimum PHP version 7.0.0 -> 7.1.0

epalmans:
Due to use of typehinted nullable type parameter in Symfony\Component\Translation\Translator constructor

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -11,7 +11,7 @@
 
 ### PHP
 
-Laravel 5.5 requires PHP 7.0.0 or higher.
+Laravel 5.5 requires PHP 7.1.0 or higher.
 
 ### Updating Dependencies
 


taylorotwell:
Link to file?

epalmans:
https://github.com/symfony/translation/blob/master/Translator.php

http://php.net/ChangeLog-7.php#7.1.0

taylorotwell:
That's master branch Symfony. That's not what 5.5 uses.

epalmans:
not directly indeed, but it is used in nesbot/carbon which Laravel (framework) does require from composer

https://github.com/laravel/framework/blob/5.5/composer.json
https://github.com/briannesbitt/Carbon/blob/1.24.1/src/Carbon/Carbon.php

laurencei:
@epalmans - but Carbon allows the use of

`"symfony/translation": "~2.6 || ~3.0 || ~4.0"`

So one of those versions, which support PHP 7.0 - will work?

epalmans:
the nullable types seem to have been introduced in this commit (diffed against master): https://github.com/symfony/translation/commit/6826748b10976339b62b04e9df00eacad95dc5ee#diff-c93f36c15d5c29291cda962db092b47b
used on their 4.0 branch. Therefore it would minimally require PHP 7.1

PS. ~2.6 and ~3.0 doesn't seem to use this, so I guess PHP 7.0 is okay for those

PR #4272, opened 2018-05-02
Clarify that `this->option()` can throw with unexpected arguments

andjd:
Hi,

I updated the documentation for what the `argument` and `option` methods do with an invalid name.  We ran into subtle bugs because we invoked `$this->option("debug")` in a trait we used in a number of commands.  From the documentation, we expected that this would just return null if a command didn't have a `debug` option, but instead it threw an error. (Since the function just delegates to the underlying Symphony Input, is the expected behavior, at least as the code is currently written.)
This PR adds a clarification to the docs to clearly explain this behavior.

### DIFF
--- a/artisan.md
+++ b/artisan.md
@@ -284,7 +284,7 @@ Options may be retrieved just as easily as arguments using the `option` method. 
     // Retrieve all options...
     $options = $this->options();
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not sure this needs to be documented.

Issue #4262, opened 2018-04-28
Missing 'weeklyOn' documentation


Binternet:
On the `Task Scheduling` section there's no reference to `weeklyOn` method though it exists.

PR #4275, opened 2018-05-04
add weeklyOn event frequency

chapeupreto:
This PR closes issue #4262


michaeldyrynda:
Every week on the 4th what? Is it day of week? Does the index start at 0 or 1? Sunday or Monday?

laurencei:
@chapeupreto - I agree with @michaeldyrynda - i think the wording can be a bit clearer.

PR #4284, opened 2018-05-10
Fix URL::temporarySignedRoute usage example

macku99:
Fix URL::temporarySignedRoute usage example to match method signature


macku99:
my bad Taylor... I should have done a composer update... :)

PR #4285, opened 2018-05-11
Mention if-isset blade shorthand code

cweiske:


taylorotwell:
We have deprecated that feature because it is not needed now that PHP has the ?? operator.

cweiske:
This could be mentioned in the docs as well.

PR #4289, opened 2018-05-13
Mention null coalescing operator in blade docs

cweiske:
Followup of https://github.com/laravel/docs/pull/4285


taylorotwell:
We don't need to document standard PHP features.

cweiske:
But having this in the docs would make it easier for people searching for that feature. Not all know about `??` yet.

PR #4292, opened 2018-05-15
Fix wrong example of envoy servers declaration

niahoo:
The current example throws an array to string conversion error, but everything works fine with this change. So I guess it is a doc error.

I know this is a very small change for a pull request, so maybe there is some place to report such errors ?

Thank you.


niahoo:
nah sorry I think I'm stupid.

PR #4303, opened 2018-05-24
Make it clear that Policies, once registered, need to be invoked via a Gate

susanBuck:
For someone new to Gates/Policies and reading these instructions, it's not clear that Policies need to be invoked via a Gate. 
This portion of the docs make it seem like you only need to register the Policy and it will work when used:

>>  "Registering a policy will instruct Laravel which policy to utilize when authorizing actions against a given model"

My edit suggests adding  `Gate::resource('posts', PostPolicy::class);` as an example usage of the registered Policy.

An alternative would be to link back to earlier in the notes where you address "Class@method style callback strings".

### DIFF
--- a/authorization.md
+++ b/authorization.md
@@ -176,8 +176,9 @@ Once the policy exists, it needs to be registered. The `AuthServiceProvider` inc
         public function boot()
         {
             $this->registerPolicies();
-
-            //
+            
+            // Define a gate that uses this policy
+            Gate::resource('posts', PostPolicy::class);
         }
     }
 


taylorotwell:
Hm, yeah I think @lagbox is correct here. This could make things pretty confusing.

susanBuck:
Ok, I realize my mistake now.

I was building an edge case where if a `Post` was of type `demo` then anyone should be able to see it, including the owner, non-owners, **and** non-logged in users.

So my policy method was something like:

```php
public function view(User $user, Post $post)
{
    return $post->slug === config('app.demoSlug') || $user->id === $post->user_id;
}
```

My mistake was in assuming the Authorization features would be triggered for non-logged in guests.  As a result, when testing as a guest, I was not seeing `view` triggered.

I realize now it's only triggered for logged-in users, and in hindsight this is clear based on the description in the docs:

>> Laravel also provides a simple way to authorize **user actions** against a given resource. 

Thanks for considering the pull request despite my misunderstanding.

PR #4305, opened 2018-05-27
update from 5.2.45 to 5.3

skecskes:
I was getting the following error:

    [Symfony\Component\Debug\Exception\FatalThrowableError]         
    Call to undefined method Illuminate\Bus\Dispatcher::mapUsing()  

I had to do remove these providers in order to upgrade to 5.3


taylorotwell:
That's not the correct solution to fixing that.

PR #4306, opened 2018-05-28
Typo line 15

mrluissan:
I think :thinking:  there should be an "if" in:
Laravel includes a middleware that verifies (if) the user of your application is authenticated.


taylorotwell:
I'm not sure it's needed.

PR #4307, opened 2018-05-29
make sure that status code can also be provided

chapeupreto:
like `view()` and `response()`, keep the same idea that status code can also be provided


taylorotwell:
I feel like these examples make it look like a required parameter now.

PR #4312, opened 2018-06-05
Remove `php artisan migrate --reset` as it doesn't exist in Laravel 5.2

mikefrancis:


mikefrancis:
@taylorotwell This is my bad - it does exist, I have no idea why but I am trying to run it with `--reset` instead of `migrate:reset`. This can be reverted. Sorry!

Issue #4302, opened 2018-05-24
[Suggestion] extended info about Blade extends and sections


bertheyman:
A colleague of mine recently asked about some Blade section and I noticed I didn't fully understand how these are working as well.

I'm wondering about these in particular:

1. To end a section: `@show` VS `@endsection` VS `@stop` VS `@apply` 
2. In [the example in the docs](https://laravel.com/docs/5.6/blade#template-inheritance), a sidebar is extended. What if one would like to remove or replace this sidebar completely in one view?
-> Remove the `@parent`, but this would still cause the full sidebar to be rendered first. (and thus causing trouble if data is missing) If I understood correctly, using `@endsection` in the layout instead of `@show` might stop this. However, will all child views now be required to use an `@show` to render the sidebar?

I'd be very grateful if anyone sees a way of improving the docs on this one. Also willing to write myself when someone provides a bit of explanation first.

laurencei:
@bertheyman - please feel free to go ahead and submit the PR for this, and Taylor will be able to look at it and deicde.

bertheyman:
Sure, will do!

PR #4320, opened 2018-06-07
Add docs about the @append Blade directive

bertheyman:
Prevent confusion about default Blade directives, see [this issue](https://github.com/laravel/docs/issues/4302).

An alternative might be to add a small section about the different Blade directives (endsection, show, append, yield) and also mention `@stop` as alias of `@endsection`. 
Not sure about this as using `@stop` might also be discouraged, so let me know if you'd like me to write it this way.


taylorotwell:
I feel like this doesn't really clarify much for me?

meduzen:
@bertheyman Thanks, it clarifies `@append` for me.

PR #4321, opened 2018-06-08
Update horizon.md

ihsanberahim:
Better mention that Horizon needs php function proc_open enabled.


taylorotwell:
I prefer not to have two red call-outs back to back generally. Can we list this in a separate prerequisites section?

ihsanberahim:
@taylorotwell i agree without red call-out and move to different place.

PR #4324, opened 2018-06-10
Add documentation for ignoring a given model in the unique rule.

Alymosul:


taylorotwell:
We already show an example of ignoring, no need to clutter I don't think.

PR #4327, opened 2018-06-11
Remove 2nd param from Event::listen callback

fitztrev:
In a listener like the example provided (a non-wildcard listener), a 2nd parameter is not passed to the closure.


fitztrev:
Nevermind, it could be if you pass an array when firing the event.

```php
event('event.name', ['foo', 'bar']);
```

PR #4334, opened 2018-06-18
Add docs about Event::fakeFor method

arjanwestdorp:
Add documentation about the `Event::fakeFor` method and that it can be used to fake the events for only a part of your test.


taylorotwell:
Needs proper documentation, not just a sentence, etc.

PR #4338, opened 2018-06-19
Remove pseudo-scientific slogan

balping:
There is no empirical evidence for the existence of learning types.

Brief summary: https://en.wikipedia.org/wiki/Learning_styles#Criticism
APS paper: https://doi.org/10.1111/j.1539-6053.2009.01038.x

This question is present in branches 5.4, 5.5, 5.6, master.

I was thinking of a better alternative, but couldn't come up with anything useful. Overall I think this little advert is fine without an introductory question.

### DIFF
--- a/installation.md
+++ b/installation.md
@@ -10,7 +10,7 @@
 <a name="installation"></a>
 ## Installation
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Super pedantic but whatever.

PR #4339, opened 2018-06-20
Update queues.md

nkiermaier:
This adds documentation for the $this->queue constructor method.


taylorotwell:
I'm not sure this is really needs to be documented in this way.

nkiermaier:
I think this could be really useful to people especially if you have different queues in SQS for different jobs.  There's one stack overflow article that talks about it and that's the only resource I could find.

browner12:
I think something like this may be better served through the API documentation.

https://laravel.com/api/5.6/Illuminate/Queue/Jobs/Job.html

PR #4341, opened 2018-06-20
Add information about same rule in confirmed

kudashevs:
This is about [PR](https://github.com/laravel/framework/pull/24633) and confirmed rule. I propose add the information about alternative way of checking attribute name part in Confirmed operation by same rule.


taylorotwell:
It's already documented.

PR #4351, opened 2018-06-28
Mention unsigned columns for foreign keys

drmmr763:
Ran into an issue where I had just created an `integer` column for my foreign key and it took me a bit of work to figure out why it wasn't working. Hopefully this will save someone else a few minutes. 

Also, not sure if this is mysql specific, if so maybe this isn't relevant for everyone.

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -440,6 +440,8 @@ If you pass an array of columns into a method that drops indexes, the convention
 
 Laravel also provides support for creating foreign key constraints, which are used to force referential integrity at the database level. For example, let's define a `user_id` column on the `posts` table that references the `id` column on a `users` table:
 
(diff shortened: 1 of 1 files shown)


enderandpeter:
I'm not sure if the docs are the best place to mention this because it is indeed a [MySQL-specific feature](https://dev.mysql.com/doc/refman/5.6/en/create-table-foreign-keys.html):

> Corresponding columns in the foreign key and the referenced key must have similar data types. **The size and sign of integer types must be the same.** The length of string types need not be the same. For nonbinary (character) string columns, the character set and collation must be the same.

drmmr763:
Hi @enderandpeter thanks for the response. If there's a better vender-specific place to add this sort of info I'll totally add it there instead.

From what I can tell the docs do seem so have some some inconsistency towards vender-related specific documentation (for example, `pusher`, `redis`, `supervisor`, etc...).

Totally get the point; but since this cost me some time I figured it was a worthwhile mention.

taylorotwell:
Not sure I really want to get into documenting MySQL behavior here. Tends to start to bloat things.

PR #4355, opened 2018-07-02
Add validation section detailing upload failure errors

garygreen:
Came across this today and wasn't sure why when I had this rule: 

```php
'image' => 'image|max:2048', // 6MB.
```

And custom error:

```php
'image.max' => 'The cover image cannot be any larger than 6MB.'
```

Why Laravel was outputting upload failed error message - seems Laravel bails validating if PHP failed to upload the file due to size restrictions etc, so worth noting this in the docs as the behaviour is hidden a bit under the hood.

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -18,6 +18,7 @@
 - [Working With Error Messages](#working-with-error-messages)
     - [Custom Error Messages](#custom-error-messages)
 - [Available Validation Rules](#available-validation-rules)
+- [Upload Failures](#upload-failures)
 - [Conditionally Adding Rules](#conditionally-adding-rules)
 - [Validating Arrays](#validating-arrays)
 - [Custom Validation Rules](#custom-validation-rules)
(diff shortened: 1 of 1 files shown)


bozhkos:
Read this 
https://laravel.com/docs/5.6/validation#rule-size

taylorotwell:
Can you just add this to the docs on the `max` rule instead of inventing a whole new section? Thanks

garygreen:
Thing is, it's not specifically a problem for `max` it's applicable to [lots of rules](https://github.com/laravel/framework/blob/5.6/src/Illuminate/Validation/Validator.php#L147) including `mimes`, `dimensions`, `min`, `between`, etc. I thought about adding a note for each rule but figured it'd make more sense to have it under it's own section about validating uploads (not exactly an uncommon thing to handle in apps).

I guess it would mainly fail if the `size` rule errored - though potentially if you've run out of disk space on the server I believe php would also fail as well with the same error? I dunno. Still think it would make sense to have it in it's own section as it's handled completely separately from all the rules. I'll leave the PR and you can add it as you see fit. 😄

PR #4357, opened 2018-07-04
improve documentation on queues

rusinov-artem:
in my project there was an error without this  because php would not
been able to finde .env file of my project and get to use default settings.


taylorotwell:
Hmm, I haven't needed this option before and Forge does not use when writing Supervisor configuration files and we don't have any issues there.

PR #4358, opened 2018-07-05
add redirects.md page in menu

Ellrion:


taylorotwell:
They are covered under the Responses documentation. That is an old page.

PR #4364, opened 2018-07-06
Added a warning to resetting database after each test

MidasLamb:
Added a warning to make clear that resetting the database will use migrations so tables not created through migrations will be deleted. This may happen early in the development cycle or with people that are getting into Laravel.

### DIFF
--- a/database-testing.md
+++ b/database-testing.md
@@ -46,6 +46,8 @@ The `--model` option may be used to indicate the name of the model created by th
 <a name="resetting-the-database-after-each-test"></a>
 ## Resetting The Database After Each Test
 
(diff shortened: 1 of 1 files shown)


michaeldyrynda:
The added note isn’t entirely accurate.

`RefreshDatabase` doesn’t use migrations to reset the database, it drops all the tables in your default connection at the start of a test run, runs the migrations to build the schema, then runs all of your tests in transactions. 

That said, I think “resets your database” succinctly covers what is happening here in any case.

MidasLamb:
I wrongly thought it meant that it would reset the database to the point before the tests, as in, if the database is populated with some records, they would remain, but any action done by the test would appear to have never happened. I thought others might make the same assumption as me, so a clarification might help.

PR #4367, opened 2018-07-09
Update assertSessionHasAll

sumityadav:
The `assertSessionHasAll` doesn't match the signature. The method accepts `array` instead of key/ value.


sumityadav:
The same issue exists for 5.6 too. Do you want me to create another pull request for 5.6?

PR #4370, opened 2018-07-10
Added info about Valet community forks.

amadeann:
For a long time, I assumed that Valet was only something for Mac users, until one day I saw someone mentioning Valet forks in comments on Laracasts. I think plenty of other people may also be missing out on Valet, as the very first sentence tells you it's only for Mac. I haven't tried personally the one for Windows, but Linux fork works great in my experience.


taylorotwell:
No plans to add this since there could be other competing forks as well.

PR #4371, opened 2018-07-11
document the --delay option for queue:work

sebsel:
This might also be useful in the Horizon docs, but it belongs here. :)


taylorotwell:
You can use `php artisan help queue:work` to see the full command documentation and options.

PR #4373, opened 2018-07-11
Correct count field name

richplane:
"relation" here is ambiguous. The related count field name isn't {relationship}_count, or even {related_model}_count, it seems to be {related_table_name}_count. In my case according to the docs it should have been productVariants_count, but it's actually product_variants_count.
Am I missing some subtlety here?


taylorotwell:
Hmm, I see it added as the relationship name:

image

Maybe we need a test for this on laravel/framework to prove it one way or the other?

richplane:
ah - so it's the snake case version of the relationship name?
In which case perhaps the document could be adjusted to say that rather than "{relation}_name"

PR #4374, opened 2018-07-12
[5.2] Update upgrade.md

### LINKED ISSUE #4372: [5.3] Missing authorization changes in upgrade path documentation
When upgrading my Laravel application today which was on version 5.2 I completely broke my authorization and it wasn't in the upgrade documentation.

Commit that changed the documentation: e2cbae5

The feature that was removed was specifically this:

---

#### Automatically Determining Policy Methods

Frequently, a policy's methods will correspond to the methods on a controller. For example, in the `update` method above, the controller method and the policy method share the same name: `update`.

For this reason, Laravel allows you to simply pass the instance arguments to the `authorize` method, and the ability being authorized will automatically be determined based on the name of the calling function. In this example, since `authorize` is called from the controller's `update` method, the `update` method will also be called on the `PostPolicy`:

    /**
     * Update the given post.
     *
     * @param  int  $id
     * @return Response
     */
    public function update($id)
    {
    	$post = Post::findOrFail($id);

    	$this->authorize($post);

    	// Update Post...
    }

---

How it broke my application is that I was using `$this->authorize($record);` in a `show` method on the controller so the policy expects a `show` method in 5.2 but in 5.3+ it seems to be looking for a `view` method. I had similar occurrences all over my other controllers and other method names.

Should this change be added to the 5.3 upgrade documentation?

ryanmortier:
Resolves #4372.

If a developer at the time (like in my case) had originally built their app with `5.1 LTS` and then later upgraded their app to `5.2` using this documentation, they will miss the change that happened to the `app/Http/Controllers/Controller.php` file which this PR addresses. Specifically, adding the `AuthorizesResources` trait.

If the developer was using the **Automatically Determining Policy Methods** feature, then they'll encounter unexpected results. Specifically, when authorizing a method name that is in the missing resource map located in the missing trait, e.g. `show`. Without the resource map authorization looks for a `show` method on the policy. With the resource map it would look for a `view` method on the policy. In `5.3` the `AuthorizesResources` and `AuthorizesRequests` traits are merged which breaks apps that missed the addition of this trait in `5.2` because they may have developed policies while missing a resource map.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -53,6 +53,25 @@ The `loginPath()` method has been removed from `Illuminate\Foundation\Auth\Authe
 
 The `Illuminate\Auth\Access\UnauthorizedException` has been renamed to `Illuminate\Auth\Access\AuthorizationException`. This is unlikely to affect your application if you are not manually catching this exception.
 
+#### The `AuthorizesResources` Trait
+
+You should add the `AuthorizesResources` trait to your `app/Http/Controllers/Controller.php` file.
+
+    <?php
+
+    namespace App\Http\Controllers;
+
+    use Illuminate\Foundation\Bus\DispatchesJobs;
+    use Illuminate\Routing\Controller as BaseController;
+    use Illuminate\Foundation\Validation\ValidatesRequests;
+    use Illuminate\Foundation\Auth\Access\AuthorizesRequests;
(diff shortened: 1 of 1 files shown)


taylorotwell:
This is optional.

PR #4375, opened 2018-07-12
Improve docs related to storage testing

introwit:
- The current definition of `Storage` facade's `fake` method in the docs is a bit different compared to what the code actually does and that method's description in the doc block itself. This PR aims to synchronise that.
- Also, in the context of testing web requests based logic, there is a bit of guessing and confusion left from the given 'API calls' based examples, which is seen in https://github.com/laravel/framework/issues/20221 and many other places such as laracasts forum, etc. This PR aims to fill that void.

### DIFF
--- a/http-tests.md
+++ b/http-tests.md
@@ -164,7 +164,7 @@ If you would like to verify that the given array is an **exact** match for the J
 <a name="testing-file-uploads"></a>
 ## Testing File Uploads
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
All the wording doesn't really match the current style of the documentation. Please try to slim down the text a bit more.

taylorotwell:
I've updated the example in the docs.

introwit:
Thanks @taylorotwell ♥️🙂

PR #4380, opened 2018-07-13
Cron entry example for schedule:run relative to project path

tylermann:
Ran into an issue where running the cron with a similar format to the existing example caused scheduled console jobs to not work if they were scheduled commands because the call to `Application::formatCommandString()` within `Schedule::command()` appears to generate the command string assuming that the caller is already in the working directory of the project. eg: `php artisan some:command`.

### DIFF
--- a/scheduling.md
+++ b/scheduling.md
@@ -24,7 +24,7 @@ Laravel's command scheduler allows you to fluently and expressively define your 
 
 When using the scheduler, you only need to add the following Cron entry to your server. If you do not know how to add Cron entries to your server, consider using a service such as [Laravel Forge](https://forge.laravel.com) which can manage the Cron entries for you:
 
-    * * * * * php /path-to-your-project/artisan schedule:run >> /dev/null 2>&1
+    * * * * * cd /path-to-your-project && php artisan schedule:run >> /dev/null 2>&1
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can you give me an example of a scheduled command that breaks with the existing documentation example so I can try to recreate it?

tylermann:
Yes, an example of a failing command would look something like this:
```
$schedule->command(\App\Sync\Console\Commands\SyncData::class)
            ->Dailyat('9:00')
            ->onOneServer();
```

Where the command name would be something like `sync:data` and running `php artisan sync:data` on the command line from within the project folder would work correctly.

I was just doing some debugging with this and it actually doesn't appear to be an issue in the homestead local environment but only on my staging/production servers (one is ubuntu and one is amazon linux) . Thing that I *thought* was the issue is illustrated in this code sample:
```
vagrant@homestead:~$ php ~/Code/app/artisan tinker
>>> use Illuminate\Console\Application;
>>> Application::formatCommandString('sync:data')
=> "'/usr/bin/php7.2' 'artisan' sync:data"
```

Where `Application::formatCommandString` seems to expect the artisan path to be in the current working directory. However maybe it is just something odd about my setup/configuration and not a problem for everyone. Happy to close the PR if you don't think this is a widespread issue.

taylorotwell:
Making this change seems pretty fine I guess.

PR #4383, opened 2018-07-14
[5.6] Improve eager loading warning

staudenmeir:
The warning should be more general and not only apply to `BelongsTo` relationships.


taylorotwell:
Hmm, I think we would need a more fleshed out explanation here. This could still be confusing.

PR #4387, opened 2018-07-20
Update doc of custom pivot models and extra attributes

aonurdemir:
Fixed missing information explaning that extra attributes should be provided by withPivot method regardless a custom pivot table is used.

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -379,6 +379,10 @@ If you would like to define a custom model to represent the intermediate table o
         }
     }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
This is already documented.

aonurdemir:
In the doc, it seems like there is no need to write withPivot when a custom model is implemented since the table name is defined in the model and attributes could be resolved automatically like any other ordinary model (This is my interpretation of the doc when I read). Additionally, the custom model example also does not contain withPivot method. That was confused me.  It is understood as if it is enough to use just using() method when implementing custom model. 

I would like to give some verbose example. Thanks for your evaluation.

PR #4391, opened 2018-07-24
Update comments for `belongsToMany` relationship

mislam:
The comments for `belongsToMany` were written such a way that they sound like `hasMany` relationship. I've updated them with the correct terms.


taylorotwell:
I think it's pretty OK. 🤷‍♂️

PR #4393, opened 2018-07-24
[5.5] Add note for using withCount together with select

mnabialek:
As showed in https://github.com/laravel/framework/issues/24797 there is a problem when `withCount` is used before `select`. So as suggested I've added this as note to documentation.


taylorotwell:
Wrong version.

PR #4403, opened 2018-08-02
[5.6] Adding example to schedule calls uisng classes

isocroft:
This PR simply adds an example to schedule calls using _Invokable_ classes which act as `callables` when instanstiated as they implement the `__invoke` method.


taylorotwell:
This is a lot of code for such a simple example. Maybe we could just mention in one sentence that Closures or invokable objects are acceptable?

isocroft:
Okay , i guess we could do that... Let me update the PR

PR #4404, opened 2018-08-02
[5.6] Add note for using withCount together with select

mnabialek:
I target now 5.6 however 5.5 is LTS and probably this note should be also for 5.5 - https://github.com/laravel/docs/pull/4393


taylorotwell:
I will have to re-write this with wording that matches our documentation.

PR #4414, opened 2018-08-06
Expand new pivot code example

isocroft:
This PR is put forward to clarify the method of overriding the `newPivot` method when creating a custom pivot table model. There are still a lot of **Laravel 4** example online that show a the old way of doing it. This has bitten a lot of developers i know really hard when surfing the internet for a **Laravel 5** example of how to correctly do this. I hope this PR serves to give developers a one-stop destination for a very lucid example right in the **Laravel 5* docs.

```php

// Old Way (Laravel 4)

public function newPivot(Model $parent, array $attributes, $table, $exists, $using = null){

			if($parent instanceof User){
                                // this thorws an error in L5 and righfully so
				return new Profile($parent, $attributes, $table, $exists); 
			}
			
                        return parent::newPivot($parent, $attributes, $table, $exists, $using);
}
```

```php

// New Way (Laravel 5)

public function newPivot(Model $parent, array $attributes, $table, $exists, $using = null){
			
                        if($parent instanceof User){
				return Profile::fromRawAttributes($parent, $attributes, $table, $exists);
			}

			return parent::newPivot($parent, $attributes, $table, $exists, $using);
}
```

Thanks

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -382,11 +382,11 @@ If you would like to define a custom model to represent the intermediate table o
         {
             if ($parent instanceof User) {
 
-                return \App\UserRole::fromRawAttributes($parent, $attributes, $table, $exists);
-			}
-            
-			return parent::newPivot($parent, $attributes, $table, $exists, $using);
-		}
+		return \App\UserRole::fromRawAttributes($parent, $attributes, $table, $exists);
+    	    }
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
Spacing is wrong. Also missing doc blocks. Please re-submit with those fixes.

PR #4416, opened 2018-08-08
Update redis.md - note about dispatching

jameshulse:
With some versions of reids you can run into an error along the lines of "-ERR only (P)SUBSCRIBE / (P)UNSUBSCRIBE / PING / QUIT allowed in this context" if you call `SomeJob::dispatch()` from within a `Redis::subscribe(...)` handler and if they are both running on the same redis connection. The subscribe connection can only be used for pubsub.

To fix this issue I introduced a new connection in `database.php` for redis pubsub and called `Redis::connection('pubsub')->subscribe(...)`

### DIFF
--- a/redis.md
+++ b/redis.md
@@ -214,6 +214,8 @@ First, let's setup a channel listener using the `subscribe` method. We'll place 
             });
         }
     }
+    
+> {note} With some versions of Redis you will not be able to dispatch jobs on the same connection that you are subscribed on. You will need to either subscribe using a new redis connection or dispatch on a different connection.
 
 Now we may publish messages to the channel using the `publish` method:
 


taylorotwell:
It's a vague comment. Why on some versions? What causes it?

jameshulse:
Fair. From the [redis pub/sub docs](https://redis.io/topics/pubsub):
```
A client subscribed to one or more channels should not issue commands, although it can subscribe and unsubscribe to and from other channels. The replies to subscription and unsubscription operations are sent in the form of messages, so that the client can just read a coherent stream of messages where the first element indicates the type of message. The commands that are allowed in the context of a subscribed client are SUBSCRIBE, PSUBSCRIBE, UNSUBSCRIBE, PUNSUBSCRIBE, PING and QUIT.
```

In the code example:

```php
Redis::subscribe('channel-name', function () {
     SomeJob::dispatch();
});
```

If the configured queue driver is redis then the same client is used for the dispatch as was used in the subscribe and an error is thrown: `ERR only (P)SUBSCRIBE / (P)UNSUBSCRIBE / PING / QUIT allowed in this context`.

My comment was to try and document this caveat as I ran into the issue - but maybe there is a way to fix the issue in Laravel itslef by creating a new client within the callback method.

PR #4417, opened 2018-08-08
Document the ability to override the app env when running a command

X-Coder264:


michaeldyrynda:
Not sure it's necessary to document something that is available with the `php artisan help` command, where do you draw the line otherwise?

X-Coder264:
Well, I didn't even know that's there as I don't even remember when I ran that command last. I stumbled upon this by accident when I was going through some framework code so I decided to write a sentence about it in the documentation. I'm probably not the only one who hasn't seen that so I think it can't hurt if we have a sentence about it in the documentation as well :)

taylorotwell:
This is already documented in the Artisan CLI help screens as noted.

PR #4423, opened 2018-08-13
Add withoutCookieSerialization as of 6.0.7.

yajra:
As per upgrade guide https://laravel.com/docs/5.6/upgrade#upgrade-5.6.30.


taylorotwell:
I've released Passport 7.0 so this won't be necessary on new installations.

PR #4425, opened 2018-08-13
Add documentation about atomic locks

tonysm:
I was looking for documentation about it, could only find the [tweet](https://twitter.com/taylorotwell/status/875095973629919234). Copied over all examples from the tweet, and tweaked the message based on commit messages and other places where distributed locks are mentioned (task scheduling on one server).

Not sure if all the examples are still true, worth checking.


taylorotwell:
Thanks

PR #4440, opened 2018-08-21
Update VerifyCsrfToken for cookie serialization

khepin:
The `Illuminate\Foundation\Http\Middleware\VerifyCsrfToken` class directly refers to `Illuminate\Cookie\Middleware\EncryptCookies::serialized('XSRF-TOKEN')` to decide whether it should decrypt and unserialize the cookie or only decrypt.

If we only update `App\Http\Middleware\EncryptCookies` with `protected static $serialized = true`, that `true` value would never seen from the CSRF token verification point of view.

Without this and only following the upgrade guide all we would get is `HTTP 419` errors.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -29,6 +29,10 @@ Since this vulnerability is not able to be exploited without access to your appl
 
 To make sure CSRF token validation still works, you will need to have the `App\Http\Middleware\VerifyCsrfToken` [middleware](https://github.com/laravel/laravel/blob/master/app/Http/Middleware/VerifyCsrfToken.php) also reference the correct value of `EncryptCookies::$serialize`:
 
+    use App\Http\Middleware\EncryptCookies;
+    
+    // ...
+    
     /**
      *
      * @return string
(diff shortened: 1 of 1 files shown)


taylorotwell:
But this is already included in the base VerifyCsrfToken class?

taylorotwell:
Also, the EncryptCookies middleware just return the value of $serialize. So I don't see the problem?

khepin:
Hi, the issue is in the following diagram:

```
CustomCsrfVerifier::serialized()         CustomEncryptCookies::serialized()
            +                               => true
            |                                     +
            |                                     |
            |                                     |
            |                                     |
            |                                     |
            v                                     v
BaseCSRFVerifier::serialized()    +--->  BaseEncryptCookies::serialized()        
                                            => false
```

Since the `BaseCSRFVerifier` refers directly to the `BaseEncryptCookies` by a "hardcoded" class name, it will never see the modified value of `$serialized` since that one has been set in the `CustomEncryptCookies` class.

I used `Base` and `Custom` to make the names shorter, but `Base` are the classes in `Illuminate\Http\Middleware` and `Custom` the ones in `App\Http\Middleware`.

PR #3529, opened 2017-08-18
Add Redirects to menu

bobbybouwmann:

PR #4443, opened 2018-08-23
[5.7] add redirects

Nationalcat:


lagbox:
Seems people have tried to add that to the menu before, but I am not sure the current status of that though.

Previous PRs:
#3626 #3531 #3530 #3529

PR #4444, opened 2018-08-23
[5.6] Add Gitlab CI/CD example to dusk documentation

dinandmentink:
This PR will add an example for Gitlab CI/CD to the Dusk documentation. Given that Gitlab CI/CD is free for small proprietary projects I believe this example to be relevant to other Laravel enthousiasts.

This example is expected to also work for Laravel 5.7. Should I open a seperate PR against the master branch?

### DIFF
--- a/dusk.md
+++ b/dusk.md
@@ -36,6 +36,7 @@
     - [Codeship](#running-tests-on-codeship)
     - [Heroku CI](#running-tests-on-heroku-ci)
     - [Travis CI](#running-tests-on-travis-ci)
+    - [Gitlab CI/CD](#running-tests-on-gitlab-ci-cd)
 
 <a name="introduction"></a>
 ## Introduction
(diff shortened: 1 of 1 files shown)


dusk.md
```diff
@@ -1317,3 +1318,25 @@ To run your Dusk tests on Travis CI, we will need to use the "sudo-enabled" Ubun
...
     script:
        - php artisan dusk
+
+<a name="running-tests-on-gitlab-ci-cd"></a>
+### Gitlab 
+
+Running your Dusk tests on Gitlab's CI/CD pipelines works easiest by explicitly running the chromedriver executable as well as PHPs development server using `php artisan serve`. See the following example `.gitlab-ci.yml` file.
+
+    Tests:
+      image: chilio/laravel-dusk-ci:stable
```
Kyslik: I think it should be image maintained by Laravel.

dinandmentink: Yes, that would be better. 

Is there an image maintained by laravel? Or should I just remove the line because it is non-laravel-related and Gitlab specific?

The set of commands is whats required for specifically dusk to work, the image is up to the developer knowing their way around Gitlab CI/CD.

Kyslik: We just need a confirmation from Laravel maintainers on this. I would leave it there for now.

taylorotwell:
Holding off on this since it's not an image we control.

Kyslik:
@dinandmentink no, since it does not work without it, does it?

PR #4448, opened 2018-08-25
Homestead configuration for running the scheduler

pgrosvenor:


taylorotwell:
This should be covered in Homestead documentation, not here.

PR #4449, opened 2018-08-25
[5.6] Update pagination docs

X-Coder264:
The `render` method is present on the `Illuminate\Contracts\Pagination\Paginator` contract while `links` is not. The `links` method is just an alias to `render` in Laravel's default concrete implementation of the `Paginator` contract so it doesn't make sense to promote the usage of the `links` method in the docs when the appropriate method on the interface for rendering the pagination is actually `render`.


X-Coder264:
@taylorotwell A sentence explaining why was this closed wouldn't have hurt...

PR #4455, opened 2018-08-28
Update mix's versioning/cache busting documentation

Zhastreaus:
Laravel Mix's versioning/cache busting works differently now. Instead of appending hashes to the file, it uses url query.


taylorotwell:
We don't need a new tip. Please remove that.

PR #4457, opened 2018-08-29
Explain defining associations with Models in addition to IDs

jk3us:
It isn't clear from the docs that you can attach/detach/sync many-to-many associations by passing in model objects.  All of the current examples pass IDs.


taylorotwell:
Maybe stick this directly under the ->attach documentation. Not sure it needs its own heading.

PR #4467, opened 2018-09-02
doc add route redirect method

jpuck:
I feel like this might be helpful considering the given route closure examples would break using `route:cache`

Let me know if you would like any edits. or a PR to 5.7.

Thank you for all your work!


X-Coder264:
This is already documented in the [routing docs](https://laravel.com/docs/5.6/routing#redirect-routes).

PR #4478, opened 2018-09-05
[5.7] Mention the resources directory update

MladenJanjetovic:


upgrade.md
```diff
@@ -15,6 +15,10 @@ Update your `laravel/framework` dependency to `5.7.*` in your `composer.json` fi
 
 Of course, don't forget to examine any 3rd party packages consumed by your application and verify you are using the proper version for Laravel 5.7 support.
 
+### Updated resources directory structure
+
+The `resources/assets` directory is removed and it's subdirectories are moved to the `resources`directory. Be sure to update your Laravel Mix configuration.
```
browner12: "it's" should be switched to "its".

also, please add a space between "resources" and "directory".

MladenJanjetovic: done, tnx

taylorotwell:
This isn't relevant to upgrading your application. It's only relevant to new applications.

MladenJanjetovic:
This was the first information I heard about 5.7 from Paul Redmond's blog post. So when I started upgrading my applications today, I didn't know what to do with it. Is new version going to handle the old structure well or I should move all of my assets to meat the upgrade? 

I think it should be at least mentioned in upgrade guide. 
my 2c

brunogaspar:
@MladenJanjetovic You don't need to touch anything on your current application, unless you want to adopt the new structure. If you prefer to adopt the new structure, then, you'll need to touch your mix config file and update the paths, therefore, and as @taylorotwell mentioned, this is not needed in the upgrade guide, since it's not a required step.

PR #4480, opened 2018-09-05
[5.7] Adjust database cache migration

staudenmeir:
Adjust suggested column type to https://github.com/laravel/framework/pull/22091.


taylorotwell:
This feels like something the user can change if they want.

PR #4460, opened 2018-08-31
Multiple versions of PHP now work on Apache

xushangning:
Confirmed to work on my own machine with a fresh install of Laravel 4.2 and PHP 5.6. See [laravel/homesteadv7.11.0](https://github.com/laravel/homestead/releases/tag/v7.11.0) and [v7.14.2](https://github.com/laravel/homestead/releases/tag/v7.14.2) for more details.

PR #4487, opened 2018-09-06
Multiple versions of PHP now work on Apache

xushangning:
This pull request is the same as #4460, except that it only deletes the note which states that multiple versions of PHP only work on Nginx.

Confirmed to work on my own machine with a fresh install of Laravel 4.2 and PHP 5.6. See laravel/homesteadv7.11.0 and v7.14.2 for more details.

I can't find a way to merge this commit into 5.6 and 5.7 without more pull requests, so I just commit to master.


taylorotwell:
Please send to 5.7 branch.

PR #4492, opened 2018-09-07
Clarify `combine` uses values

tomlankhorst:
Original doc suggests that the keys of col. A are combined with vals. of col. B.


taylorotwell:
Use 5.7 branch.

PR #4504, opened 2018-09-12
[5.7] Clarifying the need to manually update dependency to use dump-server

kp-syginteractive:
Adding a line to clarify the need to do a composer update to pull in the package before trying - php artisan dump-server.


michaeldyrynda:
Are we sure we need to clarify how to use Composer? It's assumed knowledge if you're using Laravel.

taylorotwell:
Doesn't make sense if you're using a new Laravel 5.7 applications because it is not necessary.

PR #4505, opened 2018-09-13
Add note about changes to template styles

dwightwatson:
This adds a note about changes to the email templates which broke button styles for those who had customised the default stylesheet. This snagged me on a couple of projects, didn't realise it came from the 5.7 upgrade. Because there was already a heading for the email verification feature I merged these both under an email heading.

This addresses laravel/framework#25597.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -257,7 +257,15 @@ PostgreSQL supports the float values `Infinity`, `-Infinity` and `NaN`. Prior to
 
 As of Laravel 5.7, these values will be cast to the corresponding PHP constants `INF`, `-INF`, and `NAN`.
 
-### Email Verification
+### Email
+
+#### Template Theme
+
+**Likelihood Of Impact: Low**
+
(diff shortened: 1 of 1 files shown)


laurencei:
This should probably go under the existing `Mail` heading?

Also - impact should be at least `medium` as it is a *silent* breaking change. I just got burnt with this today, and came here to document it.

dwightwatson:
Good point, I'll move it under mail and then leave email verification as it's own heading again.

I thought low might be fitting because I don't think everyone would be publishing and customising the theme. This is only an issue if you've gone out of your way to customise it.

To be honest I feel like it's easier to just point out the change and suggest a re-publish and then leave it up to the developer to pick whether they want to cherry-pick the diff or not.

laurencei:
> This is only an issue if you've gone out of your way to customise it.

If you have only publised it, even without changes, this breaks. We'll leave it for Taylor to decide.

Thanks for submitting this.

PR #4516, opened 2018-09-14
remove incorrect version

browner12:
the current LTS is 5.5, not 5.6.

This PR removes the value, and links to the 'releases' page, so we don't have to keep updating it, and people can look if they'd like.

### DIFF
--- a/contributions.md
+++ b/contributions.md
@@ -46,7 +46,7 @@ Informal discussion regarding bugs, new features, and implementation of existing
 <a name="which-branch"></a>
 ## Which Branch?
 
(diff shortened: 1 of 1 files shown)


TBlindaruk:
yes, but after this fix, link was changed to the 5.7 ;)

browner12:
I don't think we want to navigate to the latest version, we want to navigate to the current version of documentation the user is viewing. I don't think the version should change unless the user explicitly selects it in the upper corner.

michaeldyrynda:
When referencing the LTS branch, you do, because on 5.7 that’ll be 5.5. If you’re looking at the 5.7 version of the docs when the next LTS is out (assuming 5.8), then it’ll still say 5.5 is the current LTS.

PR #4523, opened 2018-09-15
Upgrading To 5.7.0 From 5.6

### LINKED ISSUE #4514: Upgrading To 5.7.0 From 5.6
Hi there!

I'm currently trying to upgrade from 5.6 to 5.7 and I found out that configuration env. key for "queue" was changed from "QUEUE_DRIVER to QUEUE_CONNECTION"

It would be great to add this to the upgrade guide. I will be happy to open a pull request.

lubimor:
Hi, 
During an application upgrade, I noticed that some of the environment variables were renamed and not specified in the upgrade guide. I thought It would be great to add a note about these changes to the upgrade guide. resolve #4514 

Cheers

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -380,6 +380,13 @@ The `validate` method [was added to the `Illuminate\Contracts\Validation\Validat
 
 If you are implementing this interface, you should add this method to your implementation.
 
+### Configuration
+
+#### The `QUEUE_DRIVER` environment key
+
+**Likelihood Of Impact: Very Low**
+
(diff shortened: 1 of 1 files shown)


lubimor:
> Hmm, this wouldn't affect existing applications though?

They should not be affected, I think. 
But there can be a case when someone will decide to upgrade application and sync files in an application with files from `laravel/laravel` repository. In this situation, will get a mismatch between the key in production .env and config file.

lubimor:
Great! Thanks a lot!

PR #4526, opened 2018-09-17
Updates eloquent-resources.md

r3oath:
Adds a note about the automatic default behaviour around collections and the singular associated resource class.

### DIFF
--- a/eloquent-resources.md
+++ b/eloquent-resources.md
@@ -126,7 +126,7 @@ After defining your resource collection, it may be returned from a route or cont
         return new UserCollection(User::all());
     });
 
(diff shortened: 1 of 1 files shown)


eloquent-resources.md
```diff
@@ -126,6 +126,8 @@ After defining your resource collection, it may be returned from a route or cont
         return new UserCollection(User::all());
     });
 
+> {note} Collections will automatically utilise any singular resources defined (a resource class matching the start of the collection class name, e.g.: `JokeCollection` will try to match `Joke`.) If you wish to define what singular resource should be used instead of this default behaviour you can define `$this->collects` from within your collection class.
```
browner12: I believe we use American English here, so 'utilise' --> 'utilize'.

r3oath: Sure thing, I'll swap out the Ozzy for American.

taylorotwell:
Hmm, I think it would be better if this included a code example. It's not super clear how to override $this->collects. Thanks!

PR #4535, opened 2018-09-20
Document the ability to specify slack channel

t1sh0o:


t1sh0o:
You could do it in Slack's configs, but this way it allows you to reuse webhooks for different projects and channels. And since it is already implemented in the framework, I thought it is a good idea to be in the docs for reference.

PR #4538, opened 2018-09-21
add note about Blade directive naming

browner12:
PR for issue #4537 

Relevant line in the compiler:

https://github.com/laravel/framework/blob/5.7/src/Illuminate/View/Compilers/BladeCompiler.php#L310

PR #4540, opened 2018-09-21
Blade directives

browner12:
sorry about that. trying #4538 again.


taylorotwell:
You know, on second thought... what if we just check the format of the name in the Blade directive method directly and just throw an informative exception?

browner12:
Great idea, but probably a breaking change? Realistically no one would be using it that way, but technically we are changing behavior.

PR #4546, opened 2018-09-23
Update contributions.md

browner12:
revert #4545.

the point of the link isn't to link to the LTS documentation, it's to link to the releases page, so people can see what the status of each version is.

I think it's a poor and confusing user experience for the links to be switching docs versions on the user.

I've also added the anchor of `#support-policy` this time, so it's more direct.

### DIFF
--- a/contributions.md
+++ b/contributions.md
@@ -46,7 +46,7 @@ Informal discussion regarding bugs, new features, and implementation of existing
 <a name="which-branch"></a>
 ## Which Branch?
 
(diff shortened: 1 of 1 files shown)


TBlindaruk:
@browner12 as for me will be better just keep only 5.5 version like:
`current LTS branch` (5.5). instead
`[current LTS branch](/docs/{{version}}/releases#support-policy)`.


Since now ex like:
- from github to contributing page
- search on contributing page need line ;)
- from contributing page to releases page
- serach in releases page version...

In case if we will have  only release number, than user expiriance will be better:

- go from github to intributiing page
- search on contributing page need line ;)

PR #4558, opened 2018-09-27
[5.7][UPGRADE] Mention new data/ directory for cache

mfn:
E.g. see the discussion in https://github.com/laravel/framework/issues/25451


halaei:
Maybe with this change, it is also required to change cache.php as well:
```php
        'file' => [
            'driver' => 'file',
            'path' => storage_path('framework/cache/data'),
        ],
```

PR #4568, opened 2018-10-01
[5.7] Fix semantics

richardkeep:
Since the action is to create a new post, probably this should mean authorizing access to create form and not the post method for storing.


taylorotwell:
The old documentation was correct.

PR #4577, opened 2018-10-03
update cache documentation on removal of indefinate block

### LINKED ISSUE #4446: [5.6] Cache blockFor() method no longer exists and block() now takes timeout parameter
In the Atomic Locks section of the Cache documentation the block() method is shown as having no parameters and the blockFor() method is shown as being the method to call if you want to block for a certain period of time.

https://laravel.com/docs/5.6/cache#atomic-locks

However, when I tried the blockFor() method I got the following error:

    Call to undefined method Illuminate\Cache\RedisLock::blockFor()

It looks like blockFor() has been removed in favour of adding the timeout parameter to the block() method.

otherpaco:
Documentation needs to updated because of the removal of indefinate block as per laravel/framework@045e6f2 (fixes #4446)

### DIFF
--- a/cache.md
+++ b/cache.md
@@ -220,19 +220,13 @@ The `get` method also accepts a Closure. After the Closure is executed, Laravel 
         // Lock acquired indefinitely and automatically released...
     });
 
-If the lock is not available at the moment you request it, you may instruct Laravel to wait until it becomes available:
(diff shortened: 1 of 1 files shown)


taylorotwell:
Ah, I just did this on 5.7 manually. Sorry didn't see this!

PR #4579, opened 2018-10-03
[5.7] PSR-11 container

deleugpn:
Docs for https://github.com/laravel/framework/pull/25870 in case we choose this path.

### DIFF
--- a/container.md
+++ b/container.md
@@ -276,4 +276,4 @@ Laravel's service container implements the [PSR-11](https://github.com/php-fig/f
         //
     });
 
-If the identifier cannot be resolve an exception is thrown. The exception will be `Psr\Container\NotFoundExceptionInterface` if the identifier was never bound or a `Psr\Container\ContainerExceptionInterface` otherwise.
+If the identifier cannot be resolved an exception is thrown. The exception will be `Psr\Container\NotFoundExceptionInterface` if the identifier was never bound or a `Psr\Container\ContainerExceptionInterface` otherwise.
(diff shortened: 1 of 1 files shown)


container.md
```diff
@@ -276,4 +276,4 @@ Laravel's service container implements the [PSR-11](https://github.com/php-fig/f
         //
     });
 
-> {note} Calling the `get` method will throw an exception if the identifier has not been explicitly bound into the container.
+If the identifier cannot be resolved an exception is thrown. The exception will be `Psr\Container\NotFoundExceptionInterface` if the identifier was never bound or a `Psr\Container\ContainerExceptionInterface` otherwise.
```
svenluijten: I feel like this needs some rewording. What do you think of this?

> If the identifier cannot be resolved, a `Psr\Container\NotFoundExceptionInterface` is thrown. If the identifier was never bound, a `Psr\Container\ContainerExceptionInterface` is thrown instead.

deleugpn: When you say `Otherwise` after `If the identifier cannot be resolved`, I imagine it's talking about `if the identifier can be resolved`, which is actually false. Furthermore, stating that a `Psr\Container\NotfoundExceptionInterface` is thrown if the identifier cannot be resolved is false because it depends on whether the identifier was bound or not. The second part of the statement is also wrong as `Psr\Container\ContainerExceptionInterface` is thrown if the identifier was bound and not resolved.

svenluijten: Ah, I completely misread the original text! Ignore me 😄

PR #4601, opened 2018-10-05
[5.7] Document Scout search callbacks

driesvints:
This will make it more clear how developers can manipulate the searching behavior. Think it's good to have this in the docs as most people didn't seem to know this was possible. The example might need some tweaking.

### DIFF
--- a/scout.md
+++ b/scout.md
@@ -334,15 +334,17 @@ When this configuration option is `true`, Scout will not remove soft deleted mod
 <a name="search-callbacks"></a>
 ### Search Callbacks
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
It's a little unclear what the arguments being passed to the callback are... For example what is $engine and $query etc... we probably should document that.

driesvints:
@taylorotwell thanks for noting this! It turns out I misunderstood the example from someone else myself. I did some further digging and I think the updates to the example are now more clear. Let me know if you want further adjustments.

PR #4602, opened 2018-10-05
[5.7] Document how to simulate executing a queued job

bryanjamesmiller:


taylorotwell:
It's unclear what this is actually doing or means? We would probably need its own paragraph explaining it. Even I'm not sure what is going on. 🤷‍♂️

PR #4603, opened 2018-10-06
Request data can be retrieved via helper function

gauravmak:


taylorotwell:
Already documented under helper documentation.

PR #4608, opened 2018-10-08
[5.7] Clarify foreign key constraint usage for SQlite

dakira:
When using SQLite enabling foreign key constraints in a migration has no
effect.

I also fixed an unescaped underscore, hope that's okay. :)

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -452,7 +452,7 @@ You may also specify the desired action for the "on delete" and "on update" prop
           ->references('id')->on('users')
           ->onDelete('cascade');
 
-To drop a foreign key, you may use the `dropForeign` method. Foreign key constraints use the same naming convention as indexes. So, we will concatenate the table name and the columns in the constraint then suffix the name with "_foreign":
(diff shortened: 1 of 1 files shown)


taylorotwell:
It seems weird to make people do this?

dakira:
@taylorotwell I agree 100%. In my opinion foreign key constraints should be enabled by default for SQLite. People who define constraints most probably want them enforced. If they never defined them, it doesn't make a difference if foreign key constraints are enabled.

If would make this configurable via a `database.connections.sqlite.foreign_keys` config key and enable it by default.

I'll gladly submit a PR for this and change this PR to a note that mentions the difference for sqlite.

SjorsO:
Related: https://github.com/laravel/ideas/issues/1254

driesvints:
Nice, I didn't know about this! I agree that we might perhaps better use a config option for this.

taylorotwell:
Yeah I basically agree with that issue that a configuration option makes more sense.

PR #4615, opened 2018-10-09
Refer to `contains` at `has` to clarify how to determine value existence

tomlankhorst:
Idiomatically, `has` could just as well refer to a collection having a certain value. 
To redirect the user to the corresponding function, I've added a reference.


taylorotwell:
I'm not sure that is needed here.

PR #4616, opened 2018-10-10
add trait

emtudo:


taylorotwell:
It's not necessary because the Foundation user already uses the trait.

PR #4617, opened 2018-10-10
Added info on configuring authentication headers

happyDemon:
Definitely handy when you work on projects where your front-end code is completely separate from your laravel application.
I had to dive into the source code of laravel/echo to figure this one out.


taylorotwell:
This doesn't really include a lot of information about where the token should come from, where should this file change be made?

PR #4620, opened 2018-10-10
[5.7] Fix nginx starter config

michaelkozakovsc:
To make the nginx config work out of the box, you need to import `fastcgi.conf` instead of `fastcgi_params`


taylorotwell:
Depends on your machine and configuration.

PR #4634, opened 2018-10-12
Document $query Type for `whereHas`

mcordingley:
I keep having to look this up and would love it to simply be a part of the docs.


taylorotwell:
At least import it please. :)

PR #4639, opened 2018-10-15
[5.8] Added upgrade notice about 401 error image.

TBlindaruk:
- related to https://github.com/laravel/framework/pull/26116 changes


taylorotwell:
I think we may need more information on where to get the image, etc. Honestly I would rather the 401 page just re-use the 403 image.

PR #4645, opened 2018-10-16
[5.7] Queue::assertPushedWithChain

tomlankhorst:


mocking.md
```diff
@@ -278,6 +278,12 @@ As an alternative to mocking, you may use the `Queue` facade's `fake` method to
 
             // Assert a job was not pushed...
             Queue::assertNotPushed(AnotherJob::class);
+          
+            // Assert a job was pushed with a specific chain...
```
tomlankhorst: 'specific' because asserting a partial or reordered chain fails.

PR #4646, opened 2018-10-16
[5.5] Mentioning one more effect of cookie serialization

msonowal:
When the application uses the below middleware in the applications
`\Laravel\Passport\Http\Middleware\CreateFreshApiToken::class`
in case of spa and mobile apps etc, then also it needs be serialized set to true

This happened with me when I followed the upgrade guide and I had no clue at first when my api are returning unauthenticated. when updated from 5.4

Then I drilled down into the framework and saw the implications from this cookie.

This will very helpfull when someone updates with the middleware in the codes


taylorotwell:
Passport has its own `withCookieSerialization` method:

https://github.com/laravel/passport/blob/7.0/src/Passport.php#L522

msonowal:
Hi @taylorotwell ,
Thanks for the quick response, but you mentioned there is the method but how it solves the actual problem I am not able to link it to. Can you through some light into it.

As my laravel app is running with Passport for mobile apps and vueJS spa on the web client so I use normal authentication users to use the same api routes that are protected through api middleware , which were returning unauthorized after upgrade.

PR #4652, opened 2018-10-19
[5.7] Passport: clarify JSON API usage

driesvints:
There seems to be a lot of confusion about the JSON API that ships with Passport:

https://github.com/laravel/passport/issues/839
https://github.com/laravel/passport/issues/371
https://github.com/laravel/passport/issues/842

The JSON API is guarded by the web and auth middlewares which means that they can only be consumed by the application they're implemented in. In most situations this works well but when you try to consume the API externally or from a subdomain it becomes cumbersome.

It would be better if the JSON API would behave as an actual api without having to go through all of the web middlewares and just use the auth:api middleware for authentication. There's a feature request open for this: https://github.com/laravel/passport/issues/379

In the meantime we can add some clarification to the docs for people struggling with this.

### DIFF
--- a/passport.md
+++ b/passport.md
@@ -206,6 +206,8 @@ However, you will need to pair Passport's JSON API with your own frontend to pro
 
 > {tip} If you don't want to implement the entire client management frontend yourself, you can use the [frontend quickstart](#frontend-quickstart) to have a fully functional frontend in a matter of minutes.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
As mentioned in a previous PR, we can't have back to back notes. The documentation will have to be worked into the flow of the document.

PR #4653, opened 2018-10-19
[5.8] Explaining why closures are important for factories

clementblanco:
Adding, I think, some useful information around the closures for factories when writing them using relationships.

I guess this happens because an array merge is involved somewhere down the line.

Credits: @tiagocyberduck.


taylorotwell:
No plans to add this.

PR #4654, opened 2018-10-19
[5.7] Passport: document integer column for user identification

driesvints:
I can't even begin to count the amount of times I encountered issues with people not realizing the user ids are saved as integers. I think an explicit note to the docs is appropriate to make people aware that user ids will be saved as integers and that they can overwrite the migrations for this is they want to. Hopefully this will solve future issues and mistakes.


taylorotwell:
Back to back notes and typos.

PR #4658, opened 2018-10-19
[5.7] Passport: client credentials scopes

driesvints:
Should explain how a client credentials route can be protected by the client middleware and specific scopes.

It's also not very logic to use the user endpoint in this example because you're not making the request with an active user. It's a machine-to-machine request. Adjusted this as well.

https://github.com/laravel/passport/issues/501


driesvints:
Rebased with 5.7 to solve the merge conflicts.

PR #4659, opened 2018-10-19
[5.7] Passport: add note about EncryptCookies middleware

driesvints:
The CreateFreshApiToken middleware cannot function without the EncryptCookies middleware being placed before it. Some people stumbled upon this so it might be better to explicitely note in the docs that it's a requirement.

https://github.com/laravel/passport/issues/484


taylorotwell:
We can't have back to back notes unfortunately. They don't display correctly.

PR #4660, opened 2018-10-19
[5.7] Extra passport configuration

driesvints:
This PR adds extra documentation about ways to configure Passport for your setup that were missing. I've tried splitting the different additions in different commit messages so it's clear what was added. Let me know if you want to describe some things differently.

### DIFF
--- a/passport.md
+++ b/passport.md
@@ -6,6 +6,7 @@
     - [Deploying Passport](#deploying-passport)
 - [Configuration](#configuration)
     - [Token Lifetimes](#token-lifetimes)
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think we would need a good example of *why* you would disable CSRF protection.

driesvints:
@taylorotwell I'll whip something up in the coming days 👍

driesvints:
@taylorotwell I'm actually gonna wait with sending that in until we've reached an agreement on https://github.com/laravel/passport/issues/379

PR #4668, opened 2018-10-22
[5.7] HTTP Tests: Add note about escaping content

nedmas:
When using `assertSeeText`, `assertSeeTextInOrder` or `assertDontSeeText` with Blade templates you need to escape content before testing to ensure you don't get any false negatives. This PR just adds a brief note to highlight that fact in the documentation.

See https://github.com/laravel/framework/pull/21389, https://github.com/laravel/framework/pull/19902 and https://github.com/laravel/framework/pull/26169 for reference.

### DIFF
--- a/http-tests.md
+++ b/http-tests.md
@@ -442,6 +442,8 @@ Assert that the given strings are contained in order within the response:
 Assert that the given string is contained within the response text:
 
     $response->assertSeeText($value);
+    
+> {note} If you escape HTML special characters in your templates (this is the default behaviour for Blade), then you will need to escape the value before testing. E.g. `$response->assertSeeText(e($value));`
 
 <a name="assert-see-text-in-order"></a>
 #### assertSeeTextInOrder


taylorotwell:
I'm just curious how common this really is? In general, I'm not sure it's really relevant?

nedmas:
@taylorotwell How common is it? I don't know, but I'd be interested to find out. Is it really relevant? Well it was to me and I'd like to think that means it might be to someone else too.

Here's the logic flow in my head as to why it's relevant:

Laravel ships with Blade which escapes content by default, Laravel also ships with Faker for use in factories which generates names that include apostrophise (actually quotes), finally Laravel ships with additional test assertions for checking if content is in a response. I expected to be able to use all of this as documented but when I did I got a false negative.

Personally I thought that it would be worth changing the test assertions, but I respect the fact that you've already decided against that. So I just wanted to make it easier for others to work out what to do as it took me a while to figure out and even then I didn't know about the `e()` function.

As an aside I've realised that this doesn't just effect the methods I listed above but all of the methods that compare the response content.

PR #4677, opened 2018-10-24
document withoutMiddleware to avoid disable route binding gotcha

jtallant:
It seems obvious now but it was very frustrating having the route binding disabled when disabling all middleware. Sometimes I like to start writing tests in a very basic way and not worrying about middleware at all until later. I do that now by only disabling specific middleware but this took me a bit to figure out. Other people have had the same issue.

https://github.com/laravel/framework/issues/16260

I would recommend just keeping the route binding middleware on when calling `withoutMiddleware` and then deleting this doc. Makes a lot more sense to me but others disagree. It should at least be documented if it's going to be that way.

### DIFF
--- a/http-tests.md
+++ b/http-tests.md
@@ -3,6 +3,9 @@
 - [Introduction](#introduction)
     - [Customizing Request Headers](#customizing-request-headers)
 - [Session / Authentication](#session-and-authentication)
+- [Without Middleware](#without-middleware)
+    - [Disabling All Middleware](#disabling-all-middleware)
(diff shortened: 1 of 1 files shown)


taylorotwell:
Personally I don't want to document disabling middleware at all. IMO it's a bad idea and generally leads to bugs. You should test your routes with their full middleware.

SachinAgarwal1337:
@taylorotwell I agree, but there are situations where disabling middleware is required. For example when I am testing the route with `signed` middleware. where I want to just test the call the route without creating the signed route in my test.

Issue #4672, opened 2018-10-23
Consider documentating how to create an eloquent model with no primary key?


samoldenburg:
Perhaps this isn't a common issue, so I wanted to ask before opening a PR, or maybe what the best way to help others might be. I see that there have been questions asked and answered about this on places like Stack Overflow before.

I'm working with a database that I have absolutely no control over the schema on - it's a `sqlsrv` driver connection, and has a table with no primary key that I wanted to write an eloquent model for.

I found that you can disable the primary key in the model by setting `protected $primaryKey = null;`, but this alone wasn't enough to do the trick, you apparently need to set `public $incrementing = false;` or it'll bomb out on a pretty tricky to debug exception:

```Too few arguments to function Illuminate\Database\Eloquent\Model::setAttribute(), 1 passed in /var/www/app/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php on line 592 and exactly 2 expected```

I'm sure a snippet about this could be added to the documentation under the model conventions section specifically under the Primary Keys heading: https://laravel.com/docs/5.7/eloquent#eloquent-model-conventions 

I think it could just be a one sentence quip, something along the lines of `If your table has no primary keys, you can set protected $primaryKey = null; and public $incrementing = false;, but note that not having a primary key is considered bad practice.`

driesvints:
Heya, thanks for sending this in. I think it's best to just send a PR for this and see if Taylor will accept it or not.

PR #4682, opened 2018-10-25
[5.7] Document Eloquent model with no primary key

samoldenburg:
See #4672 for details on my thoughts for this suggestion.

While this question can be answered with some basic google-fu should others run into it, I think having this officially documented might be ideal. Or it might not, I'll defer to your judgment on that.


taylorotwell:
I don't want to recommend using Eloquent models without primary keys.

PR #4683, opened 2018-10-25
[5.7] Added extra information about throttling.

lotharthesavior:
Hello,

I'm very thankful for Laravel project, and I always wanted to contribute somehow. During work, I saw a problem that wasn't being covered in the documentation, so here is a little addition that would help others to solve this challenge. 

I added extra information about throttling with Redis functionality, so users can be informed about a way of avoiding timeout errors. It happens when you configure it without adding the ``->block()``.

Thanks!
Savio Resende

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -394,6 +394,16 @@ If your application interacts with Redis, you may throttle your queued jobs by t
 
 > {tip} In the example above, the `key` may be any string that uniquely identifies the type of job you would like to rate limit. For example, you may wish to construct the key based on the class name of the job and the IDs of the Eloquent models it operates on.
 
+It is very important to add ``->block(60)`` with the amount of time being at least the same time you put for the ``->every(60)``, so you avoid timeout errors. The code with that would look like this:
+
+    Redis::throttle('key')->allow(10)->every(60)->block(60)->then(function () {
+        // Job logic...
+    }, function () {
+        // Could not obtain lock...
+
+        return $this->release(10);
+    });
+
 Alternatively, you may specify the maximum number of workers that may simultaneously process a given job. This can be helpful when a queued job is modifying a resource that should only be modified by one job at a time. For example, using the `funnel` method, you may limit jobs of a given type to only be processed by one worker at a time:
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hmm, I don't quite understand this sentence. Can you explain what you are trying to protect people from?

lotharthesavior:
I was using the code just like it is in the laravel documentation, but I was getting timeout errors with my queues. I have 3k requests to run hourly in sequence against an API that accepts 200/min.

After investigating on the web for similar problems with that feature, I noticed that a few other people were getting problems with that too, some with "too many attempts". In my case, the problem was the timeout. The problem was only solved for me, and for some other cases after I try to implement the same solution as this case here: https://github.com/laravel/framework/issues/21346

If I misunderstood the issue, please, let me know, but as far as I could see, this ``block`` is helping the worker to wait for a proper moment to run again, so we don't run into problems. It would really depend on the approach, but it might be very useful to have the ``block`` mentioned there.

I would mention in the documentation that the example code snippet would be located at the method ``handle``, of the job.

taylorotwell:
Sorry, I still don't understand. Maybe @driesvints can help re-word it and re-submit?

driesvints:
@taylorotwell @lotharthesavior I tried to wip something up here: https://github.com/laravel/docs/pull/4701

I still don't 100% get why this is necessary because I don't have too much experience with this kind of situations.

elbojoloco:
Thank you for this @lotharthesavior ! I have also been struggling with those timeout errors and using setting the `block()` and `every()` to the same amount of seconds finally eliminated the problem. This has cost me much more time than I wanted and I've more posts about people getting confused when trying to throttle jobs. 

Using a time-based retry simply isn't a viable option when for example bulk sending emails. I strongly agree with adding this to the docs as it will prevent a lot of wasted time for people dealing with the same issue, or at least add that it is required to set the `block()` and `every()` to the same value to prevent timeout/retry exceptions.

elbojoloco:
@taylorotwell I can provide a concise example if necessary to help decide wether this should go in the docs or not.

PR #4684, opened 2018-10-25
Fix small typo on Events page

enderandpeter:
*provide* is more appropriate because the subject, *Laravel's events*, is plural.


driesvints:
Thanks!

PR #4686, opened 2018-10-26
[5.7] Update the User Factory definition

ahinkle:
This PR adds the new `email_verified_at` to the User Factory example.

https://github.com/laravel/laravel/blob/f9cf7ff1c834bd918a21930288d12d04e0f15a00/database/factories/UserFactory.php#L18-L22


driesvints:
Thanks!

PR #4703, opened 2018-10-30
Add quotes to 'laravel-worker:*'

Shifterovich:
Absence of quotes causes zsh to misunderstand this command. See https://github.com/robbyrussell/oh-my-zsh/issues/31

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -530,7 +530,7 @@ Once the configuration file has been created, you may update the Supervisor conf
 
     sudo supervisorctl update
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't think they are needed typically.

Shifterovich:
Adding them won't break anything and will ensure it works on all systems:

```
╭─ubuntu@g551jk ~  
╰─$ sudo supervisorctl start laravel-worker:*  
zsh: no matches found: laravel-worker:*
```

PR #4715, opened 2018-11-01
[5.7] Explain issues of dispatching jobs from tinker.

mkarnicki:
I wasted half a day wondering why dispatching jobs wasn't working as expected. Turned out it was a combination of using tinker and the fact that `dispatch` method uses garbage collection to put jobs on the queue.

I filed [this bug](https://github.com/laravel/tinker/issues/59#event-1937877824), which was closed, because [two](https://github.com/laravel/tinker/issues/28) [other](https://github.com/laravel/tinker/issues/30) where already filed concerning the same problem. Many people are wasting their time when faced with this problem, so it would make sense to document that "known subtlety" (citing Graham here).

Thanks.

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -195,6 +195,8 @@ Once you have written your job class, you may dispatch it using the `dispatch` m
         }
     }
 
+> {note} The `dispatch` method depends on garbage collection to place the job on the queue. Therefore, when using Artisan `tinker` REPL in particular, you should use `Bus::dispatch` or `Queue::push` instead.
+
 <a name="delayed-dispatching"></a>
 ### Delayed Dispatching
 


taylorotwell:
Would rather add this to Tinker readme documentation.

mkarnicki:
I believe this would have much better visibility in the main docs instead. I have resubmitted it under [laravel/tinker](https://github.com/laravel/tinker/pull/61). Thank you.

ryancwalsh:
@mkarnicki I agree that it belongs in very prominent documentation. However, even with your commit, I'm unclear about how to get my job to work within Tinker. 

I've tried all sorts of ways:

```
dispatch(new \App\Jobs\CheckWhetherFacebookAdsStartedRunningJob);
$x = dispatch(new \App\Jobs\CheckWhetherFacebookAdsStartedRunningJob);
$x = Bus::dispatch(new \App\Jobs\CheckWhetherFacebookAdsStartedRunningJob);
$x = Bus::dispatch(new \App\Jobs\CheckWhetherFacebookAdsStartedRunningJob());
Bus::dispatch(new \App\Jobs\CheckWhetherFacebookAdsStartedRunningJob());
```

It would help to have a clear example in the docs.

Thanks.

cc @taylorotwell

brucek2:
Tonight I became one of the latest victims of this.  IMO adding a note to tinker is irrelevant when the queue documentation leaves out such an important fact that it might as well be false.

The documentation states that "Once you have written your job class, you may dispatch it using the dispatch method on the job itself." No one would take that to imply you must also ensure its non-documented return value is then garbage collected before the dispatch will happen. Why not add a sentence right there to explain this before the next victim loses a couple hours trying to track it down?

ryancwalsh:
@brucek2 I think I agree with you but "victim" feels a little extreme. 😉 If you feel like you know how to document it clearly (and I still don't), could you submit a Pull Request? Maybe that will help us all along.

mkarnicki:
> {note} The `dispatch` method depends on garbage collection to place the job on the queue. Therefore, when using Artisan `tinker` REPL in particular, you should use `Bus::dispatch` or `Queue::push` instead.

@ryancwalsh , that would be:
`Bus::dispatch(new \App\Jobs\Foo());` or `Queue::push(new \App\Jobs\Foo());`

mikew1:
Another victim of this (for the second time).  I think it should be added to the main queue docs. 
**_When I set up a queue, I read the queue docs, not the tinker docs._**  Also, as a matter of course I test things using tinker, so I don't get how this is not deemed to be a gotcha in the waiting.

PR #4731, opened 2018-11-06
Add some missing softwares included in homestead

pedrofurtado:
I've added some missing softwares included in homestead ( https://github.com/laravel/settler/blob/master/scripts/provision.sh ), but not documented in official page of Laravel docs.


taylorotwell:
Good grief this feels like a long list. It may be worth alphabetically organizing all this.

pedrofurtado:
Do you suggest to order alphabetically? I could update the PR, if necessary. It is not necessary to close it.

PR #4733, opened 2018-11-08
[5.7] Add whenEmpty + variants to Collection

andrey-helldar:
See a PR laravel/framework#26345


taylorotwell:
The documentation is wrong. They don't execute when the first argument is true... they execute when the collection is empty. Feel free to re-submit with fixes.

PR #4734, opened 2018-11-08
[5.7] Added note for default preserving keys on reverse collection

jordinbrwer:
This PR adds the note that the reverse function on a collection does preserve keys by default and suggests a method to not preserve the keys.

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -1455,6 +1455,8 @@ The `reverse` method reverses the order of the collection's items, preserving th
         ]
     */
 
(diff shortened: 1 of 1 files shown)


collections.md
```diff
@@ -1454,6 +1454,7 @@ The `reverse` method reverses the order of the collection's items, preserving th
             0 => 'a',
         ]
     */
+The returned reversed collection will preserve keys by default. If you do not wish to preserve the original keys, you can use the [`values`](#method-values) method to reindex them.
```
driesvints: Please add a white line above this paragraph.

taylorotwell:
All of the methods preserve keys by default where applicable though. I'm not sure we need a note on each method.

PR #4748, opened 2018-11-15
Typos in Subscription Anchor Date section

justincone:
At the beginning of the Subscription Anchor Date section, there was an erroneous period and an extra word:

>By default. the billing cycle anchor will is the date the subscription was created...

Changed to:

>By default, the billing cycle anchor is the date the subscription was created...


driesvints:
Thanks!

PR #4766, opened 2018-11-20
[5.7] Add a comment to onEachSide

miyakeylab:
I added a comment so that everyone can easily image the method


taylorotwell:
Hmm, I think it's OK as-is.

PR #4767, opened 2018-11-20
[5.7] Fixing listening to other Namespaced events example

Naoray:
I just had to deal with this yesterday. When using the example from the docs and writing it like `.Namespace.Event.Class` it does not ever get triggered, since the `format` method in `Echo` returns the event name as it is after removing the first dot:

```js
format(event: string): string {
        if (event.charAt(0) === '.' || event.charAt(0) === '\\') {
            return event.substr(1); // here the event with a `.` prefixed gets returned
        } else if (this.namespace) {
            event = this.namespace + '.' + event;
        }

        return event.replace(/\./g, '\\');
    }
```


driesvints:
I *think* this is correct. See a related PR which was accepted because of this: https://github.com/laravel/echo/pull/140/files

PR #4778, opened 2018-11-23
[5.7] Repetition removed

lovePizza:


taylorotwell:
Eh, I think it's worth keeping.

PR #4780, opened 2018-11-24
Update routing.md regarding fallbacks

TivoSoho:
Added more info on fallbacks. Mainly about assets exclusion and HTTP status. Added different declaration styles.


taylorotwell:
I've added a note that it should always be the last route registered by your application.

TivoSoho:
I am confused. If it is the last route, then it will start returning 404 HTTP status and excluding assets?

Also how does the last route exactly work if you have extensions that add routes? Are the developer declared routes last? Or last in grouping only as otherwise greedy match will match it all?

TivoSoho:
For the record, if you bothered you read, the issue with assets is that it is loading the 404 as a route, not the original 404. And loading a route means it is registered as last URL. And that has problems if you have code that redirects.

For example: 
- an asset is 404 on a page
- the user is logged out due to session expiry
- after successful login the user is redirected to the missing asset route since it registered as the "previous URL" the user was at

This is important information, I find.

PR #4785, opened 2018-11-26
[5.7] Document encoded forward slashes

driesvints:
Documents the way https://github.com/laravel/framework/issues/22125 can be solved.

Similar to the docs in Symfony: https://symfony.com/doc/current/routing/slash_in_parameter.html


driesvints:
Holding off for now because there seems to be a deeper problem.

driesvints:
Added a note to be explicit about the last route segment.

PR #4795, opened 2018-11-28
[5.7] Document replicate method

driesvints:
I wasn't super sure where to put this. It doesn't really belong under the "Retrieving Models" section but couldn't figure out a good alternative. The method itself in the `Model` class is right below the `refresh` method so this might not be the worst place.

This also comes with an important gotcha note about replicating models with their `$withCount` and `$storedAs` properties set.

See https://github.com/laravel/framework/issues/26562

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -227,13 +227,10 @@ You can replicate models using the `replicate` method. The `replicate` method wi
 
     $flight = App\Flight::first();
 
-    $flight->id; // 1
     $flight->plane; // "Concorde"
 
     $replicatedFlight = $flight->replicate();
-    $replicatedFlight->save();
 
-    $replicatedFlight->id; // 2
     $replicatedFlight->plane; // "Concorde"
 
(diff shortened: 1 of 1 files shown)


driesvints:
I was wondering if I should also explicitly state that any replicated relationships aren't necessarily copied when saved but I think that's perhaps more of a given.

tobz-nz:
Personally I would state that relationships are not replicated - better to be perfectly clear. 
Assumptions and all that.

driesvints:
@tobz-nz the thing is that only loaded relationships are replicated and when saved they're not necessary duplicated for the replicated model. Only the model itself is actually replicated.

taylorotwell:
Personally I feel like this will cause confusion with all the relationship stuff and withCount stuff that I would rather just not document it and let people replicate models themselves.

driesvints:
Yeah I agree with that. It's a pretty cumbersome method with a lot of quirks. Maybe we should remove it altogether.

PR #4802, opened 2018-12-01
[5.7] Change the hash link to the same style as other documents

kkame:


taylorotwell:
This would break the links for people who have them bookmarked, shared, etc.

PR #4811, opened 2018-12-04
[5.7] Add documentation for Telescope watchers

paras-malhotra:


taylorotwell:
This PR has conflicts. But, also, I'm not sure we need to list them all here since they are in the configuration file?

PR #4824, opened 2018-12-11
Improve custom pagination view documentation

guizoxxv:
At first I didn't understood that the `view` part of the parameter is the name of the file inside `resources/views/vendor/pagination`.


taylorotwell:
I think it's simpler to just change the view names. We really shouldn't put views in that vendor folder. I'll fix it.

Issue #2542, opened 2016-08-23
The date validation rule does not accept values like "now"


jodiedunlop:
According to the documentation the `date` validation rule accepts a value compatible with strtotime:

[date](https://laravel.com/docs/5.2/validation#rule-date)

> The field under validation must be a valid date according to the strtotime PHP function.

However, strtotime() will handle "now", but I find that in 5.2 the date validation rule does not consider this valid.

laurencei:
Hi,

I've tested this, and it works as expected for Laravel 5.5 and above.

If you believe there are still improvements that can be made, please feel free to submit a PR with the changes, or open a new ticket with the details of the problem.

Thanks.

afraca:
@laurencei  , perhaps reopen this, as it does fail with 'yesterday' for example. It's easy to see the problem:

https://github.com/laravel/framework/blob/v5.7.16/src/Illuminate/Validation/Concerns/ValidatesAttributes.php#L377-L383

This would be happening:
https://ideone.com/KD5YAN

in:
```
<?php
 
var_dump(strtotime('yesterday'));
 
var_dump(date_parse('yesterday'));
```
out:
```
int(1544400000)
array(13) {
  ["year"]=>
  bool(false)
  ["month"]=>
  bool(false)
  ["day"]=>
  bool(false)
  ["hour"]=>
  int(0)
  ["minute"]=>
  int(0)
  ["second"]=>
  int(0)
  ["fraction"]=>
  float(0)
  ["warning_count"]=>
  int(0)
  ["warnings"]=>
  array(0) {
  }
  ["error_count"]=>
  int(0)
  ["errors"]=>
  array(0) {
  }
  ["is_localtime"]=>
  bool(false)
  ["relative"]=>
  array(6) {
    ["year"]=>
    int(0)
    ["month"]=>
    int(0)
    ["day"]=>
    int(-1)
    ["hour"]=>
    int(0)
    ["minute"]=>
    int(0)
    ["second"]=>
    int(0)
  }
```

PR #4825, opened 2018-12-12
Fixed date validation not accepting relative dates

afraca:
The date validation does not accept _any_ value that `strtotime` understands, as it does `date_parse` with `check_date` afterwards. 
There was an issue for this, but was closed, although I could reproduce it.
https://github.com/laravel/docs/issues/2542

`app('validator')->make(['foo' => 'yesterday'], ['foo' => ['date']])->validate()`

edit: Fixed typo


afraca:
Another option would be using 'absolute'...

PR #4828, opened 2018-12-13
Remove ::class from use statement

GuidoHendriks:


GuidoHendriks:
Or should this even be changed to `App\Providers\TelescopeServiceProvider`? 🤔

PR #4829, opened 2018-12-13
[5.7] Make it clear that /vendor/mail/markdown refers to the plain text version of emails

mikew1:
The clarification that 'markdown' directory refers to plain text version of emails is also applicable to 5.5!

After the discussion [here](https://github.com/laravel/framework/issues/23076) it appears 5.6+ docs were updated, but 5.5 weren't.  LTS is still first recourse for me at least at this time.

### DIFF
--- a/mail.md
+++ b/mail.md
@@ -411,7 +411,7 @@ You may export all of the Markdown mail components to your own application for c
 
     php artisan vendor:publish --tag=laravel-mail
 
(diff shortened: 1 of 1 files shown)


mikew1:
Apologies - wasn't sure why you tagged this "[5.7]"....  - the change is a suggestion applicable only to the 5.5 docs. This is my first pull request, just finding my feet here  :)

driesvints:
@mikew1 you're right! My bad, I used the wrong tag. Sorry about that!

mikew1:
@driesvints  no problem :)

PR #4833, opened 2018-12-13
[5.7] remove PHP mail driver

browner12:
the PHP mail driver is no longer used. It is simply an alias for the sendmail driver.

the 'mail' driver will continue to work (for now), we just shouldn't document it to help avoid confusion.

with the current wording, the program may think the native PHP `mail()` function is still being used, which it is not.

### DIFF
--- a/mail.md
+++ b/mail.md
@@ -25,7 +25,7 @@
 <a name="introduction"></a>
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


browner12:
https://github.com/laravel/framework/commit/21f7240f68c03651d5f55927aed120651b2e05ff

PR #4839, opened 2018-12-17
Explain vue assertions in components

stepanorda:
A common use case would be to create a dusk component to test vue component. I had problems with it because I didn't understand why it's not working like in example. Problem was I didn't take in account that we are already in context of component and selector in third param of assertVue() would look for it inside this component DOM. So this is my suggestion how to describe how to use them together.


driesvints:
5.6 isn't supported anymore. Can you send this to the 5.7 branch?

stepanorda:
Sure, just created https://github.com/laravel/docs/pull/4840

PR #4841, opened 2018-12-17
[5.7] add note about compiled files

browner12:
all compiled files should be generated by maintainers to defensively prevent malicious code from being snuck in.

https://github.com/laravel/laravel/pull/4882


Pyker:
There's a typo in the first "resources" word.

PR #4852, opened 2018-12-21
Add documentation for installing Telescope on Sub-Folder Laravel Installs

nowendwell:
This PR adds to the docs the clarity for users looking to use Telescope in their sub-folder projects. Before you could not have a sub-folder instance and have telescope. Sites like /admin or /app as their root would not be able to call the telescope-api routes correctly.


taylorotwell:
Eh, I don't like that you have to duplicate the other variables. This will be really brittle.

PR #4853, opened 2018-12-21
[5.7] Update exception report() snippet to match actual code

fitztrev:
In `App\Exceptions\Handler.php`, the `report()` method does not return.

https://github.com/laravel/laravel/blob/d1db3cd089d6e9f81a4c8a65d401939f85cbbd5c/app/Exceptions/Handler.php#L29-L38


driesvints:
Thanks!

PR #4855, opened 2018-12-21
[5.7] Document Echo leaveOne method

driesvints:
Was added in https://github.com/laravel/echo/pull/216


driesvints:
@taylorotwell ah didn't saw that you renamed the methods afterwards. Thanks for correcting!

PR #4858, opened 2018-12-22
[5.7] Route name prefixes use the 'as' attribute in a group

tomlankhorst:
Counter-intuitive because all other methods represent equivalent attributes in a `Route::group` so this is worth mentioning.


taylorotwell:
Probably better demonstrated in the documentation around this option itself.

PR #4872, opened 2018-12-31
Improving documentation on choice method

alexjoffroy:
Add details Illuminate\Console\Command::choice method:
- Explanation and example about `attempts` parameter
- Explanation and example about `multiple` parameter


taylorotwell:
Wording is a little confusing here. Passing ... to `true`.

alexjoffroy:
So what do you prefer ? "Setting the fifth parameter to `true`" ?

PR #4891, opened 2019-01-08
Update upgrade.md

robgeorgeuk:
Blade `or` has been removed in favor of php 7's native null coalescing operator.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -79,6 +79,15 @@ If you would like to render unescaped content in a section, you must declare the
         {!! $content !!}
     @stop
 
+#### `or` Removed
+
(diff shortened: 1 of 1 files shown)


michaeldyrynda:
Laravel 5.4 has a minimum requirement of PHP >= 5.6.4, so this could lead to confusion for those still on this (ancient) version of PHP.

robgeorgeuk:
Fair enough, it could still be worth mentioning that the blade `or` command has been removed and give details of an alternative.

robgeorgeuk:
ah, my bad, sorry for the hassle.

PR #4894, opened 2019-01-09
[5.7] Update collection every() method documentation

rbarden:
The `every()` method returns true for empty collections which is not documented but should be, since it goes against what the description would lead most developers to believe. After discussion with a few developers, we decided at least a documentation change should be put in, since changing the behavior may break many apps who rely on this in the wild.

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -517,6 +517,16 @@ The `every` method may be used to verify that all elements of a collection pass 
 
     // false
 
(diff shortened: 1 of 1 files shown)


rbarden:
I'd just like to add that I did some digging and found when this method was introduced which leads to javascript documentation with this behavior and a link to Wikipedia, so I understand why the method works the way it does now.

PR #4904, opened 2019-01-13
[5.7] Override Default Verification Functions

aschbacd:
This commit will give an example how to override the default verification functions in Laravel.


taylorotwell:
This wouldn't be a good way to implement the use case in your PR. You would just register an event listener to the Verified event to send the email without overriding the controller at all.

PR #4909, opened 2019-01-15
Remove PHP 5.6 and 7.0 from Homestead docs

sevenpixels:
PHP 5.6 and 7.0 no longer included in Homestead due to reaching EOL.


browner12:
I don't think this is happening until box v8, so let's hold off on adding this to the docs just yet.

As much as we'd like to live in this world where 5.6 and 7.0 don't exist anymore, lots of people will still have old projects that need support for this, so when we do remove it, we should probably also add a note saying if you need support for these versions, make sure you install box v7.

driesvints:
Thanks for sending this in but I think it's best if @svpernova09 sends this in when he feels the time is right for it.

driesvints:
Okay, @sevenpixels can you send in a new PR because I saw your PR was conflicting with the base branch.

PR #4911, opened 2019-01-15
Add note to passport docs

clnt:
This PR adds the following note to the Client Credentials Grant section of the Passport documentation:

> {note} When protecting routes with the `client` middleware, make sure you are using `client` instead of the `auth` middleware.

This just caught me out (I was using both auth and client) and could have been avoided if docs made it a bit more clear. 

Not sure if its best to add it here or in the Protecting Routes section.

### DIFF
--- a/passport.md
+++ b/passport.md
@@ -507,6 +507,8 @@ To restrict access to the route to specific scopes you may provide a comma-delim
         ...
     })->middleware('client:check-status,your-scope');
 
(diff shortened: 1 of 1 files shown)


clnt:
@taylorotwell Yes, it does. But also further down the docs it goes on about protecting with the `auth` middleware.

This note simply adds more clarification, what is the harm in it? It would have saved me time.

PR #4916, opened 2019-01-18
Update queries.md

dwoodard:
Should the $role as the 2nd argument be taken out and put in a 'use' ($role)
```
$role = $request->input('role');

    $users = DB::table('users')
                    ->when($role, function ($query) use($role) {
                        return $query->where('role_id', $role);
                    })
                    ->get();
```

### DIFF
--- a/queries.md
+++ b/queries.md
@@ -584,7 +584,7 @@ Sometimes you may want clauses to apply to a query only when something else is t
     $role = $request->input('role');
 
     $users = DB::table('users')
-                    ->when($role, function ($query, $role) {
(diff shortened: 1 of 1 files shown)


shadoWalker89:
No, because the $role is being passed to the callback as the second argument by the when() method when the $role evaluates to true.

driesvints:
As @shadoWalker89 explained, this is documented correctly:

https://github.com/laravel/framework/blob/8ccb11d7bf376248f2a362f0e11e874965b3ad3b/src/Illuminate/Database/Concerns/BuildsQueries.php#L80-L97

PR #4917, opened 2019-01-18
[5.7] Fix documentation for updating Homestead VM

bandgeekndb:
Just went through updating my Homestead VM and wanted to update the documentation a bit.

**Notable changes:**

- We're telling users to checkout a tagged release upon install, so they need to ensure they checkout a new tagged release after pulling the latest code.
- Once the new box is downloaded, you need to destroy and regenerate the VM to use the latest Vagrant box.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -551,15 +551,26 @@ When customizing Homestead, Ubuntu may ask you if you would like to keep a packa
 <a name="updating-homestead"></a>
 ## Updating Homestead
 
-You can update Homestead in two simple steps. First, you should update the Vagrant box using the `vagrant box update` command:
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't understand the git pull origin master, fetch, checkout a branch flow. Seems weird?

bandgeekndb:
I'm not a git expert by any means, so if there's a better way to achieve the same result I'm eager to learn it. I cobbled that together using the documentation from the Homestead installation section and my (albeit) limited knowledge of Git.

Since we checked out a particular tag during install, we need to update our code from GitHub with the git pull and git fetch, then checkout the latest tag.

PR #4931, opened 2019-01-25
[5.8] Expand section on self-contained exceptions and add an example of utilizing DI in the report method

spawnia:
See https://github.com/laravel/framework/pull/27301 for the implementation of DI for the report method.

Seperated out a section from **The Exception Handler**. While it achieves the same result, the approach of using **Self-Contained Exceptions** is orthogonal in terms of structure, so i think it makes sense to have them as two distinct sections.


taylorotwell:
Added a simple tip.

PR #4939, opened 2019-01-29
[5.7] Documents how to dispatch a job synchronously

renoirfaria:
Sometimes people tend to believe that jobs only can be used asynchronously, how about to document that they can do more than they think with jobs?


thauansantosdev:
Helpful!

PR #4940, opened 2019-01-29
[5.5] Document Delay property for Queued event listener

gmsantos:
It is possible to delay a queued event listener execution defining a `$delay` property (see https://github.com/laravel/framework/blob/5.5/src/Illuminate/Events/Dispatcher.php#L473-L475).

I'm adding that on documentation and targeting 5.5 because it's the current LTS version, but it is valid for the subsequent versions.


taylorotwell:
This version of the documentation is no longer updated. Please send the PR to the 5.7 branch.

gmsantos:
rebased in #4943

PR #4955, opened 2019-02-04

phpduck:
`orderBy` doesn't work.

Testet with MySQL and MariaDB.

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -933,12 +933,6 @@ Sometimes you may wish to eager load a relationship, but also specify additional
         $query->where('title', 'like', '%first%');
     }])->get();
 
-In this example, Eloquent will only eager load posts where the post's `title` column contains the word `first`. Of course, you may call other [query builder](/docs/{{version}}/queries) methods to further customize the eager loading operation:
-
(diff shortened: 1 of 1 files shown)


taylorotwell:
It does. Just tested on MySQL.

phpduck:
No it doesn't! I had a User model with Roles and there was no sorting at all when applied to the relation with this array like syntax.

Like I said, tested on two different systems with MySQL and MariaDB.

Show me the code.

ahinkle:
`orderBy` is tested and used often within the framework. I tried it with a relation and it works -- how about you share your query builder, instead?

phpduck:
@ahinkle @taylorotwell here is the code:

```php
class User extends Model
{
    public function roles()
    {
        return $this->belongsToMany('App\Role')->withTimestamps();
    }
}
```

```php
class Role extends Model
{
    public function users()
    {
        return $this->belongsTo('App\User')->withTimestamps();
    }
}
```

```php
class UserController extends Controller
{
    public function index()
    {
        $users = User::with(['roles' => function ($query) {
            $query->orderBy('created_at', 'asc');
        }])->get();
    }
}
```

Switching between `asc` and `desc` doesn't make any difference.

PR #4958, opened 2019-02-06
Control Paths of cache files for services, packages and routes

ahinkle:
This PR adds documentation for the newly added ENV Control Paths of cache files for services, packages and routes. https://github.com/laravel/framework/pull/27389


taylorotwell:
I will probably wait to document this a bit longer.

PR #4975, opened 2019-02-13
[master] Eliminate (almost) all instances of "of course" and "as you might expect"

### LINKED ISSUE #4973: Eliminate "of course", and "as you might expect"
These phrases can read as slightly condescending, and are massively overused in the docs.

There seems to be an implication of "aren't you silly for not knowing this thing that I just told you?" The purpose of docs is to educate, so if all of these things were simply a matter "of course", there wouldn't be the need to document them!

markjaquith:
These phrases were overused, didn't add any value, and seemed a bit condescending. People are reading the Laravel docs to become educated. The docs shouldn't imply that everything Laravel does is natural or a matter "of course", because that could make people feel bad for having to **learn** how it works instead of instinctively knowing it already.

I left three instances of "of course" intact, because they were expressing solidarity with a natural reaction to doing something in a more cumbersome way. Most instances of "of course" were merely removed, but a few were replaced with "note that" or "importantly," if it flowed better.

fixes #4973 [master branch]


taylorotwell:
I merged forward from 5.7. Thanks.

PR #4987, opened 2019-02-18
DropboxServiceProvider register in config\app.php

Parthfaladu:


driesvints:
Heya, unfortunately we don't support this version anymore. Please check out [our support policy](https://laravel.com/docs/releases) on which versions we are currently supporting.

PR #4993, opened 2019-02-18
[5.7] Describe updateOrInsert method

jaureguivictoria:
### Added
- Short description of the [updateOrInsert](https://github.com/laravel/framework/blob/5.7/src/Illuminate/Database/Query/Builder.php#L2670) Query Builder method and an example of it's usage.

### DIFF
--- a/queries.md
+++ b/queries.md
@@ -18,6 +18,7 @@
 - [Updates](#updates)
     - [Updating JSON Columns](#updating-json-columns)
     - [Increment & Decrement](#increment-and-decrement)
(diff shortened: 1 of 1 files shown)


ahinkle:
I think it should be inserted here instead to follow the documentation pattern: https://github.com/laravel/docs/blob/5.7/eloquent.md#other-creation-methods

taylorotwell:
@ahinkle this is a query builder method, not an Eloquent method.

ahinkle:
ah good catch. Thanks.

PR #4995, opened 2019-02-19
[5.7] Add token based authentication documentation

driesvints:
This adds separate docs for the built-in token based authentication which ships with Laravel. It keeps it separate from the Passport documentation which is a totally different implementation.

I only documented the bare minimum to get started with the feature and not all the ways you can present and use tokens in the front-end. These things are better covered in tutorials I believe.

### DIFF
--- a/api-authentication.md
+++ b/api-authentication.md
@@ -1,0 +1,96 @@
+# API Authentication
+
+- [Introduction](#introduction)
+- [Configuration](#configuration)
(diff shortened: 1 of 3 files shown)


taylorotwell:
Moved this to `master` branch, did some clean-up, and merged.

driesvints:
Awesome. Thanks!

PR #5000, opened 2019-02-22
[5.4] Add upgrade note for Model::newPivot() signature change

shadoWalker89:
While upgrading an app from 5.3 to 5.4 i noticied that the signature of Model::newPivot() changed by adding a new argument `$using`

This PR adds a note in the upgrade guide

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -282,7 +282,7 @@ Calling `factory(User::class, 1)->make()` or `factory(User::class, 1)->create()`
 
 #### The `newPivot` Method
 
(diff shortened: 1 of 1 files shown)


driesvints:
Heya, unfortunately we don't support this version anymore. Please check out [our support policy](https://laravel.com/docs/releases) on which versions we are currently supporting.

shadoWalker89:
I'm aware that the version is not supported, this is just an addition for the upgrade guide to allow people to upgrade their old apps. It does not require any maintenance, i guess it's a good thing to accept additions for the upgrade guide.

driesvints:
Okay, I'll re-open and see how it goes.

PR #5002, opened 2019-02-22
[5.7] Updated atomic lock code example

MiniCodeMonkey:
The return value of blocking for an atomic lock is always going to be true, so it is not necessary to wrap it in an `if` statement


taylorotwell:
Why is it always going to be true? What if it can't retrieve the lock even after 5 seconds?

MiniCodeMonkey:
Good question! If it can’t retrieve the lock, it will throw an Exception: https://github.com/illuminate/cache/blob/f78fc5ae5fc124659052030a045f39f7a357c14e/Lock.php#L113

The only place the return value is applicable is for when a callback is passed in.

PR #5007, opened 2019-02-25
Manually creating a paginator addition

jimgwhit:
I have a guide on forum:  https://laracasts.com/discuss/channels/guides/length-aware-paginator

Offers a good explanation.  This is hopefully for 5.7 and any newer versions.

I put in the PR, but where's the code.  Please assist.

 I just changed this in the docs on pagination:

```ruby
{note} When manually creating a paginator instance, if the result set is expected to be small you can manually "slice" the array of results you pass to the paginator. If you're unsure how to do this, check out the array_slice PHP function. Otherwise for a larger number of results use a LIMIT clause, or the skip and take methods for better efficiency.
```

Could you help and just add in the LIMIT, skip, and take part.


jimgwhit:
PR #5007  I have no idea where the code is, it should be there due to the PR.  I just changed this in the docs on pagination:

```ruby
{note} When manually creating a paginator instance, if the result set is expected to be small you can manually "slice" the array of results you pass to the paginator. If you're unsure how to do this, check out the array_slice PHP function. Otherwise for a larger number of results use a LIMIT clause, or the skip and take methods for better efficiency.
```

PR #5008, opened 2019-02-25
5.8] Update note on Manually Creating A Paginator

jimgwhit:
I have a guide on forum:  https://laracasts.com/discuss/channels/guides/length-aware-paginator

Offers a good explanation.  This is hopefully for 5.8 and any newer versions.

### DIFF
--- a/pagination.md
+++ b/pagination.md
@@ -79,7 +79,7 @@ The `Paginator` class does not need to know the total number of items in the res
 
 In other words, the `Paginator` corresponds to the `simplePaginate` method on the query builder and Eloquent, while the `LengthAwarePaginator` corresponds to the `paginate` method.
 
-> {note} When manually creating a paginator instance, you should manually "slice" the array of results you pass to the paginator. If you're unsure how to do this, check out the [array_slice](https://secure.php.net/manual/en/function.array-slice.php) PHP function.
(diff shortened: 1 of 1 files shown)


taylorotwell:
This wouldn't address how to provide the total number of results to the pagination instance. I'm fine with updating the comment but some direction on how to do that would need to be provided as well.

jimgwhit:
You can count any query, as example a query with group by that shows the pet owners and how many pets:
```
SELECT dc_pets.ownerid, dc_pets.petowner, COUNT(petid) AS total FROM dc_pets
WHERE dc_pets.ownerid IS NOT NULL AND dc_pets.ownerid > 0
GROUP BY dc_pets.ownerid;
```
The count in the above is not it, that's just part of the query, for example it would show Bob has 5 pets.
To get the count of the above (number of results, you just count the whole query:

```
SELECT COUNT(*) AS kount FROM
(SELECT dc_pets.ownerid, dc_pets.petowner, COUNT(petid) AS total FROM dc_pets
WHERE dc_pets.ownerid IS NOT NULL AND dc_pets.ownerid > 0
GROUP BY dc_pets.ownerid) AS t1;
```

Basically doing this:
```
SELECT COUNT(*) AS kount FROM
```
And wrapping original query in parenthesis with an alias.

PR #5012, opened 2019-02-26
[5.6] Added info about Eloquent events in case of restoring the model.

rudolfratusinski:
Added:

> Events `saving` / `saved` will also fire in case of firing `restoring` event.


taylorotwell:
Please send to an active branch of the documentation.

PR #5013, opened 2019-02-26
[5.8] Split Stripe and Braintree docs

driesvints:
Since the Stripe and Braintree libraries are drifting apart further and further in terms of functionality and in the future even totally different implementations splitting the docs up is the most sane thing to do instead of trying to make both work side-along.


taylorotwell:
Do we want to update the side navigation?

driesvints:
@taylorotwell I was thinking that people would find the Braintree docs through the note which I added and not necessarily add a new navigation item. But if you want I'll add the link.

PR #5015, opened 2019-02-27
[5.7] Adjust installation after release of 5.8

staudenmeir:


driesvints:
I'm not sure if we should encourage people to install older releases like this.

staudenmeir:
I wouldn't call it *encouraging*. IMO, the installation instructions for older versions should keep working.

There'll always be people who want to install an older version and have good reasons for it.

driesvints:
> There'll always be people who want to install an older version and have good reasons for it.

I disagree with this. The way I see it is that the old docs are just there as a reference for people who are still using it. If people want to install Laravel we should only instruct on how to do it for the latest release. Otherwise you just get people onto old unsupported versions which isn't wanted.

michaeldyrynda:
The 5.6 docs have this notation (to install 5.6.* from Composer). I'd say add it for consistency; not because it's recommended.

Alternatively, remove the install docs for unsupported versions entirely...

driesvints:
> The 5.6 docs have this notation (to install 5.6.* from Composer). I'd say add it for consistency; not because it's recommended.

Hmm, wasn't aware. I'd recommend removing those constraints on the older docs as well.

taylorotwell:
This is the 5.7 docs. It just shows how to install 5.7.

PR #5016, opened 2019-02-27
[5.7] Add note for method increment on null value

tuanpht:
I think we should add a note for the issue https://github.com/laravel/framework/issues/23954, Query Builder methods `increment`, `decrement` will not work if value of column is null.

### DIFF
--- a/queries.md
+++ b/queries.md
@@ -678,6 +678,8 @@ You may also specify additional columns to update during the operation:
 
     DB::table('users')->increment('votes', 1, ['name' => 'John']);
 
+> {tip} This does not work if the value of column is null, be sure to set a numeric default value for your columns.
+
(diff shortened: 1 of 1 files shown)


driesvints:
Is this only specific to MySQL? I'm not sure if it's even the responsibility of the framework to instruct you to set proper defaults for your column types.

tuanpht:
I think it is [SQL standard behavior](http://www-cs-students.stanford.edu/~wlam/compsci/sqlnulls) (tested with MySQL, Postgres), `select NULL + 1` => `NULL`

> An arithmetic operation involving a NULL returns NULL. For example, NULL minus NULL yields NULL, not zero. [

Maybe someone doesn't know about it. Do you think this instruction is necessary?

PR #5023, opened 2019-02-28
[5.8] Remove string from the comment

setkyar:
`via($notifiable)` must return an array. Otherwise, it will get `Invalid argument supplied for foreach()` error.


driesvints:
You've discovered a bug in the `NotificationSender`. Thanks for reporting this. I'll send in a fix.

PR #5024, opened 2019-02-28
Add information for overriding setUp and tearDown

mnabialek:
Alttough I'm using PHPUnit 7.5.6 and PHP 7.2 when running tests I got PHP error that signatures don't match. There were updated in this commit https://github.com/laravel/framework/commit/8fc3a75b413267a30e41464d358c1d0714a6a186 so I believe we should add this to upgrade guide - it has impact also when users don't use PHPUnit 8

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -482,6 +482,17 @@ The `defer` boolean property on the service provider which is/was used to indica
 <a name="testing"></a>
 ### Testing
 
+#### The `setUp` and `tearDown` methods
+
+**Likelihood Of Impact: Medium**
+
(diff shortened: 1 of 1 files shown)


driesvints:
It's already there: https://laravel.com/docs/5.8/upgrade#testing

mnabialek:
@driesvints  Not really. This part you showed suggests this is needed only for PHPUnit 5.8 and it's optional (at least I understand it like this) but in fact it's not true - because you need to change this also when using PHPUnit 5.7 and PHP 7.2 and you override setUp or tearDown methods from Laravel's TestCase

driesvints:
You're right. I've tried to clarify this differently by simply adding a title to that section: https://github.com/laravel/docs/pull/5025

PR #5025, opened 2019-02-28
[5.8] Add separate title to void notice

driesvints:
This adds a separate title to the void return type notice because currently it falls under the "PHPUnit 8" title and it wasn't clear that this is a general note which applies to all tests.

Also moved it one section up to make it even more explicit.

See: https://github.com/laravel/docs/pull/5024


upgrade.md
```diff
@@ -488,6 +488,8 @@ The `defer` boolean property on the service provider which is/was used to indica
 
 By default, Laravel 5.8 uses PHPUnit 7. However, you may optionally upgrade to PHPUnit 8, which requires PHP >= 7.2. In addition, please read through the entire list of changes in [the PHPUnit 8 release announcement](https://phpunit.de/announcements/phpunit-8.html).
 
+#### `setUp` And `tearDown`
```
mnabialek: To avoid further confustion I would put it above PHPUnit 8 subsection

driesvints: Good thinking. Done.

PR #5037, opened 2019-03-05
[5.9] Cover both arguments and options

ac1982:
The docs is only calling the artisan option in sample code.
In artisan command docs, it called like ```Artisan::call('email:send', ['user' => 'Taylor', '--force']);``` which is difference with in Scheduling.
It is also compatible with branch master.


driesvints:
Please send this to the 5.8 branch.

PR #5039, opened 2019-03-05
[5.8] Document read-only behavior of env helper

driesvints:


mathieutu:
Thanks @driesvints! 😉

PR #5052, opened 2019-03-12
[5.8] Added clarity on rate limiting for queues

bkuhl:
Unfortunately throttling does not work when using `sync` as the queue driver, only when using `redis`.  The documentation was incorrectly written in a way that gave the impression that merely having a `redis` connection is sufficient.

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -420,7 +420,7 @@ However, you may also define the maximum number of seconds a job should be allow
 <a name="rate-limiting"></a>
 ### Rate Limiting
 
-> {note} This feature requires that your application can interact with a [Redis server](/docs/{{version}}/redis).
+> {note} This feature requires that your application uses Redis as the queue driver.
 
 If your application interacts with Redis, you may throttle your queued jobs by time or concurrency. This feature can be of assistance when your queued jobs are interacting with APIs that are also rate limited.
 


taylorotwell:
Eh, not sure this is true. I just confirmed it with my own sync job:

```php
<?php

namespace App\Jobs;

use Illuminate\Bus\Queueable;
use Illuminate\Support\Facades\Redis;
use Illuminate\Queue\SerializesModels;
use Illuminate\Queue\InteractsWithQueue;
use Illuminate\Contracts\Queue\ShouldQueue;
use Illuminate\Foundation\Bus\Dispatchable;

class TestJob implements ShouldQueue
{
    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;

    /**
     * Create a new job instance.
     *
     * @return void
     */
    public function __construct()
    {
        //
    }

    /**
     * Execute the job.
     *
     * @return void
     */
    public function handle()
    {
        Redis::throttle('key')->allow(1)->every(60)->then(function () {
            info('Job Ran...');
        }, function () {
            info('Job was throttled...');

            return $this->release(10);
        });
    }
}
```

Log file:

```
[2019-03-12 13:21:23] local.INFO: Job Ran...  
[2019-03-12 13:21:29] local.INFO: Job was throttled...  
```

PR #5056, opened 2019-03-12
[5.8] Add missing use in Str::endsWith

madeITBelgium:
Every code block has the use statement expect in Str::endsWith.


driesvints:
Thanks!

PR #5060, opened 2019-03-14
Clarify @json usage in element attributes

Travis-Britz:
If I tried using double quotes instead of single quotes around `attr='@json'`, I'm sure somebody else will.

### DIFF
--- a/blade.md
+++ b/blade.md
@@ -195,7 +195,7 @@ However, instead of manually calling `json_encode`, you may use the `@json` Blad
         var app = @json($array);
     </script>
 
-You may also find it useful for seeding Vue components or `data-*` attributes:
+The `@json` directive is also useful for seeding Vue components or `data-*` attributes:
 
     <example-component :some-prop='@json($array)'></example-component>
 
--- a/blade.md
+++ b/blade.md
@@ -195,6 +195,12 @@ However, instead of manually calling `json_encode`, you may use the `@json` Blad
         var app = @json($array);
     </script>
 
+You may also find it useful for seeding Vue components or `data-*` attributes:
+
+    <example-component :some-prop='@json($array)'></example-component>
+
+> {note} Using `@json` in element attributes requires that it be surrounded by single quotes.
+
 #### HTML Entity Encoding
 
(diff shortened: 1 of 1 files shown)


browner12:
I feel like this goes more towards correct Vue usage, rather than something specific to the Blade directive.

Travis-Britz:
@browner12 I demonstrated by using the example Vue component that's included by default in a new installation because I thought Vue is the most common use case for data attributes in Laravel, but I specifically referenced `data-*` attributes for cases like `<ul data-items='@json($array)'></ul>` that would have nothing to do with Vue.

browner12:
are you sure that single quotes are required?

my guess is it has to do with correctly escaping characters when output to HTML. Are you overriding the default encoding options of `@json`? I think it's important to have `JSON_HEX_APOS` and `JSON_HEX_QUOT` enabled when outputting to HTML.

http://php.net/manual/en/json.constants.php

Travis-Britz:
@browner12 yes, single quotes are required because the json output uses double quotes.

For example:

```
:tasks='[{"id":70,"description":"foo"}]'
```

If you tried to quote this json with double quotes, you would end up with `[{` as the attribute value, and then random garbage after that (and probably most of the json displayed directly on the page).

Also, the `@json` directive uses these options for json_encode by default: `JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_AMP | JSON_HEX_QUOT`.
https://github.com/laravel/framework/blob/e7064ca184e395f86fc924e0d58d94955c734675/src/Illuminate/View/Compilers/Concerns/CompilesJson.php#L12

So quotes *within* the string are escaped with their `\u` escape sequence.

browner12:
I guess that's why I'm confused. If you're using the default encoding, you won't run into the issue with the double quotes, correct?

Travis-Britz:
The default encoding will encode the quotes *inside the string*.

So this description: `"foo" 'bar'`

will be encoded to

```
[{"id":72,"description":"\u0022foo\u0022 \u0027bar\u0027"}]
```

So the only way to include the json string inside an html attribute value is to surround it with single quotes:
```
<ul data-tasks='[{"id":72,"description":"\u0022foo bar\u0022"}]'></ul>
```

Because json has literal double quotes, unescaped. Using double quotes would terminate the attribute value early.

browner12:
*ahh*, got it. thanks for clearing that up for me.

Travis-Britz:
No problem! I was confused too, hence the PR.

PR #5065, opened 2019-03-15
Described how to chain artisan commands

galliraf:
I described how to execute two artisan command one after the other.


taylorotwell:
I don't think needs its own section in the docs.

PR #5070, opened 2019-03-19
[5.8] Update eloquent-relationship.md

clementblanco:
Adding in some notes around Eloquent relationships being case sensitive.

This is linked to a few places I could see this debated/explained:

* https://github.com/laravel/framework/issues/3419#issuecomment-33844607
* https://github.com/laravel/framework/issues/13614#issuecomment-221724076


taylorotwell:
Added a note about this to the serialization page. I don't think we need such a large section that primarily states that we use PSR-2. Thanks!

PR #5072, opened 2019-03-20
Fix link in braintree.md and add Braintree link in documentation.md

clemir:


driesvints:
Thanks but we don't plan on adding the link to braintree in the menu. I'll fix the incorrect link in the note.

PR #5074, opened 2019-03-21
[5.8] Intermediate model name change.

Tpojka:
Respecting advice[/rule] to set intermediate table name in alphabetical order first_table_singular->underscore->last_table_singular, in same fashion here is proposed name for pivot model of that table.

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -335,7 +335,7 @@ You can also filter the results returned by `belongsToMany` using the `wherePivo
 <a name="defining-custom-intermediate-table-models"></a>
 ### Defining Custom Intermediate Table Models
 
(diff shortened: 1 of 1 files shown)


driesvints:
My 2 cents: you don't need to map your model/pivot name to your DB scheme. An ORM allows for these to be decoupled and the current class name reads better. I do agree with you that the current example is wrong. But maybe this can more easily be solved by adding the table name property to the `UserRole` class example:

```php
<?php

namespace App;

use Illuminate\Database\Eloquent\Relations\Pivot;

class UserRole extends Pivot
{
    /**
     * The table associated with the model.
     *
     * @var string
     */
    protected $table = 'role_user';
}
```

Tpojka:
@driesvints 
For sure, what you said is correct. I just think that docs should contain mvp with no misleading. I saw so many questions on Stack Overflow that had more errors because of not following basic proposed rules before extending custom solutions (not implying `$table` property is one). Like, to be able to answer on those would require more time to set naming of classes/methods/properties right (with or without overwriting similar to your mentioning in code block) than what time would require solving problem itself. 
Without much custom changes, it's easier to debug error and determine if it's up to application or repository (framework itself) bug/feature level.
Either way it should be changed in docs - changing class name (which I think would be better solution) either setting property name in example class in documents.

PR #5075, opened 2019-03-21
[5.8] Add upgrade note about publishing Horizon's assets

drbyte:
While this could be more detailed and verbose for the benefit of troubleshooting, it's unclear whether that would be accepted, so I chose to use the same brief wording as the Telescope docs use.


driesvints:
@drbyte awesome, thanks. I've updated the PR to make it a bit more in line with Cashier's upgrading instructions.

PR #5076, opened 2019-03-21
[5.8] Add note about getenv function

driesvints:
This should make the change to this behavior more obvious without just refering to the upgrade guide of phpdotenv.


driesvints:
Previous behavior was reverted.

PR #5077, opened 2019-03-21
[5.8] Set correct guard in example code

jasonmccreary:


taylorotwell:
When you authenticate using a given guard, that guard is automatically set as the "default" guard, so this shouldn't be necessary unless I'm mistaken?

jasonmccreary:
Didn't work for me. I got `null` without it. Copied the code from the docs and put it in `routes/api.php`. I may be missing something as well. But adding it just worked…

ankurk91:
I can confirm that the default guard is being set automatically.
No need to specify the guard name.

crnkovic:
Works fine for me too.

jasonmccreary:
Laravel 5.8, and doesn't work for me without passing the `api` guard.

jasonmccreary:
I see now based on the details of @taylorotwell's original comment that because my `default` guard (as set in `config/auth.php`) is set to `web`. As such, I needed to supply the 'api' guard.

So this code change, while more universal, is not _always_ necessary. As such, might be worth a _Note_ in docs. Glad to change this PR if you agree. Otherwise, close it.

PR #5090, opened 2019-03-27
[5.8] Recommend Fake Cron package for development

alexbowers:
There are a few reasons why having a cron setup isn't necessarily ideal.

- If you want to have something in the scheduler run more often than once a minute. Cron doesn't have finer granularity than once a minute, however, the recommended package does.

- If you don't want to modify the crontab on that machine (for instance, in development mode), where if you work on several projects you could have many cron entries setup, constantly working in the background, and not in your control.

### DIFF
--- a/scheduling.md
+++ b/scheduling.md
@@ -29,6 +29,14 @@ When using the scheduler, you only need to add the following Cron entry to your 
 
 This Cron will call the Laravel command scheduler every minute. When the `schedule:run` command is executed, Laravel will evaluate your scheduled tasks and runs the tasks that are due.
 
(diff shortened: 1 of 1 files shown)


ahinkle:
Community packages aren't typically in the documentation. Perhaps creating a blog post to share with the community would suffice.

alexbowers:
Whilst I'd agree, there is precedent for this https://laravel.com/docs/5.8/filesystem#custom-filesystems

taylorotwell:
I don't like to make a habit of recommending community packages from the official documentation.

PR #5092, opened 2019-03-28
[5.8] Added more clear description to the unique rule

DarkGhostHunter:
It sounds like the value should exists only once in the database. Added a tiny phrase in the description to clearly show that `unique` is just like an hypothetical `doesnt_exists`.

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -1003,7 +1003,7 @@ The field under validation must be a valid timezone identifier according to the 
 <a name="rule-unique"></a>
 #### unique:_table_,_column_,_except_,_idColumn_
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Feels really redundant.

DarkGhostHunter:
> Feels really redundant.

In what aspect? Then what about putting this?

> The field under validation must not exists in a given database table. If the `column` option is not specified, the field name will be used.

I'm putting down this because [I'm not the only confused about this](https://itnext.io/laravel-validation-rule-doesnt-exist-d828eec6eb62).

lethalwire:
Perhaps change the docs to read: 

The field under validation must not exist within the given database table. If the `column` option is not specified, the field name will be used.

PR #5093, opened 2019-03-29
[5.8] Add jQuery tip about AjaxSetup

mariovalney:
jQuery has a alert about the ajaxSetup:

link: https://api.jquery.com/jquery.ajaxsetup


taylorotwell:
I think that is already implied by the code example.

PR #5097, opened 2019-03-30
Removed an "a" article before a plural form

DCzajkowski:


taylorotwell:
It's not plural it's possessive?

DCzajkowski:
Interesting. I was 100% sure the article is for the "sessions", not the "user". I was wrong.
Living you learn. Sorry for the confusion.

PR #5100, opened 2019-03-31
Update migrations.md

emehrkay:
It isn't very clear that the migration name must match the file name. I made a migration via the command line called `resources` and edited the class name to be `CreateResource` and it kept complaining that the the class `Resources` was not found.

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -41,6 +41,8 @@ The `--table` and `--create` options may also be used to indicate the name of th
 
     php artisan make:migration add_votes_to_users_table --table=users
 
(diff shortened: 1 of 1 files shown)


taylorotwell:

emehrkay:
You're free to reject this PR in particular, but this is still an issue. You can either explicitly outline that the class name must match the migration file name in the docs, or even better, update the error message to reflect that structural decision. This was an error that I caused (by following tutorials or whatever), but it didn't click for me until I saw a random comment while searching for `artisan migrate class not found` and someone happened to say `it must be named $date_time_create_table.php`

PR #5103, opened 2019-04-03
Delete repeated example code

asilvabe:
Delete repeated example code

It seems that the sample code is repeated.


asilvabe:
I apologize for sending the PR so hastily. I did not read well the difference between both portions of code.

PR #5104, opened 2019-04-03
[5.8] Update telescope version in documentation

DemianD:
Currently I'm trying to add Telescope to my Laravel 5.7.28 app. The documentation says that it needs to be at least 5.7.7. However, this doesn't work as it 5.8+.

I've updated the documentation according that it only supports versions 5.8+ 

https://github.com/laravel/telescope/blob/2.0/composer.json#L19

image

### DIFF
--- a/telescope.md
+++ b/telescope.md
@@ -38,7 +38,7 @@ Laravel Telescope is an elegant debug assistant for the Laravel framework. Teles
 <a name="installation"></a>
 ## Installation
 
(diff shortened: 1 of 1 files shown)


themsaid:
You need to specify `"laravel/telescope": "1.*"` in your composer.json file for it to work.

taylorotwell:
I've pushed an update to 5.7 docs to show the correct version.

DemianD:
Thank you 😊

PR #5115, opened 2019-04-08
[5.8] In version 5.8 we have "Auto-Discovery Of Model Policies". This make …

Polfo:
In version 5.8 we have "Auto-Discovery Of Model Policies". This make the "Resource Gates" section redundant and confusing.

- 'App\Policies\PostPolicy@view' and all its sibling methods will also be auto-discovered.  
- The dot notation used in the Resource Gates section is confusing: `posts.view` is used here instead of `view-post` elsewhere. This might suggest that something automatic is hapening and that `Gate::allows('view', Post::class);` could be valid while `Gate::allows('posts.view', 'foo');` is.
- Creating method `view()` in `App\Policies\PostPolicy` will automatically define Gate::allows('view', Post::class);` This overlaps `Gate::define('posts.view', 'App\Policies\PostPolicy@view');` in the "Resource Gates" section.

### DIFF
--- a/authorization.md
+++ b/authorization.md
@@ -64,26 +64,6 @@ Gates may also be defined using a `Class@method` style callback string, like con
         Gate::define('update-post', 'App\Policies\PostPolicy@update');
     }
 
-#### Resource Gates
-
-You may also define multiple Gate abilities at once using the `resource` method:
-
-    Gate::resource('posts', 'App\Policies\PostPolicy');
-
-This is identical to manually defining the following Gate definitions:
-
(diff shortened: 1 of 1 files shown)


DarkGhostHunter:
Whatever is done to clarify Gates and Policies is welcome.

taylorotwell:
I actually agree about removing the Gate::resource section. To me that is just not a great method to use. 

However, your other two additions don't make sense to me. They seem to be stuck in random spots and don't provide much valuable information. I have removed those.

PR #5129, opened 2019-04-14
📝 Authentication Guard Customization docs added

williamcruzme:
The `guard()` function in "Authentication Guard Customization" is not documented.


taylorotwell:
Please re-submit with fully qualified class name on return.

PR #5138, opened 2019-04-19
[5.8] Add section for queue workers in config file

intellow:
I spent a long time not using Horizon in staging because I couldn't figure out why it wouldn't work there (answer - because the worker needs to be defined by environment in the config file). I believe this change will help others who may encounter the same problem.


taylorotwell:
I've added a note in a slightly different location. Thank you for the suggestion.

PR #5139, opened 2019-04-19
[5.8] Add additional eager loading documentation

jasonlbeggs:
This PR adds documentation for the `$with` property and the `without` method.

Feel free to change any of the wording if necessary.

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -1018,7 +1018,7 @@ Sometimes you might want to always load a relationship when retrieving a model. 
         /**
          * Always load the related author when retrieving a book
          */
-        protected $with = ['author];
+        protected $with = ['author'];
         
         /**
          * Get the author that wrote the book.
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -1003,6 +1003,36 @@ You may not always need every column from the relationships you are retrieving. 
 
 > {note} When using this feature, you should always include the `id` column and any relevant foreign key columns in the list of columns you wish to retrieve.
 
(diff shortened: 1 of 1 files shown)


eloquent-relationships.md
```diff
@@ -1003,6 +1003,36 @@ You may not always need every column from the relationships you are retrieving.
 
 > {note} When using this feature, you should always include the `id` column and any relevant foreign key columns in the list of columns you wish to retrieve.
 
+#### Eager Loading By Default
+
+Sometimes you might want to always load a relationship when retrieving a model. For this, you can simply use the `$with` property on your model. Example:
```
svenluijten: I'd personally shy away from using the word "simply": https://jameshfisher.com/2017/02/22/dont-use-simply/

```suggestion
Sometimes you might want to always load a relationship when retrieving a model. For this, you can use the `$with` property on your model. Example:
```

staudenmeir:
We could also mention the `$withCount` property.

eloquent-relationships.md
```diff
@@ -1003,6 +1003,36 @@ You may not always need every column from the relationships you are retrieving.
...
+    namespace App;
+
+    use Illuminate\Database\Eloquent\Model;
+
+    class Book extends Model
+    {
+        /**
+         * Always load the related author when retrieving a book
+         */
+        protected $with = ['author];
```
intellow: Looks like you are missing an apostrophe. `['author]` should be `['author']`

jasonlbeggs: Good catch!

PR #5151, opened 2019-04-25
add syntax highlight

hsiaoyi0504:
It seems to me that we should utilize syntax highlight in some document like this.


driesvints:
This doesn't follows the conventions of indenting code blocks. The current blocks are fine.

PR #5152, opened 2019-04-25
[5.8] Document change in behaviour as routes are now throttled on a domain level

garygreen:
Since PR https://github.com/laravel/framework/pull/19807 which changed the throttlers request signature, routes are now throttled at a domain level and not on a per-route/group basis. This is a major breaking change and so I've been suggested by @driesvints to send a PR to the docs to explain this behaviour change.

I personally believe this change is unintentional, as the docs explain you can have different throttle settings per route, but it seems this is no longer the case so it should be documented.

The original PR also meant you can no longer have different throttle settings for [api/web requests](https://github.com/laravel/laravel/blob/v5.8.3/app/Http/Kernel.php#L41), which Laravel uses out the box.

See here for more info: https://github.com/laravel/framework/issues/28323

### DIFF
--- a/passport.md
+++ b/passport.md
@@ -374,7 +374,7 @@ If the user approves the authorization request, they will be redirected back to 
 
 This `/oauth/token` route will return a JSON response containing `access_token`, `refresh_token`, and `expires_in` attributes. The `expires_in` attribute contains the number of seconds until the access token expires.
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
Probably best to just attempt to fix this.

PR #5164, opened 2019-04-29
Wrong method name for the except example on Eloquent collection.

vtalbot:
The `except` example was using `only` instead of `except`.


driesvints:
Thanks!

PR #5171, opened 2019-05-02
[5.8] Document new ChromeDriver command

driesvints:
See https://github.com/laravel/dusk/releases/tag/v5.1.0

### DIFF
--- a/dusk.md
+++ b/dusk.md
@@ -67,7 +67,7 @@ If you had test failures the last time you ran the `dusk` command, you may save 
 
     php artisan dusk:fails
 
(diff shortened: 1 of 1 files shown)


dusk.md
```diff
@@ -67,6 +67,17 @@ If you had test failures the last time you ran the `dusk` command, you may save
 
     php artisan dusk:fails
 
+If you want to install a different version than the current provided ChromeDriver than you may install that binary with the `dusk:chrome-driver` command
```
staudenmeir:
```suggestion
If you want to install a different version than the current provided ChromeDriver, then you may install that binary with the `dusk:chrome-driver` command
```

PR #5172, opened 2019-05-03
Fixed typo

radarhere:
Aside from just being a typo, it's also inconsistent with how the same word is spelt in the next sentence.


driesvints:
Thanks!

PR #5179, opened 2019-05-06
[5.8] Spelling and grammar fixes

MrRio:
Changes:

- Docs ran through Grammarly and manually checked all the critical errors. There were many false positives due to the unique language needed to describe programming aspects which I've skipped.
- Remove ... from the end of code comments

Let me know if you want this against master instead.


ahinkle:
I would steer clear from changing any links (`<a name"... `) as they will break permanent html links. 

The `...` on the end of comments is pretty common across frameworks/applications as a helper for action that is about to take place.

ahinkle:
I would steer clear from changing any links (<a name="... ) as they will break bookmarked html links from blog posts, etc.

MrRio:
Cool, sorted that so we always keep the same URLs.

taylorotwell:
Sorry, I'm not sure a lot of these changes are necessary and it's a ton of files to review.

PR #5185, opened 2019-05-10
[5.8] Readability update

leecrosdale:
Added the word 'then' so that the sentences are less of a mouthful.


taylorotwell:
I'm not sure this is necessary?

PR #5197, opened 2019-05-17
[5.8]  Add list of middleware and their default names and groups

imliam:


driesvints:
Hmm, not sure it's wanted that we list these all here. We'd have to maintain the list every time something's added or removed and I don't feel it provides much value.

taylorotwell:

PR #5201, opened 2019-05-20
[5.8] Stripe and Braintree Documentation

andresayej:
Stripe and Braintree Documentation reflects the Blueprint in the migrations examples thus the examples are consistent with the default users migration that is created with a new Laravel 5.8 installation and when a new migration is created using the **php artisan make:migration** command


andresayej:
Thanks for merging.
Follow-up question:
I've noticed that Blueprint exists from Laravel 5.3 up to 5.8
Would it be helpful for you that i update the billing.md & braintree.md of the 5.3, 5.4, 5.5, 5.6 and 5.7 branches as well?

driesvints:
@andresayej I don't think that's really necessary.

PR #5202, opened 2019-05-20
[5.8] Add a note about orWheres canceling relationships' main constraints

stancl:
Resolves https://github.com/laravel/framework/issues/28486 by adding a note to Eloquent Relationships docs.

I hope the wording is good.

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -830,6 +830,8 @@ You may query the `posts` relationship and add additional constraints to the rel
 
 You are able to use any of the [query builder](/docs/{{version}}/queries) methods on the relationship, so be sure to explore the query builder documentation to learn about all of the methods that are available to you.
 
+> *Note:* An `->orWhere()` clause will be applied on the same level as the relationship's main constraint (e.g. `where user.id = 1`). If you want to avoid this, use [constraint groups](/docs/{{version}}/queries#parameter-grouping).
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not sure this is needed here.

stancl:
I disagree.

The following query should translate as "the authenticated user's tasks which are due today or have not yet been completed", since expressive syntax is the goal.

```php
auth()->user()->tasks()->where('date', $today)->orWhere('completed', false);
```

The fact that this is the way such queries are expected to work can be proven by the amount of issues that have been created about it.
* https://github.com/laravel/framework/issues/28486
* https://github.com/laravel/framework/issues/28207
* https://github.com/laravel/framework/issues/4702
* https://github.com/laravel/framework/issues/20033
* https://github.com/laravel/framework/issues/27591

The first part of the query, `auth()->user()->tasks()`, explicitly says that only the currently authenticated user's tasks will be returned. That should be an unbreakable promise.

The following query returning **other users' data** looks like a security issue.

```php
$tasks = auth()->user()->tasks(); // authenticated user's tasks
return $tasks->where('date', $today)->orWhere('completed', false); // return only the tasks that are due today or have not been completed
```

This behavior is unexpected and should be documented.

taylorotwell:
OK will re-review.

taylorotwell:
Fleshed this out into a fuller example.

stancl:
Awesome, thanks!

PR #5204, opened 2019-05-22
Document array parameter for groupBy

gauravmak:
The developer can also pass an array as the first parameter of the `groupBy` method to group items by multiple keys. This PR adds a note to the documentation for the same.


taylorotwell:
Your example passes a string, not an array.

PR #5205, opened 2019-05-24
Remove line about how 0 removes entries when using the Cache Facade

dominikb:
The handling of a duration of 0 seconds is vastly different among storage implementations.

These drivers store values forever:
- [ApcStore](https://github.com/laravel/framework/blob/bf7076f9a325a65770aa802d6014993422eb8f6c/src/Illuminate/Cache/ApcStore.php#L97)
- [ArrayStore](https://github.com/laravel/framework/blob/bf7076f9a325a65770aa802d6014993422eb8f6c/src/Illuminate/Cache/ArrayStore.php#L102)
- [FileStore](https://github.com/laravel/framework/blob/bf7076f9a325a65770aa802d6014993422eb8f6c/src/Illuminate/Cache/FileStore.php#L121)
- [MemcachedStore](https://github.com/laravel/framework/blob/bf7076f9a325a65770aa802d6014993422eb8f6c/src/Illuminate/Cache/MemcachedStore.php#L178)

These stores actually expire the value immediately:
- [DatabaseStore](https://github.com/laravel/framework/blob/bf7076f9a325a65770aa802d6014993422eb8f6c/src/Illuminate/Cache/DatabaseStore.php#L100)
- [DynamoDbStore](https://github.com/laravel/framework/blob/bf7076f9a325a65770aa802d6014993422eb8f6c/src/Illuminate/Cache/DynamoDbStore.php#L459)
- [RedisStore](https://github.com/laravel/framework/blob/bf7076f9a325a65770aa802d6014993422eb8f6c/src/Illuminate/Cache/RedisStore.php#L93) (value will be cached for 1 second)

### DIFF
--- a/cache.md
+++ b/cache.md
@@ -182,6 +182,8 @@ The `add` method will only add the item to the cache if it does not already exis
 
     Cache::add('key', 'value', $seconds);
 
+> {note} Using a duration of 0 seconds is handled differently by the underlying cache stores and does not immediately expire the stored value.
+
 #### Storing Items Forever
 
 The `forever` method may be used to store an item in the cache permanently. Since these items will not expire, they must be manually removed from the cache using the `forget` method:
@@ -197,9 +199,7 @@ You may remove items from the cache using the `forget` method:
 
     Cache::forget('key');
 
-You may also remove items by providing a zero or negative TTL:
-
-    Cache::put('key', 'value', 0);
+You may also remove items by providing a negative TTL:
 
     Cache::put('key', 'value', -5);
 


taylorotwell:
Then it should be made consistent in the framework most likely. /cc @driesvints

driesvints:
This is on purpose. The `Illuminate\Cache\Repository` is a PSR-16 implementation and will handle the 0 values correctly but the individual stores are just implementations of the drivers and they're not individual PSR-16 implementations. 

The reason why the stores themselves aren't PSR-16 implementations is to let them be used as close to their native behavior as possible should anyone want to use them directly. If you need a PSR-16 implementation you should use the `Repository` with a driver for whatever storage you're using. If you want the native behavior of a driver then you can use the store directly.

The fact that "The handling of a duration of 0 seconds is vastly different among storage implementations." is simple because each storage has a different native behavior.

PR #5211, opened 2019-05-31
[5.8] Updated telescope.md

rockfreak:
Updated the watchers to show the snippet for each one. Also updated default options to match default install.

Added a new section to list out all the environment variables you can use as I didn't feel it was obvious when I was looking for such information.


taylorotwell:
I feel like this is all pretty standard information anyone can learn by reading the configuration file.

PR #5214, opened 2019-05-31
[5.8] Correct number of method

HiroKws:
There are eight methods in following codes. However explain "seven" method to implement.

### DIFF
--- a/scout.md
+++ b/scout.md
@@ -367,7 +367,7 @@ If you need to customize the search behavior of an engine you may pass a callbac
 
 #### Writing The Engine
 
-If one of the built-in Scout search engines doesn't fit your needs, you may write your own custom engine and register it with Scout. Your engine should extend the `Laravel\Scout\Engines\Engine` abstract class. This abstract class contains seven methods your custom engine must implement:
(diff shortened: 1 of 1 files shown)


scout.md
```diff
@@ -367,7 +367,7 @@ If you need to customize the search behavior of an engine you may pass a callbac
 
 #### Writing The Engine
 
-If one of the built-in Scout search engines doesn't fit your needs, you may write your own custom engine and register it with Scout. Your engine should extend the `Laravel\Scout\Engines\Engine` abstract class. This abstract class contains seven methods your custom engine must implement:
+If one of the built-in Scout search engines doesn't fit your needs, you may write your own custom engine and register it with Scout. Your engine should extend the `Laravel\Scout\Engines\Engine` abstract class. This abstract class contains eight methods your custom engine must implement:
```
driesvints: Rather than a fixed number which we need to update all the time let's use some relative wording for this:

```suggestion
If one of the built-in Scout search engines doesn't fit your needs, you may write your own custom engine and register it with Scout. Your engine should extend the `Laravel\Scout\Engines\Engine` abstract class. This abstract class contains the following methods your custom engine must implement:
```

PR #5217, opened 2019-06-04
Fix ConnectionInterface file path

misagh:
Laravel 5.7 Upgrade doc:

https://laravel.com/docs/5.7/upgrade

The "ConnectionInterface" file is in the "Illuminate\Database" path.


driesvints:
Thanks!

PR #5219, opened 2019-06-04
[5.8] Mention the tinker problem with Job::dispatch()

mikew1:
I saw a previous PR on this which has alternate (maybe better) wording than mine.

https://github.com/laravel/docs/pull/4715

This is the second time I've wasted time on this same issue.

More evidence of the same problem here

https://github.com/laravel/framework/issues/23167

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -53,6 +53,8 @@ Some applications may not need to ever push jobs onto multiple queues, instead p
 
     php artisan queue:work --queue=high,default
 
+> {note} When manually pushing jobs to queue from tinker, you should use `Bus::dispatch(new TestJob())` or `Queue::push(new TestJob())`, because with `Job::dispatch()`, the first job dispatched upon opening tinker will always be lost. This is due to an implementation subtlety involving garbage collection and applies only to tinker.
+
 <a name="driver-prerequisites"></a>
 ### Driver Notes & Prerequisites
 


driesvints:
I don't believe this is the correct place for this and that this belongs to the Tinker docs since it's Tinker specific.

mikew1:
I'd love to know what developers better than I am are doing in that case.  My process is

 If I want a queue, I read the queue docs and follow the steps, which leads the reader directly to using Job::dispatch().  

At that point, I simply *assume* that I will try out that command in tinker, as I normally do with any new command before building code, and that it would therefore work there.

So, your response here leads me to inevitably question my process as a developer.  
What am I doing wrong in the above two steps?

driesvints:
@mikew1 In my eyes I'd read the docs of Tinker first before using it. If you start off using a tool and using it in a way you *think* it works before reading how it *actually* works then that's a bit contra-productive imo. This goes for all software out there.

mikew1:
The comment concerned was absent from the tinker docs until [November 2018](https://github.com/laravel/tinker/commit/1baadfe1721f85cd4e776aa323128dcd329f170d#diff-04c6e90faac2675aa89e2176d2eec7d8).

mikew1:
If I did only laravel I could cope with this, but it slows things down for people who don't necessarily remember everything and rely on the documentation to get them up to speed when they repeat somethign they did a while ago.   have a nice day...  got work to do...

taylorotwell:
I agree this is more of a Tinker documentation thing.

ryancwalsh:
Wow, I really would have appreciated that proposed note in the Queue docs. I never would have thought to check Tinker. This violates the Principle of Least Surprise.

slavicd:
> If you start off using a tool and using it in a way you think it works before reading how it actually works then that's a bit contra-productive imo

Spent 1 hour debugging this. How is describing the internals of how `dispatch` method works relevant to Tinker? Reading how Job classes work is the actual intention of a coder affected by this; and yet the only mention about garbage collection being involved in on Tinker section.

Please put the info in the Queue section too.

verybigelephants:
I understand that this is an unresolvable issue because of nature of how things work; but considering how Tinker is tightly interconnected with Laravel framework and [even mentioned in the documentation](https://laravel.com/docs/12.x/artisan#tinker) this should definitely be mentioned [here](https://laravel.com/docs/12.x/queues)

This information is so non-existant in any docs that even when prompting an LLM about the specific configuration and problem about Laravel jobs not being dispatched in Tinker it doesn't answer why is it so but gives suggestions about misconfiguration.

PR #5230, opened 2019-06-11
[5.8] Add docs for eager typed morph relationships

brendt:
Addressing the changes made in https://github.com/laravel/framework/pull/28647

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -1109,9 +1109,9 @@ To load a relationship only when it has not already been loaded, use the `loadMi
 
 #### Nested Lazy Eager Loading & `morphTo`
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't think it makes sense to document morphWith here. It's not lazy eager loading and this entire section is about lazy eager loading?

brendt:
I'm not sure what the original idea of this section was, but it's the only place where `loadMorph` is documented, and this feature is highly related to it.

Any suggestions on how I can improve this? Should I make a separate section documenting `morphWith`, should I rename the current section and keep the two concepts together?

taylorotwell:
Added this to a different section.

PR #5235, opened 2019-06-13
[5.4] Add method name change to upgrade guide

bpedroza:


taylorotwell:
Not in the correct section.

bpedroza:
@taylorotwell Where do you want it? I just put it here as the section was already there and copied the verbiage from the BelongsToMany section directly above.

Happy to make any corrections and resubmit.

PR #5237, opened 2019-06-15
Add docs that there exists a `ImplicitRule` interface

mfn:
Which is required if you want your rule object validation also to run on empty attributes.


mfn:
Because I think reads should be made more aware of the "implicit extensions" beheaviour I decided to promote the chapter so it gets listed in the ToC

PR #5242, opened 2019-06-18
[Dusk] Add note about registering `service provider`

atefBB:
Add note about registering the `service provider` after adding `Browser macro`.


taylorotwell:
I think this can be determined from the linked service provider documentation.

PR #5263, opened 2019-06-25
[5.8] Update the major version of pusher-php-server

WillSewell:
A new major version has been released: https://github.com/pusher/pusher-http-php/pull/222.


driesvints:
@WillSewell have you tested the new version with Laravel's broadcasting system already? Afaik there don't seem to be any major breaking changes.

WillSewell:
@driesvints I didn't test it, but that's because the only breaking change in version 4.0 of pusher-php-server was to drop support for older versions of PHP.

PR #5266, opened 2019-06-27
[5.8] Documenting `--parent` artisan make:controller option

itainathaniel:
Two years ago https://github.com/laravel/framework/pull/18606 was merged to the framework without a documentation.

### DIFF
--- a/controllers.md
+++ b/controllers.md
@@ -176,6 +176,12 @@ If you are using route model binding and would like the resource controller's me
 
     php artisan make:controller PhotoController --resource --model=Photo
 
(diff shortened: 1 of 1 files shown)


ahinkle:
You can use `php artisan help make:controller` to see the full command documentation and options.

itainathaniel:
@ahinkle I know.
But the fact that something is explained on the help of the command doesn't exclude it from being on the official Laravel documentation

taylorotwell:
Capitalization and grammar does not match rest of the framework.

PR #5275, opened 2019-07-02
[5.8] Add note about complex Pivot classes

driesvints:
Add a note about complex pivot classes and the best practice to migrate them to an Eloquent model class. 

See https://github.com/laravel/framework/issues/29017


taylorotwell:
They can have all the normal relationships an Eloquent model can have, you just may need to be explicit about the keys because of the key guessing logic.

PR #5277, opened 2019-07-03
[5.8] Correct middleware terminate method description.

mvestil:
As opposed to https://github.com/laravel/docs/commit/984d7851832261261f3fa05badc416ed988a27cf

The previous version is actually true when i tested using nginx 1.16.0 and php 7.1.29 with fastcgi enabled in a Vagrant environment, not php artisan serve.

I put a 5 seconds sleep inside the terminate and added a log.
Made a request to the server via AJAX, got a JSON response from the server, after 5 seconds, my log was logged in the laravel log file.

### DIFF
--- a/middleware.md
+++ b/middleware.md
@@ -252,7 +252,7 @@ Middleware parameters may be specified when defining the route by separating the
 
 Sometimes a middleware may need to do some work after the HTTP response has been sent to the browser. For example, the "session" middleware included with Laravel writes the session data to storage after the response has been sent to the browser. If you define a `terminate` method on your middleware, it will automatically be called after the response is sent to the browser.
 
-> **Note:** If your web server is not configured to use FastCGI, the `terminate` method on your middleware will be called after the response is prepared to be sent to the browser.
+> **Note:** If your web server is not configured to use FastCGI, the request will run the `terminate` method on your middleware before the HTTP response is sent to the browser. This may result in an increased HTTP response time as it will wait for the code to complete its processing in the `terminate` method.
 
     <?php
 
--- a/middleware.md
+++ b/middleware.md
@@ -250,7 +250,9 @@ Middleware parameters may be specified when defining the route by separating the
 <a name="terminable-middleware"></a>
 ## Terminable Middleware
 
-Sometimes a middleware may need to do some work after the HTTP response has been prepared. For example, the "session" middleware included with Laravel writes the session data to storage after the response has been fully prepared. If you define a `terminate` method on your middleware, it will automatically be called after the response is ready to be sent to the browser.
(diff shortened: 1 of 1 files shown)


mosufy:
Verified the same on Postman with vagrant and FastCGI enabled and verify that the log was only written AFTER the api response has been sent back, and not affected by a 5s sleep imposed in the terminable middleware

taylorotwell:
I don't understand the wording. I thought if you're using FastCGI it will be run after the request is sent to the browser.

mosufy:
Yes, if you are using FastCGI, it will automatically be run after the request is sent to the browser. However, this commit https://github.com/laravel/docs/commit/984d7851832261261f3fa05badc416ed988a27cf#diff-37ab2f15ff048f67959bcac0a6032f32 (merged as of Laravel 5.6 docs), was the misleading wording.

We were struggling to determine what was the intended outcome of using the terminable middleware. Whether it is run AFTER the request is sent OR the request WAITS for the terminable middleware to be completed PRIOR to the sending of the request.

If you are using FASTCGI, it will run AFTER the request is sent.
If you are not, it will wait for the terminable middleware to run, THEN request is sent.

This is a very important distinction to how terminable works and affects how one makes a decision if terminable middleware should be used or otherwise.

For example, should I intend to have a long process job to be completed AFTER request is sent (e.g; logging request and response to a log service provider that might have tcp overheads) AND I am NOT using FastCGI, I should put it into a queued job and not terminable middleware (since it will affect response time). Should I have used FastCGI, I can achieve using the same effect without putting it into a queue and using terminable middleware should suffice (since response time is not affected by this).

middleware.md
```diff
@@ -250,7 +250,9 @@ Middleware parameters may be specified when defining the route by separating the
 <a name="terminable-middleware"></a>
 ## Terminable Middleware
 
-Sometimes a middleware may need to do some work after the HTTP response has been prepared. For example, the "session" middleware included with Laravel writes the session data to storage after the response has been fully prepared. If you define a `terminate` method on your middleware, it will automatically be called after the response is ready to be sent to the browser.
+Sometimes a middleware may need to do some work after the HTTP response has been sent to the browser. For example, the "session" middleware included with Laravel writes the session data to storage after the response has been sent to the browser. If you define a `terminate` method on your middleware, it will automatically be called after the response is sent to the browser.
+
+> **Note:** If your web server is not configured to use FastCGI, the `terminate` method on your middleware will be called after the response is prepared to be sent to the browser.
```
mosufy:
```suggestion
> **Note:** If your web server is not configured to use FastCGI, the request will run the `terminate` method on your middleware before the HTTP response is sent to the browser. This may result in an increased HTTP response time as it will wait for the code to complete its processing in the `terminate` method.
```

PR #5286, opened 2019-07-07
[5.8] Note to update morphable_type columns when using Morph Map

BrandonSurowiec:
A common issue I see when implementing a "morph map" is that Laravel stops querying for the fully qualified name `App\Post` and only uses the morph map name `posts`. This has been the cause of some frustration as it makes various data in the database seem to "disappear" when querying for it.

I documented the requirement to swap the database's `{morphable}_type` columns with the "morph map" name instead of the fully qualified name when implementing a "morph map" on an existing application.

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -799,6 +799,8 @@ By default, Laravel will use the fully qualified class name to store the type of
 
 You may register the `morphMap` in the `boot` function of your `AppServiceProvider` or create a separate service provider if you wish.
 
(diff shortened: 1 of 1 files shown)


BrandonSurowiec:
I need to stop rewriting it, but here's another variation that could replace what I changed. Feel free to modify however you see fit:

```
Eloquent will no longer use fully qualified names like `App\Post` or `App\Comment` to query the database after setting up a "morph map". Be sure to create a migration to transform every morphable `*_type` column to use its newly mapped name, like `posts` or `comments`.
```

PR #5289, opened 2019-07-08
[5.8] Add hash in auth.php

kossa:
Like the auth.php file here https://github.com/laravel/laravel/blob/master/config/auth.php#L47


kossa:
I think we dont need the hash when using passport

PR #5294, opened 2019-07-10
Consistent type hinting in Eloquent relationships docs

danmichaelo:
Noticed that the docs used type hints some places, but not others, so I tried hinting the Eloquent Builder and MorphTo classes consistently throughout the file.


staudenmeir:
FYI: We removed the eager loading type hints: #5333

PR #5297, opened 2019-07-10
Update migrations.md

wolfpack4417:
I know this is minor, but this confused one of my junior devs when I told him to create a migration that drops a table. He put the dropTable in the down method instead of up because the docs said up was for adding to the database and he assumed down would be for removing it.


taylorotwell:
I don't think we need to change the current wording. IMO the new wording can be equally confusing.

PR #5300, opened 2019-07-11
[5.8] Update migrations.md

crazycodr:
`set()` is not implemented anymore in 5.8.11, At least i can't find the definition at all in the class file.


taylorotwell:
image

crazycodr:
Didn't realize it until now but someone locked the project into an old version because of an outdated dependency. Sorry for the disturbance!

PR #5302, opened 2019-07-12
[5.8] Add uses of dynamic where

kevinb1989:
Add use of dynamic where and its examples.


taylorotwell:
I don't think I'm really going to document this. Just will leave it as a hidden gem, heh. I would technically prefer people use explicit wheres.

PR #5306, opened 2019-07-13
[5.8] Add info

SOHELAHMED7:
Added info that if `users` table is seeded with default  config, password is `secret`


taylorotwell:
Typos.

taylorotwell:
Also no need to make this a call-out.

PR #5315, opened 2019-07-20
Document array parameter for groupBy

gauravmak:
The developer can also pass an array as the first parameter of the groupBy method to group items by multiple keys. This PR adds a note to the documentation for the same.


taylorotwell:
I don't get it. An array with a single element is passed.

PR #5317, opened 2019-07-23
[5.8] Update validation docs to reduce confusion on weakly checked integer validation

mattwithoos:
specify that integer validation does not behave as expected and will not check type

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -830,7 +830,7 @@ The field under validation must exist in _anotherfield_'s values.
 <a name="rule-integer"></a>
 #### integer
 
-The field under validation must be an integer.
+The field under validation must be like an integer, but will not check the type.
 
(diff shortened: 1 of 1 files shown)


mattwithoos:
Many examples online of people saying "read the docs" but quickly being corrected. Here's one example:
https://stackoverflow.com/a/47954416/2985968

omarjebari:
Modifying the documentation isn't exactly a fix for this issue. We absolutely need strict type checking.

DFoxinator:
@mattwithoos thanks for updating the docs to reflect this distinction. Like others have said, this really seems like a bug/incorrect behavior. Does anyone know if there's any update on getting it fixed or at least something for strict type checking? This issue pretty much renders using integer to check params in a json-based API useless.

PR #5319, opened 2019-07-23
Add tip to customise mail notification theme

troccoli:
This PR is to update the docs following the release of the feature to be able to choose which theme to use for a single Mail Notification (commit #29132 on laravel/framework)


taylorotwell:
Can we work this into the documentation instead of adding another tip? Thanks.

PR #5320, opened 2019-07-23
Add tip to customise mailable theme

troccoli:
Following PR #5319 I noticed that the fact that a theme can be assigned to a Mailable was not documented. Hence this PR


taylorotwell:
Can you work this into the flow of the documentation maybe instead of adding another tip? Thanks.

PR #5322, opened 2019-07-24
Typo in code snippet

swaplicado:
Name of route parameter in code snippet should be actually singularized.


swaplicado:
The code snippet is OK.

PR #5338, opened 2019-08-03
Add "visibility to permissions" mapping

tontonsb:
It was hard for me to find out how to specify permissions for nodes created via local driver. I propose adding this to docs as the expected syntax is far from obvious.


taylorotwell:
"the configuration file"

What configuration file?

tontonsb:
I assumed the file name is implied by the first statement in the configuration section. The very previous paragraph refers to it as "your configuration file". If explicit naming is preferred, I shall add it to both paragraphs then, just like the later driver paragraphs do.

PR #5341, opened 2019-08-03
[5.8] Add "visibility to permissions" mapping, with file named

tontonsb:
Documenting how to specify permissions for local driver. Same #5338 with name of the config file added in this and the previous paragraph.

### DIFF
--- a/filesystem.md
+++ b/filesystem.md
@@ -50,6 +50,23 @@ When using the `local` driver, all file operations are relative to the `root` di
 
     Storage::disk('local')->put('file.txt', 'Contents');
 
+The `public` [visibility](#file-visibility) translates to `0755` for directories and `0644` for files. You can override the mappings in the configuration file:
+
+    'someLocalDisk' => [
+        'driver'   => 'local',
+        'root'     => '/your/store',
+        'permissions' => [
+            'file' => [
+                'public'  => 0664,
+                'private' => 0600,
+            ],
+            'dir'  => [
+                'public'  => 0775,
+                'private' => 0700,
+            ],
+        ],
+    ],
+
 <a name="driver-prerequisites"></a>
 ### Driver Prerequisites
 
--- a/filesystem.md
+++ b/filesystem.md
@@ -53,15 +53,15 @@ When using the `local` driver, all file operations are relative to the `root` di
 The `public` [visibility](#file-visibility) translates to `0755` for directories and `0644` for files. You can override the mappings in the `filesystems.php` configuration file:
 
     'someLocalDisk' => [
-        'driver'   => 'local',
-        'root'     => '/your/store',
+        'driver' => 'local',
+        'root' => '/your/store',
         'permissions' => [
             'file' => [
-                'public'  => 0664,
+                'public' => 0664,
                 'private' => 0600,
             ],
-            'dir'  => [
-                'public'  => 0775,
+            'dir' => [
+                'public' => 0775,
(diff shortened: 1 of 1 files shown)


filesystem.md
```diff
@@ -46,10 +46,27 @@ Once a file has been stored and the symbolic link has been created, you can crea
...
 
-When using the `local` driver, all file operations are relative to the `root` directory defined in your configuration file. By default, this value is set to the `storage/app` directory. Therefore, the following method would store a file in `storage/app/file.txt`:
+When using the `local` driver, all file operations are relative to the `root` directory defined in your `filesystems.php` configuration file. By default, this value is set to the `storage/app` directory. Therefore, the following method would store a file in `storage/app/file.txt`:
 
     Storage::disk('local')->put('file.txt', 'Contents');
 
+The `public` [visibility](#file-visibility) translates to `0755` for directories and `0644` for files. You can override the mappings in the `filesystems.php` configuration file:
+
+    'someLocalDisk' => [
+        'driver'   => 'local',
```
driesvints:
```suggestion
        'driver' => 'local',
```

filesystem.md
```diff
@@ -46,10 +46,27 @@ Once a file has been stored and the symbolic link has been created, you can crea
...
+        'root'     => '/your/store',
```
driesvints:
```suggestion
        'root' => '/your/store',
```

filesystem.md
```diff
@@ -46,10 +46,27 @@ Once a file has been stored and the symbolic link has been created, you can crea
...
+        'permissions' => [
+            'file' => [
+                'public'  => 0664,
```
driesvints:
```suggestion
                'public' => 0664,
```

filesystem.md
```diff
@@ -46,10 +46,27 @@ Once a file has been stored and the symbolic link has been created, you can crea
...
+                'private' => 0600,
+            ],
+            'dir'  => [
+                'public'  => 0775,
```
driesvints:
```suggestion
                'public' => 0775,
```

filesystem.md
```diff
@@ -46,10 +46,27 @@ Once a file has been stored and the symbolic link has been created, you can crea
...
+
+    'someLocalDisk' => [
+        'driver'   => 'local',
+        'root'     => '/your/store',
+        'permissions' => [
+            'file' => [
+                'public'  => 0664,
+                'private' => 0600,
+            ],
+            'dir'  => [
```
driesvints:
```suggestion
            'dir' => [
```

tontonsb:
Some of the other drivers (ftp, rackspace) also have alignment spaces in their examples. Should those be removed?

driesvints:
@tontonsb yeah you can update them as well. Thanks!

tontonsb:
I am not a native speaker so I might be wrong here, but...

I think we have ended up with a misleading wording here. The word "default" seems to imply that the permission mode could be overriden on file by file basis. As far as I know, this is not the case. 

There is no way to specify explicit permissions via `Storage` or change the mappings once a disk is created (Flysytem's Local adapter only sets permissions in constructor, Laravel's FilesystemManager only creates each disk once). One can only select among `public` and `private` and the permission map will be the one that was in the config at the moment of disk adapter creation. Or the default one.

driesvints:
@tontonsb feel free to send in a follow up pr if you think that's necessary.

PR #5342, opened 2019-08-03
[5.8] Adds vendor:publish --tag=passport-config

peterfox:
I found trying to use envs for the public and private keys in the Laravel Passport package only worked once the config has been published. Before that I would get complaints from the framework that the keys could not be found.

Multiline ENVs are also supported by the dotenv library used by Laravel so I think it would be better to advertise that option rather than having to put it all on one line and inserting the new line characters.


peterfox:
🎉 thank you @taylorotwell @driesvints

PR #5356, opened 2019-08-07
typo

codefir:
Since the DOM element id is "card-holder-name" in your documentation, the getElementById should should select the same id (currently it's typo'ed to "cardholder-name"). This was wrongly imported from the Stripe's documentation.


driesvints:
Looks like something went wrong here with branching.

PR #5361, opened 2019-08-12
[5.8] Added a section on testing webhooks locally.

danmatthews:
Just thought this might benefit folk who aren't used to the tooling and process surrounding developing with Webhooks.

### DIFF
--- a/billing.md
+++ b/billing.md
@@ -30,7 +30,6 @@
     - [With Payment Method Up Front](#with-payment-method-up-front)
     - [Without Payment Method Up Front](#without-payment-method-up-front)
 - [Handling Stripe Webhooks](#handling-stripe-webhooks)
-    - [Testing Webhooks Locally](#testing-webhooks-locally)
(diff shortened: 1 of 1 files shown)


driesvints:
I'm not sure if a section here is warranted. I mean I get the why you would do it, just saying that I'm not sure it's the correct place.

danmatthews:
Where do you think would be better suited?

driesvints:
Maybe just point to Valet itself? I'm not sure that we should mention another tool if we already have one in the ecosystem.

danmatthews:
Maybe just a note somewhere then? "{note} You can test webhooks locally using Laravel Valet and the valet share command"?

driesvints:
@danmatthews that sounds good 👍

danmatthews:
Do you think it's worth removing the section and just placing a note, or keeping the section?

driesvints:
Just the note seems fine to me.

danmatthews:
Updated - let me know if you think it would be better served elsewhere in the document.

PR #5364, opened 2019-08-14
remove dead link to webhooks

clarkeash:
Webhooks are no longer under config there is a dedicated section for them: https://laravel.com/docs/5.8/billing#handling-stripe-webhooks


driesvints:
Thanks!

PR #5365, opened 2019-08-14
[Telescope] Add note about Dump Watcher using cache.

barryvdh:
Add note about requiring a persistent cache.


taylorotwell:
Now that we have that exception in place I'm not sure we need to clutter docs with this.

barryvdh:
Sure thing. Not sure what the target is for the master branch though, because that's listed as 3.x-dev?

PR #5367, opened 2019-08-15
[6.0] Categorized method list

DarkGhostHunter:
Rather than spitting out the methods in alphabetical order, a categorized list of methods based on their purpose allows the reader to **quickly point out what main functionality they need** (combine two arrays, remove an item) and check what methods are available to achieve that.

This could be also replicated into the 5.8 docs.


driesvints:
Can't we do this on 5.8 instead?

driesvints:
We always merge 5.8 changes into master so they'd get to the new docs as well eventually

taylorotwell:
No plans to change this. Makes it slower to find things IMO.

PR #5373, opened 2019-08-19
[Mail] text isn't necessary for markdown emails

zack6849:
It seems that markdown actually generates the text emails for you, I spent a good bit of time trying to figure out why something like the following wasn't working: https://gist.github.com/zack6849/7dcd8183b71868cf55d0c2b417399a63

Turns out passing a text view will just pass it through markdown again and give you html even for the plaintext view, oops.

### DIFF
--- a/mail.md
+++ b/mail.md
@@ -183,6 +183,8 @@ If you would like to define a plain-text version of your email, you may use the 
                     ->text('emails.orders.shipped_plain');
     }
 
+This is __not necessary__ with markdown emails, as markdown will render plain-text emails for you.
+
 <a name="view-data"></a>
 ### View Data
 


taylorotwell:
We already state this in the opening introduction to Markdown mail.

zack6849:
@taylorotwell , I can appreciate that you already it here:
```
HTML templates for the messages while also automatically generating a plain-text counterpart.
```

I still feel like it should be noted somewhere that any attempt to define a plaintext version of the template will in fact result in HTML in your plaintext template, as it stands the wording wouldn't make me think that If i set one it would **break things**, just that it'd work if i didn't explicitly set one.

I lost a good hour to this issue this morning looking for a cause before I guessed because I assumed anything related to this would be in the section on plaintext emails, not the markdown emails.

PR #5379, opened 2019-08-20
[5.8] Mentioning mass assignment protection is disabled when using a factory

gpressutto5:
I think it's worth mentioning as in [seeding](https://github.com/laravel/docs/blob/5.8/seeding.md#L23).
I had a problem with this and had to read the source code to find out what was causing it.

### DIFF
--- a/database-testing.md
+++ b/database-testing.md
@@ -173,6 +173,8 @@ If you would like to override some of the default values of your models, you may
         'name' => 'Abigail',
     ]);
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
What problem did you have?

gpressutto5:
@taylorotwell I was creating a new record on table B whenever I created a new record on table A using an Observer. The problem is that, since the static property `$unguarded` is defined on `GuardsAttributes`, it unguards model `B` too when using a factory to create a new record on table A. I don't know if that's a bug, but I couldn't find any info about this side effect.

PR #5383, opened 2019-08-21
Explain array syntax for validation rules

MarceauKa:
I see a lot of people using only the `|` syntax for their rules because the array syntax is not well-documented (just in comment of the regex and unique rules).


taylorotwell:
Added this to 6.0 docs.

PR #5385, opened 2019-08-21
Add note for HTTP test route calls

AbdelkaderBah:
Quote: "This is expected behavior. You can only call a route once per test."

Source: https://github.com/laravel/framework/issues/29673#issuecomment-523352043


taylorotwell:
I'm not sure that is entirely accurate.

taylorotwell:
I know for a fact I call routes multiple times in a single test.

PR #5390, opened 2019-08-22
[5.8] Add insertOrIgnore method in Query Builder

kennith:
Reference https://github.com/laravel/framework/pull/29639

### DIFF
--- a/queries.md
+++ b/queries.md
@@ -628,7 +628,7 @@ You may even insert several records into the table with a single call to `insert
 
 #### Insert Or Ignore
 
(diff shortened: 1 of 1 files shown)


queries.md
```diff
@@ -626,6 +626,16 @@ You may even insert several records into the table with a single call to `insert
         ['email' => 'dayle@example.com', 'votes' => 0]
     ]);
 
+#### Insert Or Ignore
+
+You may use `insertOrIgnore` to allow duplicate entry to fail silently. In this case, the table `users` already has a record with unique `id` 1. The `insertOrIgnore` method will ignore duplicate entries and only insert entries that does not violate the integrity constraint.
```
staudenmeir:
```suggestion
You may use `insertOrIgnore` to allow duplicate entry to fail silently. In this case, the table `users` already has a record with the unique `id` 1. The `insertOrIgnore` method will ignore duplicate entries and only insert entries that do not violate the integrity constraint. 
```

PR #5391, opened 2019-08-23
[5.8] Added min and max processes to horizon page

tomswinkels:
We have fix the min and max processes, very nice to have this in the docs! :)


taylorotwell:
Unclear what you mean by the "min" strategy and "max" strategy.

PR #5395, opened 2019-08-23
[5.8] Add seeding test data for test cases

kevinb1989:
I found this super helpful in heaps of scenarios when I have to seed the status/type tables prior to running a test case. Some examples include order_statuses, item_statuses, product_types, book_genres... So maybe we should mention this in the docs.


driesvints:
@kevinb1989 as someone who has used seeds in tests extensively I can wholeheartedly recommend not to use them in tests. If you use them in tests you lock all the tests where you use them on the seed data of a single point. Using factories per-test is a better (and more isolated) way to go.

PR #5396, opened 2019-08-24
added documentation to database testing

yihyang:
Explained the need of using a closure instead of array when the factory need to call another factory in state.

Refer to discussions here: https://github.com/laravel/framework/issues/29709


taylorotwell:
Some grammatical and formatting problems here. Perhaps someone else can pitch in and help?

PR #5404, opened 2019-08-28
[5.8] Fix example. Remove last ',' in temporaryUrl

ingspadavecchia:
The second example of the temporaryUrl is not working because there is an extra ',' at the end of the last argument.


taylorotwell:
Technically valid on PHP 7.3 :)

PR #5406, opened 2019-08-28
[5.5] Document add deprecated predis

YoshinoriN:
### Description

Add [predis deprecated](https://github.com/laravel/framework/pull/29688) document to `v5.5`.

### Reason

The `v5.5` is not yet EOL. IMO should add a document for users notice.

PR #5407, opened 2019-08-28
[5.8] Document add deprecated predis

YoshinoriN:
Same with #5406


taylorotwell:
This should be present on the 6.0 documentation.

PR #5412, opened 2019-08-30
Update passport.md to point to proper Axios url

reed-jones:
mzabriskie/axios just redirects to axios/axios


taylorotwell:
Let's fix this on 6.0 branch.

PR #5420, opened 2019-08-31
Add an example of option shortcut with value

imaple7:
- Added an example of option shortcut with value

Currently, the artisan console options docs only have example of regular option. The usage of option shortcut with value is different from the regular option.

```
email:send {user} {--Q|queue=}

$ php artisan email:send 1 -Q default
$ php artisan email:send 1 -queue=default
```


taylorotwell:
I think that is the same syntax as usual?

PR #5428, opened 2019-09-02
Add named anchors to reference specific steps

jasonmccreary:
Having these additional anchors helps link directly to the upgrade steps in posts, tweets, etc.


jasonmccreary:
May want to add more post release, but these were a few I already was linking.

PR #5431, opened 2019-09-02
[6.0] Spacing

u01jmg3:
In case you are interested in being consistent with your spacing for Blade control structures, this PR does just that. The convention appears to be to have no space.


taylorotwell:
Looking at my own applications it looks like I use a space, so that should be the default convention.

PR #5432, opened 2019-09-02
[6.0] Spacing

u01jmg3:
Related to #5431

---

> ___taylorotwell:___
> _Looking at my own applications it looks like I use a space, so that should be the default convention._

Blade control spacing is now consistent across the docs and uses a single space.


driesvints:
erhm.. you don't want to do this for everthing (like yield and section). I think this is mostly for if statements and foreach statements.

u01jmg3:
Shall resubmit

u01jmg3:
Actually, why would some have and others not? Just trying to ascertain the rules behind your convention.

PR #5433, opened 2019-09-02
[6.0] Instructions for compiling assets being optional

jimgwhit:
The docs for version 6.0 has no indication that compiling assets is optional. Even version 5.8 had the command:
```
php artisan preset none
```
So there should be at least some minimum instructions as to how to use bootstrap.css in a freshly scaffolded authentication.

Such as:

> If you aren't compiling assets, you can manually download bootstrap.css and put in your assets css folder.  And in views\layouts\app.blade.php replace:
```ruby
<link href="{{ asset('css/app.css') }}" rel="stylesheet">
// with
<link href="{{ asset('css/bootstrap.css') }}" rel="stylesheet">
```
Or rename bootstrap.css to ap.css.


driesvints:
Seems something has gone with branching here

PR #5440, opened 2019-09-03
[6.x] Update authentication.md

jimgwhit:
Manually using bootstrap.css if not compiling assets


taylorotwell:
I don't think we need to document this.

PR #5441, opened 2019-09-03
[5.8] Update mail.md

iwasherefirst2:
Mention that user should put translatable text in the build method and not the in the constructor to prevent issues.


taylorotwell:
This is sorta unclear I think? Even I'm not sure what it means.

PR #5444, opened 2019-09-05
[5.6] Improve note for onOneServer()

ikari7789:
The "command" used for generating the SHA1 to ensure a task is run on one server in a group, if an artisan command, will include the full path to the PHP executable. If there is a difference in the install location here, even if using a centralized memcached or redis server, the task will be run twice as a different SHA1 hash will be generated.

### DIFF
--- a/releases.md
+++ b/releases.md
@@ -54,7 +54,7 @@ In addition, it is now easier to customize existing log channels using the loggi
 
 ### Single Server Task Scheduling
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
We may want to send this to a more modern version of the docs... 6.0 branch perhaps.

ikari7789:
I wasn’t sure as it applies even to 5.6 since it was introduced. I ran into this issue. Wasn’t sure how the docs were handled and whether or not the commit could be cherry picked through to the newer versions or not.

Will re-make the PR against 6.0 though.

PR #5445, opened 2019-09-05
[6.x] Improve note for onOneServer

ikari7789:
The "command" used for generating the SHA1 to ensure a task is run on one server in a group, if an artisan command, will include the full path to the PHP executable. If there is a difference in the install location here, even if using a centralized memcached or redis server, the task will be run twice as a different SHA1 hash will be generated.

Note applies to all versions since 5.6. This note seems to have not changed text at all.


taylorotwell:
We should maybe try to fix that in the framework so it isn't so dependent on install location of PHP rather than adding a note to the documentation. Are you able to do that?

PR #5446, opened 2019-09-05
Update upgrade.md

kreitje:
Add a note about the REDIS_CLIENT environment variable and how the default redis client changed from predis to phpredis.

In the repo, it's a bit hard to tell if it should be phpredis or PhpRedis.


Arkanius:
Is this a real BC? If so, IMO it isn't about a "Medium" impact.

But following the discussions at https://github.com/laravel/framework/issues/29864 I really don't think about it as a BC

kreitje:
Yep... You have to update config/database.php to 'break'.  I will close this out.

PR #5447, opened 2019-09-05
[6.x] Remove upgrade guide section on RegisterController

bakerkretzmar:
This PR updates the 6.0 upgrade guide to reflect laravel/framework#29874, by removing the instructions that suggested calling `parent::register` and `parent::registered`, which would cause an error.

Because lots of people will have already upgraded and followed those instructions, I moved that paragraph to the bottom and rephrased it to note the changes being made here. Not sure if that's the best way to do this so that people see it. Thanks!

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -14,7 +14,6 @@
 ## Medium Impact Changes
 
 <div class="content-list" markdown="1">
-- [Authentication `RegisterController`](#the-register-controller)
 - [Carbon 1.x No Longer Supported](#carbon-support)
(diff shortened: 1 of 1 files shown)


taylorotwell:
Do not move it to the bottom.

taylorotwell:
I don't think I'm going to revert this after all. I will just adjust the upgrade guide.

taylorotwell:
Eh, nevermind, I think you are right.

taylorotwell:
I just removed this section entirely.

PR #5449, opened 2019-09-05
[6.x] Correcting comma and period placements when used around quotes.

jeffwray:
Commas and periods should be placed inside quotes instead of outside. This PR addresses these grammatical conventions without touching code examples where commas would be out of the quotes.


taylorotwell:
That's mainly a US thing I think. Even though I live in the US I don't really like it, heh.

PR #5450, opened 2019-09-05
[6.x] Document new behavior on serialization failure

jansennerd10:
This PR documents the changes made in https://github.com/laravel/framework/pull/29067.

I'm not sure how you guys assess the Likelihood of Impact. I chose Medium because the `DB::transaction()` method is commonly used but it seems that the relevant transaction modes may be rare. If it's incorrect, feel free to change it.


taylorotwell:
Since there is no actual "upgrade step" here, I don't see a need to include it in the upgrade guide.

fitztrev:
The PR says it was a breaking change. If that's true, where should breaking changes be documented?

PR #5456, opened 2019-09-06
[6.x] Added note about npm install on Windows machines

Halfwhit:


taylorotwell:
Maybe we should note this somewhere else? Maybe in the output of the UI command? Do the NPM documentations cover this? It feels out of context here because there is no mention of even running NPM here.

svpernova09:
We're almost to the point of needing some kind of FAQ / Common issues section. Virtualbox especially feels like it's in a tailspin of dumpster fires lately.

power777:
This issue is not specific to homestead.

PR #5459, opened 2019-09-09
Change link for Date class documentation from es to en

agasigp:
Change link for Date class documentation from es to en


driesvints:
Thanks!

PR #5461, opened 2019-09-10
Wrong TTL unit variable naming

dael:
second parameter of the Cache::remember method receives minutes, not seconds.


driesvints:
Seconds is correct.

PR #5462, opened 2019-09-10
Typo fix

fridzema:


driesvints:
The current value is correct. phpredis is the new default so if you need to keep using predis you should set that as the default.

PR #5466, opened 2019-09-11
[6.x] add notes about default Expressions

browner12:
explain how `default` modifier can take either a value or an `Illuminate\Database\Query\Expression` and provide an example.

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -307,31 +307,29 @@ The `default` modifier can take either a value or an `\Illuminate\Database\Query
 
 One place this is particularly useful is assigning default values to JSON columns:
 
-```php
-<?php
+    <?php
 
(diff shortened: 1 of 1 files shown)


migrations.md
```diff
@@ -301,6 +301,40 @@ Modifier  |  Description
...
+    public function up()
+    {
+        Schema::create('flights', function (Blueprint $table) {
+            $table->bigIncrements('id');
+            $table->json('movies')->default(new Expression('(JSON_ARRAY())'));
+            $table->timestamps();
+        });
+    }
+}
+```
```
driesvints: You probably want to indent the code block here like the other examples. instead of using the ``` signs.

PR #5468, opened 2019-09-12
[6.x] Documents how to ignore a resource at the unique rule

Arkanius:
I had some troubles to understand at the docs how to ignore a column with the unique clause. Actually, in the docs the only way to do this is by using the **Rule** facade.

This PR creates an example of how to ignore some id by using the simple string syntax. Also, it documents how to use the argument *(idColumn)* at unique rule.


taylorotwell:
You should not use the string approach for this rule. Let's document the methods instead.

PR #5470, opened 2019-09-13
[6.x] Update installation.md

timgreenwood:
Updated Composer's system-wide vendor bin directory for updated Laravel installer from $HOME/.composer/vendor/bin to $HOME/.config/composer/vendor/bin


driesvints:
My composer isn't located here but maybe the default got changed? Can you link to any Composer documentation that proves that this is the new default location?

timgreenwood:
https://getcomposer.org/doc/03-cli.md#composer-home

I think it is due to this line:

driesvints:
Which line is that?

timgreenwood:
> Which line is that?

Sorry, I hit cmd+return by accident - reply updated

PR #5473, opened 2019-09-13
Fixed typo in MySQL version - 5.7.7 to 5.7.27

fylzero:
I think there is a typo here.  MySQL 5.7.7 is not a current version of MySQL.  I'm assuming this was supposed to be 5.7.27.


chelout:
It's not a typo. Read the context.

fylzero:
My bad, confused the version scheme.

PR #5476, opened 2019-09-16
Adds chrome version number to artisan command

dellow:
Github Actions has specific software in virtual environments: https://help.github.com/en/articles/software-in-virtual-environments-for-github-actions#ubuntu-1804-lts

Currently the version of Chrome used in Actions is 76. Now the version of chrome available is 77 you need to specify the version number.


taylorotwell:
We would have to keep this updated on every release and it could get out-dated? Is there a better way?

PR #5479, opened 2019-09-18
Removed type-hinting for explicit model binding

habeebdev:
type-hinting not required for variable names in explicit model binding


taylorotwell:
It's fine.

PR #5484, opened 2019-09-20
Supervisor configuration details on horizon page

shreyanshpanchal:
There is a small guide about how to install & configure supervisor on Queue documentation page.
It would be nice if we can add the same guide on Horizon Page.

That's what i have done. :)


taylorotwell:
I don't think the program you are starting and the worker configuration match.

shreyanshpanchal:
@taylorotwell I'm so sorry, I've fixed it.

driesvints:
@shreyanshpanchal can you send this to the 6.x branch? Thanks

PR #5486, opened 2019-09-21
[6.x] Added password grant revoke documentation

benyou1969:
The Laravel Passport documentation doesn't include a code for revoking access tokens manually, which is needed when performing logout


taylorotwell:
We would need more documentation around when this actually is useful. It feels really out of context and unclear when you are supposed to use it.

PR #5488, opened 2019-09-24
[6.x] Fix Passport environment keys example

rogervila:
Hello,

As shown in the [original Pull Request example](https://github.com/laravel/passport/pull/683), environment keys for Passport need to be placed in one line.


taylorotwell:
I'm not sure this is true? /cc @GrahamCampbell

tdondich:
This absolutely is 100% true with the latest dotenv. @taylorotwell , please repoen?

PR #5491, opened 2019-09-25
[6.x] Added easy description of Pivot tables and utilization

cmg0x1e:
After trying to learn this concept properly myself for a few hours, I decided to write a small tl;dr about what Pivot tables are and a quick example on how to utilise the belongsToMany relationship.

I believe this will help readers better understand Pivot tables using a high-level concept of Users and Departments, and showing the benefit of a pivot table being the most practical option.


taylorotwell:

PR #5493, opened 2019-09-25
[6.x] Add assertJsonPath

ahinkle:
Adds the new validation rule, `assertJsonPath`.

https://github.com/laravel/framework/pull/29957

### DIFF
--- a/http-tests.md
+++ b/http-tests.md
@@ -192,6 +192,30 @@ If you would like to verify that the given array is an **exact** match for the J
         }
     }
 
+<a name="verifying-json-paths"></a>
+### Verifying JSON Paths
+
+If you would like to verify that the JSON response contains the given data at the specified path, you should use the `assertJsonPath` method:
+
+    <?php
+
+    class ExampleTest extends TestCase
+    {
+        /**
+         * A basic functional test example.
+         *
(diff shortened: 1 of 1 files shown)


taylorotwell:
Could you may flesh out the example a bit so it's clearer what you should be passing?

ahinkle:
Will do. To keep the Available Assertions section uniform, I've added an example within the Testing JSON APIs section. Let me know if I need to provide any additional examples.

dailytabs:
The documentation for this is pathetic. Standard json path notation starts with `$.` This is not supported and not documented. Array access is impossible to figure out. I just need to test that the first array item has a certain property and value.
`[0].name` is null
`$[0].name` is null
`$.[0].name` is null
`.name` is null
`..name` is null

I can dump my response and it's right there! 40 times in this array! Plain as day: `"name": "value"`

Like seriously, any explanation of WTF `$path` is would be...ya know, proper documentation!

ahinkle:
Feel free to open a new PR; suggesting that your entitled comment of "pathetic" doesn't help anyone.

driesvints:
@dailytabs can you please keep your comments more civil? Your remarks aren't helping anyone.

PR #5496, opened 2019-09-26
[6.x] Fixes grammar for Lazy Collection note

kayrunm:
Noticed a very tiny mistake in the grammar for the docs where it says "are are" for the following note:

> Methods that mutate the collection (such as `shift`, `pop`, `prepend` etc.) are are not available on the `LazyCollection` class.


driesvints:
Thanks!

PR #5500, opened 2019-09-28
Missing UploadController added in use statement in example

mitulgolakiya:
In the example of Contextual Binding, UploadController class was missing in the use parameter.


driesvints:
Thanks!

PR #5503, opened 2019-09-30
[6.x] Add support policies for databases

driesvints:
This adds version policies for the different database types and the minimum versions we support. I couldn't find a support policy for SQLite so I added the latest version (which was released in 2004). 

Also linked to the versions from the main support policy docs.


staudenmeir:
Laravel requires SQLite 3.8.8+: https://github.com/laravel/framework/pull/25995

I think we should also support older versions of SQL Server. A lot of people still use SQL Server 2012 or even 2008.

PR #5507, opened 2019-10-02
[6.x] Add Nexmo Shortcode usage

dwightwatson:
This just extends the Nexmo part of the docs indicating how to use Shortcodes.

Relevant: laravel/nexmo-notification-channel#24

### DIFF
--- a/notifications.md
+++ b/notifications.md
@@ -31,6 +31,7 @@
 - [SMS Notifications](#sms-notifications)
     - [Prerequisites](#sms-prerequisites)
     - [Formatting SMS Notifications](#formatting-sms-notifications)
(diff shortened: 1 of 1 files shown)


taylorotwell:
Is "alert", "2fa", and "marketing" predefined API values that Nexmo expects or are those user defined values based on what you setup in Nexmo?

dwightwatson:
Those are the 3 values that Nexmo supports for short codes.

If preferable we could add constants for them, or I could PR a simple class similar to the `NexmoMessage` to make building the array a little nicer?

```php
(new NexmoShortcode)
    ->alert()
    ->with(['code' => 'ABC123']);

PR #5510, opened 2019-10-03
[6.x] Show example for multiple middleware

lawsonjt:
The middleware method accepts a single string for one middleware, or an array of strings for a group of middleware.


driesvints:
I'm not 100% sure this works like this. I've seen issues with this syntax and it's not documented. It's also a bit weird since you're defining middleware up front without having a group created first.

lawsonjt:
In the past, this was the only way I've been able to use multiple middleware.  I agree that it's not documented!

PR #5513, opened 2019-10-04
[6.x] Zsh note

SketchNI:
Add note to homestead docs that the Bash function also works on ZSH.


mattmcdonald-uk:
@taylorotwell Starting with macOS Catalina [ZSH is the default shell](https://www.theregister.co.uk/2019/06/04/apple_zsh_macos_catalina_default/).

Would you accept a PR adding a ZSH note to the docs if better context was provided?

PR #5515, opened 2019-10-08
[6.x] Confirm password

driesvints:
Document new password rule and password confirmation functionality.

### DIFF
--- a/authentication.md
+++ b/authentication.md
@@ -208,6 +208,14 @@ When attaching the `auth` middleware to a route, you may also specify which guar
         $this->middleware('auth:api');
     }
 
+#### Confirming Password
+
+Laravel ships with a `password.confirm` middleware that allows you to protect routes with password confirmation. Adding the `password.confirm` middleware will first redirect users to a screen where they need to enter their password before they can continue. After they've successfully entered their password, they're redirected to the route they tried to access and their session is remembered for a default of three hours. This value can be changed with the `auth.password_timeout` config value.
+
+    Route::get('settings/security', function () {
+        // Users need to confirm their password before continuing...
+    })->middleware('password.confirm');
+
 <a name="login-throttling"></a>
 ### Login Throttling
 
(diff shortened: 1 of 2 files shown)


authentication.md
```diff
@@ -208,6 +208,14 @@ When attaching the `auth` middleware to a route, you may also specify which guar
         $this->middleware('auth:api');
     }
 
+#### Confirming Password
+
+Laravel ships with a `password.confirm` middleware that allows you to protect routes with password confirmation. Adding the `password.confirm` middleware will first redirect users to a screen where they need to enter their password before they can continue. After they've successfully entered their password, they're redirected to the route they tried to access and their session is remembered for a default of three hours. This value can be changed with the `auth.password_timeout` config value.
+
+    Route::get('settings/security', function () {
+        // Users need to confirm their password before continuing...
+    })->middleware('password.confirm');
```
arcanedev-maroc: I think you should mention `auth` middleware or specify it in the example:

```php
Route::get('settings/security', function () {
    // Users need to confirm their password before continuing...
})->middleware(['auth', 'password.confirm']);
```

driesvints: Done. Good call 👍

arcanedev-maroc:
I was playing with this new feature and i love it a lot :heart_eyes:

But there is one little thing that bothers me and i need a feedback about it. 

I've added the `password.confirm` middleware to the `/home` route.

Then when a user create an account, it redirects to the `/home` page. Password confirm shows up!!

Is it weird to keep it like that for an account created less than `auth.password_timeout` (3 days by default) or do i need to change the middleware to skip the password confirmation ?

driesvints:
@arcanedev-maroc this isn't typical something you'd want to add to your home route. Only to some sensitive ones. The password timeout is also three hours by default, not three days.

arcanedev-maroc:
Yeah, that make sense. 

Applying it only in the *danger zone* :eyes: 

Thanks :+1:

decadence:
> to a screen where they need to enter their password before they can continue

How does this screen look like and can we customize it?

driesvints:
@decadence yes. See https://github.com/laravel/ui/pull/34 and https://github.com/laravel/framework/pull/30214

PR #5517, opened 2019-10-09
[6.x] add a note on new 'confirm' middleware

browner12:
since we are redirecting, we can only send the user on to a `GET` route.

I would like to explore the ability to confirm `POST` routes, but with the current code, it is not possible.

### DIFF
--- a/authentication.md
+++ b/authentication.md
@@ -222,6 +222,8 @@ To accomplish this, Laravel provides a `password.confirm` middleware. Attaching 
 
 After the user has successfully confirmed their password, the user redirected to the route they originally tried to access. By default, after confirming their password, the user will not have to confirm their password again for three hours. You are free to customize the length of time before the user must re-confirm their password using the `auth.password_timeout` configuration option.
 
+> {note} The `password.confirm` middleware can only be applied to `GET` routes.
+
 <a name="login-throttling"></a>
 ### Login Throttling
 


taylorotwell:
I think this reads a little confusing. You definitely want to attach it to your POST routes that are protected because otherwise a user could craft a POST to skip the password confirm.

driesvints:
Yeah, on second thought I also agree with that. Even though form data won't be sent along, you'd never execute a post route first. You always want to visit a get route first to see your settings screen for example.

browner12:
ok, let me see if I follow you.

let's say we are creating a user, so our routes are:

```php
Route::get('users/create', 'UserController@create');
Route::post('users', 'UserController@store');
```

You are saying the `auth.password` middleware should be applied to both routes

```php
Route::get('users/create', 'UserController@create')->middleware(['auth.password'];
Route::post('users', 'UserController@store')->middleware(['auth.password'];
```

##### Scenario 1

The user visits the `/users/create` `GET` route where they will get redirected to the confirmation page to enter their password again, and upon success will get sent back to the intended `GET` route.  After successfully completing that form, it will move onto the `POST` route which will *also* check that the password has been confirmed. Ideally it will already be confirmed, and processing will proceed as normal.

##### Scenario 2

The user directly visits the `/users` `POST` route. The middleware will redirect them to the confirmation page to enter their password. Upon successfully submitting their password, it will **redirect** them to `/users`, but since it is a redirect it will be a `GET` request, which means they will probably end up back at the `UserController@index` method.

---

I think I'm in agreement with you we need to guard the `POST` route, otherwise someone could craft a way to skip over the check. However, I do think we need to make it a little clearer that this is not intended to be placed ***only*** on a `POST` route.

Let me see if I can try again and craft up something a little clearer.

PR #5521, opened 2019-10-10
[6.x] Job Middleware example, added missing return statement as per docblocks

pathak-vikash:
Job Middleware example, added missing return statement as perdocblocks. It's mentioned the return type as `@mixed` but missed to return from example.

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -222,16 +222,16 @@ Instead of rate limiting in the handle method, we could define a job middleware 
          */
         public function handle($job, $next)
         {
-            Redis::throttle('key')
+            return Redis::throttle('key')
                     ->block(0)->allow(1)->every(5)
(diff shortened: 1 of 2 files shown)


taylorotwell:
It is not necessary to return anything from job middleware.

pathak-vikash:
Ok, @taylorotwell , in such case, can we update the docblock to say something like:

```

/**
     * Process the queued job.
     *
     * @param  mixed  $job
     * @param  callable  $next
     * @return mixed | void  Not required to return value from Job Middleware
     */
 ```

thought?

pathak-vikash:
Sir, It seems, it may skipped... any thought on the comments?
https://github.com/laravel/docs/pull/5521#issuecomment-540831189

@taylorotwell

PR #5526, opened 2019-10-11
[6.x] Removed unwanted quotes

gurrrung:
removed unwanted quotes


gurrrung:
I'm confused, isn't this legit?

PR #5531, opened 2019-10-14
[6.x] Update upgrade guide to include HandlesAuthorization trait change

ankurk91:
Hi,

This PR adds instructions for a change in `Illuminate\Auth\Access\HandlesAuthorization` trait that was introduced in this pull request.
https://github.com/laravel/framework/pull/29117

#### Laravel 5.8 behaviour
The `deny()` method throws an exception and there is no need of `return` statement.
The return is useless here knowing that it will throw exception.
```php
    public function update(User $user, Post $post)
    {
        if($user->role !== 'editor') {
            // works without return
            $this->deny("You must be an editor to edit a post.")
        }
        
        return $user->id === $post->user_id;
    }
```
#### Laravel 6.0 behaviour
The `deny()` method no longer throws an exception, developer need to add a `return` statement in order to make the policy work.
Without the return statement, all of the `deny()` calls are being ignored.
```php
    public function update(User $user, Post $post)
    {
        if($user->role !== 'editor') {
           // need to add return
            return $this->deny("You must be an editor to edit a post.")
        }
        
        return $user->id === $post->user_id;
    }
```

Feel free to suggest edits.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -79,6 +79,22 @@ The constructor signature of the `Illuminate\Auth\Access\Response` class has cha
 
 The `Illuminate\Contracts\Auth\Access\Gate` contract has received a new `inspect` method. If you are implementing this interface manually, you should add this method to your implementation.
 
+#### The `Illuminate\Auth\Access\HandlesAuthorization` Trait
+
+**Likelihood Of Impact: Low**
+
+The `Illuminate\Auth\Access\HandlesAuthorization` trait's method `deny()` has been updated to return `\Illuminate\Auth\Access\Response`. Previously it would throw an exception. 
+You may need to update your policy classes to add a `return` statement like:
+
+    public function update(User $user, Post $post)
+    {
+        if($user->role !== 'editor') {
+            return $this->deny("You must be an editor to edit a post.")
+        }
+        
+        return $user->id === $post->user_id;
+    }
+    
 ### Carbon
 
 <a name="carbon-support"></a>


upgrade.md
```diff
@@ -79,6 +79,22 @@ The constructor signature of the `Illuminate\Auth\Access\Response` class has cha
 
 The `Illuminate\Contracts\Auth\Access\Gate` contract has received a new `inspect` method. If you are implementing this interface manually, you should add this method to your implementation.
 
+#### The `Illuminate\Auth\Access\HandlesAuthorization` Trait
+
+**Likelihood Of Impact: Low**
+
+The `Illuminate\Auth\Access\HandlesAuthorization` trait's method `deny()` has been update to return `\Illuminate\Auth\Access\Response`. Previously it used to throw an exception.
```
drbyte: 1. `has been update` should be `has been updated` or `has been changed`
2. `Previously it used to` might read better: `Previously it would`

upgrade.md
```diff
@@ -79,6 +79,22 @@ The constructor signature of the `Illuminate\Auth\Access\Response` class has cha
 
 The `Illuminate\Contracts\Auth\Access\Gate` contract has received a new `inspect` method. If you are implementing this interface manually, you should add this method to your implementation.
 
+#### The `Illuminate\Auth\Access\HandlesAuthorization` Trait
+
+**Likelihood Of Impact: Low**
+
+The `Illuminate\Auth\Access\HandlesAuthorization` trait's method `deny()` has been update to return `\Illuminate\Auth\Access\Response`. Previously it used to throw an exception. 
+If you have been using this method in your policy classes to have a `return` statement like:
```
drbyte: This sentence seems to be missing a word or two. Maybe the word "add"?
Perhaps instead of starting with "If" maybe it should just be "You now need to ..." and then be clear about what's needed.

upgrade.md
ankurk91: @drbyte 
Thanks, pushed new changes

upgrade.md
ankurk91: please check latest commit

PR #5532, opened 2019-10-14
[6.x] Update minio control panel url

skylerkatz:
On a fresh install of Homestead using Laravel 6 the Minio control panel is available at localhost, not homestead.


driesvints:
Thanks!

PR #5535, opened 2019-10-17
[6.x] Directly clone release branch if we only want to use it

abenettt:
This saves us the extra step of checking out the branch later


driesvints:
Since the docs literally say this is the "Alternatively" approach I don't think this should be the default?

PR #5537, opened 2019-10-17
Add missing Engine::map argument to scout docs

matt-allan:
The `Engine::map` method [takes a `Builder $builder` parameter](https://github.com/laravel/scout/blob/31ddbaa62791d0148bb79a0456e1c9a287c50e2c/src/Engines/Engine.php#L59) as the first argument.


driesvints:
Thanks!

PR #5538, opened 2019-10-17
Consistency: API token length 80 characters

repat:
```php
Str::random(80)
```

 because of

```php
$table->string('api_token', 80)->after('password')
```


driesvints:
Thanks!

PR #5539, opened 2019-10-18
[6.x] Demonstrate simplest method to attach relation

dwightwatson:
I notice a lot of examples where developers are using complex setups in order to build associations in their model factories - either using `factory(Model::class)->create()->id`, or doing it in a callback to make it happen lazily.

I think the docs should demonstrate that the easiest solution is simply to associate `factory(Model::class)` and let Laravel do the rest of the work, while still showing how closures can be used for more complex configurations.

### DIFF
--- a/database-testing.md
+++ b/database-testing.md
@@ -227,7 +227,7 @@ You may also attach relationships to models in your factory definitions. For exa
         ];
     });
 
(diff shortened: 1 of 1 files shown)


decadence:
> depends on the facotry

PR #5540, opened 2019-10-18
Topic Title

mayursojitra:
- In the API Authentication documentation, Table of content titled **Database Migrations** and the main topic was **Database Preparation**, So I correct it with **Database Migrations**.


driesvints:
Unfortunately we can't change the anchor links because that'd break links to this section.

PR #5542, opened 2019-10-18
Update scheduling.md

Naoray:
Adds a reference to the `schedule_timezone` config which eliminates the need to add a `scheduleTimezone()` method in the `app/Console/Kernel.php` when all tasks should be scheduled in the same timezone.


taylorotwell:
I don't think the existing documentation should be removed.

Naoray:
And what about adding a reference to the config value? Would that be ok?

PR #5545, opened 2019-10-19
Added tip to add api_token to fillable attributes

CaddyDz:
Users may spend some time to figure out why they're having null API tokens due to the Mass Assignment on the `create` method of the `RegisterController` so I think it's a good idea to mention this


taylorotwell:
Honestly that's probably not a good attribute to make fillable.

PR #5548, opened 2019-10-21
[6.x] Add section for action() helper in upgrade guide

mrk-j:
I noticed the `action()` helper is also affected if there is no matching key within the route path.

This PR makes that more clear in the upgrade guide.


taylorotwell:
Your example is incorrect. It won't add an "id" variable to the query string.

PR #5553, opened 2019-10-22
Add a note about the Route Name Prefixes key

Sicklou:
When using the  `Route::group` with an array as  the first parameter, the key for name prefixes is not `name` but `as` and it is not mentioned in the documentation.

I let you change the formulation because English is not my native language.

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -298,6 +298,8 @@ The `name` method may be used to prefix each route name in the group with a give
         })->name('users');
     });
 
+When using the array as a first parameter to the `Route::group` method, the key must be `as` and not `name`.
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
This feels really out of place here.

Sicklou:
> 
> This feels really out of place here.

At the beggining of the route section it says : 

> Shared attributes are specified in an array format as the first parameter to the Route::group method

Then I thought the `name` key in the array will work.
Just like `namespace` and `prefix`.
Lost 30 minutes.
There is no note about that.

It was mentioned in the 5.2 then disappear since the 5.3 documentation.
Maybe it is a bad way to use it ?

PR #5554, opened 2019-10-22
Add route using callable array syntax

brunodevel:
It is possible to define routes to controller actions using the callable array syntax, but it was not documented in the routing section.


taylorotwell:
That won't register correctly when using the default 'namespace' directive in your RoutServiceProvider.

brunodevel:
@taylorotwell I think the condition bellow returns false when using this syntax, so the method `convertToControllerAction` is not called and the namespace not appended in this case.

https://github.com/laravel/framework/blob/6.x/src/Illuminate/Routing/Router.php#L461

PR #5558, opened 2019-10-25
Add notice to php artisan storage:link command

anotherssa:


taylorotwell:
I don't want to add to many of these little notes everywhere.

PR #5564, opened 2019-10-26
[6.x] Clarification on assertSessionHasErrors

dwyer14:
Adding some clarification on `assertSessionHasErrors` and `assertSessionHasErrorsIn`. Most notably, that you can pass either a simple, indexed array of fields to check for (generally), or an associate array of fields and specific error messages. Very open to suggestions on wording, but thought this would be a worthwhile, helpful addition.

### DIFF
--- a/http-tests.md
+++ b/http-tests.md
@@ -556,16 +556,16 @@ Assert that the session has a given list of values:
 <a name="assert-session-has-errors"></a>
 #### assertSessionHasErrors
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
The source code is actually `$keys` though, so I think we would want to keep that.

dwyer14:
Ahh, good point. 👍 I might resubmit this after reverting `$fields` to `$keys`, if you think the other edits/additions are helpful.

PR #5565, opened 2019-10-26
Document Tap

deleugpn:
Sorry I didn't follow up on this earlier. I changed the documentation I had made for `apply` to match the existing `tap` behavior.


taylorotwell:
I appreciate the PR but this particular example doesn't really demonstrate when or how this would be useful.

deleugpn:
I'll work on another sample taken from my real life context and resubmit for one last evaluation

PR #5569, opened 2019-10-28
[6.x] Document closure as second parameter

AndrewFeeney:
- Anticipate can now accept a closure as a second parameter instead of an array, as of 6.0


taylorotwell:
No idea why you removed "Closure" from table of contents. Fixed that.

AndrewFeeney:
😅Whoops, total accident. Thanks!

PR #5570, opened 2019-10-29
Master

obscurecodes:
In your documentation about <a href="https://laravel.com/docs/6.x/seeding#writing-seeders">seeding</a> you used Str::random() for name but it returns string with numbers.


driesvints:
Seems like something went wrong with branching here.

PR #5571, opened 2019-10-29
Document Mistake

obscurecodes:
In your documentation about <a href="https://laravel.com/docs/6.x/seeding#writing-seeders">seeding</a> you used Str::random() for name but it returns string with numbers.


driesvints:
Heya. Seems like something is still going wrong with branching. If you need some tips on getting started with git, Atlassian has a great guide: https://www.atlassian.com/git

PR #5584, opened 2019-11-04
[6.x] Installation - Add missing PHP extensions

polarathene:
Adds 3 PHP extensions to the installation docs which may be required to get a running example project working.

---

I have tried following the [Laravel Installation docs](https://laravel.com/docs/master/installation) with Alpine Linux.

`composer create-project --prefer-dist laravel/laravel blog` fails to install due to missing extensions for PHPUnit, `DOM` and `XMLWriter`. These two are required while the others listed in [PHPUnit requirements](https://phpunit.readthedocs.io/en/8.4/installation.html#requirements) should already be available. 

`XDebug` appears to be optional, or at least the default Laravel install via Composer succeeds without it.

After a successful installation, `php artisan serve` fails to load the `public/index.php` due to an error about sessions. Installing the `Sessions` PHP extension resolves this.

### DIFF
--- a/installation.md
+++ b/installation.md
@@ -28,6 +28,8 @@ However, if you are not using Homestead, you will need to make sure your server 
 - PDO PHP Extension
 - Tokenizer PHP Extension
 - XML PHP Extension
+- DOM PHP Extension
+- XMLWriter PHP Extension
 </div>
 
 <a name="installing-laravel"></a>
--- a/installation.md
+++ b/installation.md
@@ -28,6 +28,7 @@ However, if you are not using Homestead, you will need to make sure your server 
 - PDO PHP Extension
 - Tokenizer PHP Extension
 - XML PHP Extension
+- Sessions PHP Extension
 - DOM PHP Extension
 - XMLWriter PHP Extension
 </div>


taylorotwell:
Sessions is enabled by default in PHP and the other two are not required to run Laravel itself.

polarathene:
@taylorotwell That's not true? I can provide a reproducible Docker image with alpine where the lack of the `php7-session` package will cause a failure.

While the other two extensions are not required to run Laravel itself, the installation docs will fail with the composer install twice as only one extension error is presented each time for PHPUnit.

What is a new user following those installation docs meant to experience? There is no harm in mentioning those extensions(they can be separate and point out that they're for PHPUnit in order for the project example to successfully install), there is nothing on the docs that indicate a failure due to these missing is to be expected.

With the lack of the sessions extension and a user new to PHP, it is not obvious that it's due to a missing extension. The errors stem from two files about session handling. The Laravel installation docs make no mention of such an extension possibly being needed(because you seem to assume it's there by default?). It might be available by default on certain platforms/distros, but this is not the case, otherwise I wouldn't have created the PR.

polarathene:
`php7 php7-fpm php7-json php7-openssl php7-phar` base Alpine packages for PHP-FPM and Composer(JSON, OpenSSL, Phar extensions).

`php7-xml php7-mbstring php7-ctype php7-pdo php7-tokenizer php7-bcmath` are the remaining extensions from the laravel docs. But it will fail.

`php7-dom php7-xmlwriter` will get you a successful install. The project dependency of PHPUnit requires them. `php7-session` will provide Sessions extension, otherwise you'll get an error from serving the page.

polarathene:
@taylorotwell hello?

Please try `docker run -it --rm alpine:3.10 sh -c 'apk add --no-cache php7 && php -m'`, you will see the following output:

```
fetch http://dl-cdn.alpinelinux.org/alpine/v3.10/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.10/community/x86_64/APKINDEX.tar.gz
(1/9) Installing php7-common (7.3.11-r0)
(2/9) Installing argon2-libs (20171227-r2)
(3/9) Installing ncurses-terminfo-base (6.1_p20190518-r0)
(4/9) Installing ncurses-terminfo (6.1_p20190518-r0)
(5/9) Installing ncurses-libs (6.1_p20190518-r0)
(6/9) Installing libedit (20190324.3.1-r0)
(7/9) Installing pcre2 (10.33-r0)
(8/9) Installing libxml2 (2.9.9-r2)
(9/9) Installing php7 (7.3.11-r0)
Executing busybox-1.30.1-r2.trigger
OK: 20 MiB in 23 packages
[PHP Modules]
Core
date
filter
hash
libxml
pcre
readline
Reflection
SPL
standard
zlib

[Zend Modules]

```

No `session` module listed. PHP does not provide this by default, if it does it is because the package has bundled it, or if using a Docker image like the official PHP ones which compile source code, they have chosen to include it. That does not equate to the extension being available by default.

Arkanius:
Sounds wierd to me that Laravel needs php sessions to run. I have a docker image based on alpine that don't need all this extensions. Please check the **vtrgomes/composer-alpine:7.3** image.

Here is the [dockerfile](https://github.com/Arkanius/docker-images/blob/master/php73/Dockerfile)

ahinkle:
Laravel has ran on Docker for years without needing the extensions. Please check your PHP image.

polarathene:
In seeking clarification for how these 3 extensions are installed by default elsewhere, I came across [these docs](https://www.php.net/manual/en/extensions.alphabetical.php) and each of the mentioned extensions install pages note that they're built/shipped by default with PHP, but can optionally be disabled/excluded.

The PR could be adjusted to separately mention them for troubleshooting, but I guess it's rare that Alpine Linux would be used by the guides reader. Often when it is, the reader is using it via an existing Alpine image for Docker, commonly based of the official DockerHub one which as mentioned differs in that it compiles from source adding additional extensions.

---

@ahinkle You have a command provided with the above output, you're making an assumption based off the DockerHub official image, where it's common for these types of images to build from source, rather than use the distros native packages. Feel free to try the command yourself.

@Arkanius same to you, you're not using Alpine's php packages. I am as their provided version is fine allowing my Dockerfile to make an image a third of the size(for a basic Laravel base) of what you use as a base, or ~35MB when matching PHP extension parity to `php:7.3-alpine`(~77MB).

PR #5586, opened 2019-11-04
Update authorization.md to correct phpdoc output

adamhopkinson:
The code sample under `Policy Responses` mistakenly had a phpdoc block stating that it returns a `bool`, when the code shows it returning a `Response`

### DIFF
--- a/authorization.md
+++ b/authorization.md
@@ -299,7 +299,7 @@ So far, we have only examined policy methods that return simple boolean values. 
      *
      * @param  \App\User  $user
      * @param  \App\Post  $post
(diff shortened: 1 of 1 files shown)


authorization.md
```diff
@@ -299,7 +299,7 @@ So far, we have only examined policy methods that return simple boolean values.
      *
      * @param  \App\User  $user
      * @param  \App\Post  $post
-     * @return bool
+     * @return Response   $response
```
fitztrev:
```suggestion
     * @return \Illuminate\Auth\Access\Response
```

PR #5590, opened 2019-11-05
Update Billing.md

rossity:
Replaced instances of 'main' subscriptions with 'default'

Inside the Billable.php trait, the default name for subscriptions are 'default' in onTrial(), subscribed(), subscription(), etc. so it might make sense to be consistent with this in the docs.


driesvints:
Then again this doesn't makes it obvious that you could it omit it for certain methods as it's often an optional parameter. But I'm good with this change either way.

PR #5592, opened 2019-11-07
[6.x] Add nested resources to route resources

zoomyboy:
I was wondering why the documentation says nothing about nested resources, since I really like this feature and use this a lot.

This can be found in the laravel 5.1 docs, however, since laravel 5.2 it's not documented anymore even though this is still possible. So I merged this over ;)


taylorotwell:
I'm honestly not a big fan of using these so probably wanted to discourage people from doing it.

PR #5599, opened 2019-11-11
[6.x] Improved warning on `CreateFreshApiToken` middleware effects

satyago:
This pull request is an attempt to remove future confusions about the execution order requirements of the `CreateFreshApiToken` middleware as it previously lacked detail on the reasoning and sufficient warning on the consequences.

This is a result of an issue I entered on the `laravel/passport` package before: https://github.com/laravel/passport/issues/1114

**UPDATE:** This may also need to be merged back into the 5.8 documentation or versions prior.


taylorotwell:
Don't like this wording. I'll fix it.

PR #5611, opened 2019-11-16
[6.x] Upgrade chrome driver for travis ci

nikolaynizruhin:
This PR adds chrome driver update command during travis ci installation.

Without this command travis ci fails:
Снимок экрана 2019-11-16 в 12 09 22

### DIFF
--- a/dusk.md
+++ b/dusk.md
@@ -1456,6 +1456,7 @@ To run your Dusk tests on [Travis CI](https://travis-ci.org), use the following 
       - cp .env.testing .env
       - travis_retry composer install --no-interaction --prefer-dist --no-suggest
       - php artisan key:generate
+      - php artisan dusk:chrome-driver
 
     before_script:
       - google-chrome-stable --headless --disable-gpu --remote-debugging-port=9222 http://localhost &


ahinkle:
Laravel Dusk ships with a Chrome Driver that can be consumed by Travis CI: `laravel/dusk/bin/chromedriver-linux`

driesvints:
I don't think this is necessary. This should be an indication that your local installation differs from your CI one. It also wouldn't make sense that Travis was the only one failing but the other ones are okay?

nikolaynizruhin:
We have this update driver command in github actions example. I just tried to run github action without driver update command and I got the same error as I posted above. I think it might be the case for all other CIs

driesvints:
I just tried out the Travis example and the chromedriver install step indeed gets me one step further but I can't seem to get tests passing with a vanilla Laravel app with Dusk installed. Getting this failed build: https://travis-ci.org/driesvints/test-dusk/builds/615075856

I'm wondering is anything has changed in the meantime. Did you get the tests to pass after updating the chrome driver in the install steps?

nikolaynizruhin:
Screenshot 2019-11-21 at 15 35 24
Can you try this one? It was an issue for me. It is in the very end of dusk docs page

driesvints:
@nikolaynizruhin that indeed worked. Thanks!

It indeed wasn't clear that you needed to do that for every CI. I'll send in a follow-up pr.

nikolaynizruhin:
Great!

driesvints:
Sent in a follow up here: https://github.com/laravel/docs/pull/5621

PR #5612, opened 2019-11-19
[5.8] Removing global scopes on relationship

akmet:
In the current 5.x branch are no information about the possibility to remove a global scope on relationships.

Since I am not a native speaker I am very pleased about any enhancements on the phrasing.

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -724,6 +724,28 @@ If you would like to remove several or even all of the global scopes, you may us
         FirstScope::class, SecondScope::class
     ])->get();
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
What is a common use case for doing this?

akmet:
I'm not sure if it is common but we use a user model for 2 different types of users. Active members and former members. Former members have no personal data like phone number. In most cases we only want to query the active members. So we created a globalscope on boot. But for some edge cases we want the relationship to use both active and former members.

PR #5613, opened 2019-11-20
[6.x] fix description of successful status in assertSuccessfull()

smknstd:


smknstd:
shoud fix in others versions as well ?

PR #5616, opened 2019-11-21
[6.x] Update collections.md

a2sc:
Add note to use dot notation and wilcard to pluck values in a collection with nested keys


taylorotwell:
I find this a bit complicated. Even for me I'm a bit confused by why the names are returned in arrays. I think it's clearer to just use `map` in these situations 😅

PR #5618, opened 2019-11-21
Update cache docs with minutes not seconds

owalerys:
All of the methods in `Illuminate\Cache\Repository` use minutes, not seconds, as the docs currently state.


driesvints:
That's not true. All cache methods use seconds, not minutes. Where do you see the docs mentioning minutes?

owalerys:
@driesvints The docs say seconds, the source code says minutes. Take a look at `Illuminate\Cache\Repository`. My code was silently-failing because I was passing in seconds but `::put()` was expecting minutes and so the value was out of range for the `database` cache driver.

owalerys:
@driesvints 
image

driesvints:
@owalerys https://laravel.com/docs/5.8/upgrade#cache-ttl-in-seconds

PR #5619, opened 2019-11-21
[6.x] Added path() to the file docs

tomswinkels:
Added path() to the docs


taylorotwell:
I'm not sure this is super useful and is a little confusing when using cloud storage like S3.

PR #5622, opened 2019-11-21
[6.x] fix closing @componentFirst tag

browner12:
per this commit

https://github.com/laravel/framework/commit/7ff698ddab815888759946a5a2b71b019de39911


taylorotwell:
Since it just proxies into endComponent, can't you just use endComponent?

browner12:
you can.

if you'd prefer to keep it documented this way we can. the reason I switched the docs (and made the PR originally) is because the Blade IDE plugins don't know how to match up these directives, and they show errors, which is annoying, and can cause issues with IDE auto-formatters.

Ideally, the plugin would be able to recognize this, and we wouldn't need the directive at all.

PR #5623, opened 2019-11-22
[6.x] Documents the explicit Model definitions in database rules

Arkanius:
Documents how to use the Model class at the validation rules: https://github.com/laravel/framework/pull/30653

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -805,6 +805,12 @@ If you would like to customize the query executed by the validation rule, you ma
         ],
     ]);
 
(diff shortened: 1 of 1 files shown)


ahinkle:
Might be a bit cleaner and more uniform (with the rest of the docs) to use `App\User` but I'll let Taylor decide. 👍

`'user_id' => 'exists:App\User,id'`

Arkanius:
You might be right, I can do this.

PR #5631, opened 2019-11-26
Add missing "static" declaration in custom resolution logic example

nathanpurcell:
Personally, I quite often copy the example you provide in the docs - this one was missing the `static` declaration.


driesvints:
Service providers don't use a static keyword for the boot method.

PR #5633, opened 2019-11-26
Document accessing nested localisation files.

jeff-h:
Not sure if this is desirable behaviour, but the ability to nest localisation files is helpful in large sites. However, I don't believe it's currently documented.


ahinkle:
#3957

jeff-h:
Ah, thanks. I suspected it might be unintentional, given the inconsistency of `/` vs `.`. Shame though.

PR #5634, opened 2019-11-27
[6.x] Add anchor tag to introduction.

eric-famiglietti:
There's no number sign next to the **Introduction** heading (and the whitespace is incorrect):

Screen Shot 2019-11-27 at 1 58 06 PM

I'm not sure if this fixes the issue, but the anchor tag for the introduction heading is present in every other file.


driesvints:
Thanks!

PR #5636, opened 2019-12-01
[6.x] Add use of selectSub

kevinb1989:
I want to add an use of selectSub, which is an alternative for selecting a sub query. This is in accordance with talks and [an article](https://reinink.ca/articles/dynamic-relationships-in-laravel-using-subqueries) by @reinink.

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -328,6 +328,20 @@ Using the subquery functionality available to the `select` and `addSelect` metho
         ->limit(1)
     ])->get();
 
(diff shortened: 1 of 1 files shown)


reinink:
Yes, you can definitely use `selectSub()` for this as well, but I think it's better to use `select()` and `addSelect()`, as per the [subquery enhancements in Laravel 6.0](https://laravel-news.com/eloquent-subquery-enhancements). I think having one way to add selects, be it a regular column or a subquery, is better than having to remember another method. And `selectSub()` really doesn't add anything here, other than a slightly different syntax.

So, I guess I'd suggest not including this in the documentation, simply because I don't think it's necessary anymore.

PR #5643, opened 2019-12-02
Additional information for in: validation rule

JeremyM1985:


taylorotwell:
I guess I'm not sure how else people might assume they should implode it?

PR #5645, opened 2019-12-03
[5.8] Specifies from which version insertOrIgnore is available

Hakim-Bou:
Because I was on v5.8.11 and I couldn't find this method even I was on the 5.8 docs.


taylorotwell:
I personally don't want to start adding these types of notes all over the docs. You should be on the latest version of 5.8.

PR #5664, opened 2019-12-17
[6.x] Add prepareForValidation method to docs

intellow:


browner12:
thanks for adding this! just recently discovered it, and had it on my list to document.

PR #5666, opened 2019-12-17
[6.x] Add note about forceCreate

andrewminion-luminfire:


taylorotwell:
Not sure this is necessary.

andrewminion-luminfire:
No, not strictly necessary, but it wasn’t immediately obvious to me that I could use `forceCreate` rather than adding `api_token` to the `User::fillable` array, and in fact I didn’t even notice right away that it was `forceCreate` instead of the default `create`. Figured it might help someone else down the road ¯\\\_(ツ)_/¯

PR #5668, opened 2019-12-18
Setting the recipient name in a notification

highstrike:
Ref latest answer in https://github.com/laravel/framework/issues/18439#issuecomment-567038986


taylorotwell:
This should not be a tip. It should just be documented normally.

highstrike:
Do you want me to change it into a normal paragraph rather than a tip?

PR #5670, opened 2019-12-18
Add a note on stopwaitsecs

themsaid:


themsaid:
@halaei That's for Horizon only. I think we can mention the fast_termination option in the docs too.

halaei:
Yes. I added my comment in a wrong file. The same note is in horizon.md too.

PR #5684, opened 2019-12-29
Added additional info on running NPM.

ravigehlot:
Vagrant can be stubborn depending on the operating system and various settings. On macOS, running NPM from within Homestead may not work as expected. That is not necessarily an issue with Laravel Mix or NPM. I found that running NPM from the host machine works. This commit adds info on that.


taylorotwell:
Hmm, I'm not sure we need a note for this.

PR #5688, opened 2019-12-30
[6.x] Correct front-end and front end spelling

zepfietje:
Note that the file `frontend.md` has been renamed to `front-end.md`, which might result in unexpected behavior when switching between versions on the docs page. However, this change is needed to fully reflect the correct spelling.


taylorotwell:
I'm OK leaving this as is.

PR #5690, opened 2019-12-31
Edit a Typo in Redis Default Client Section

K1-Aria:
I had found a typo in the (redis default client) section and corrected it.


zepfietje:
This is not a typo; the current text is correct.

K1-Aria:
the document said : default redis client has changed from "predis" to "phpredis". but in the end of paragraph is said : ensure the redis.client configuration option is set to "predis", In the event that it should be "phpredis"

PR #5692, opened 2019-12-31
Document 'WithFaker' trait

gdebrauwer:


taylorotwell:
I feel like this would be better documented with its own little section and a code example.

PR #5693, opened 2019-12-31
Document 'findAndValidateForPassport' method

gdebrauwer:
Was added to Passport by [#1144](https://github.com/laravel/passport/pull/1144)


taylorotwell:
Personally I think we would need more explanation of when you would actually use this... like a specific scenario.

PR #5694, opened 2019-12-31
Clarification about 'logoutOtherDevices' method

lsurma:
Hi,
I think part of documentation about 'logoutOtherDevices' method is missleading/wrong.
"Then, you may use the logoutOtherDevices method on the Auth facade. This method requires the user to provide their current password, which your application should accept through an input form:"

As far as i know, method 'logoutOtherDevices ' does not required to provide current password. Internally it will use any provided string without validation against user password. Developers in theirs apps should make validation, so users won't change their password unintentionally.

### DIFF
--- a/authentication.md
+++ b/authentication.md
@@ -417,7 +417,7 @@ Laravel also provides a mechanism for invalidating and "logging out" a user's se
         // ...
     ],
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
This should typically be used during a change password flow.

PR #5696, opened 2020-01-02
[6.x] Add note for Horizon does not work with Redis Cluster

bolechen:
Note: Horizon does not work with Redis Cluster.


taylorotwell:
This should just belong on the Horizon docs.

PR #5701, opened 2020-01-02
[7.x] Adds note about collision on upgrade guide

nunomaduro:
This pull request adds a note about collision on the upgrade guide.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -38,11 +38,9 @@ The new minimum PHP version is now 7.2.5.
 <a name="updating-dependencies"></a>
 ### Updating Dependencies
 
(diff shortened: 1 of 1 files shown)


browner12:
Is this necessary for the upgrade to v7?

Isn't this covered by our miscellaneous note at the bottom?

If we're going to specifically call out this, I say we specifically call out the `laravel/laravel` `composer.json` file instead.

nunomaduro:
> Is this necessary for the upgrade to v7?

Yes.

> If we're going to specifically call out this, I say we specifically call out the laravel/laravel composer.json file instead.

I think it's fine to keep it simple like this.

PR #5704, opened 2020-01-04
Added documentation for macros on TestResponses

ricardoboss:


taylorotwell:
Feels like this should be in the HTTP testing section.

ricardoboss:
I agree, this belongs more in the "testing" region. But macros aren't specific for HTTP tests (or `TestResponse`). Maybe I can move it to the end of "Testing: Getting Started" section or even create a new page for macros on `TestResponse` (HTTP) or `PendingCommand` (Console).

PR #5705, opened 2020-01-04
Component and Slot componentfirst directives tag changes.

Kmg-Kumar:
component first directives tag case sensitive changes in sample code.


taylorotwell:
The built-in Blade directives are not case sensitive.

PR #5707, opened 2020-01-05
[6.x] Update passport.md to add Authentication code w/ PKCE grant

filipegar:
Adds docs for Authentication code with PKCE Grant - laravel/passport#1065


driesvints:
@filipegar thanks for this!

PR #5709, opened 2020-01-07
[6.x] Clarified that failed() does not get called with disptchNow()

Patabugen:
It is not clear from the docs that the failed method only applies when dispatched via a queue, and will not be called if an error occurs when the job has been dispatched with dispatchNow(). This clarifies it, based on a discussion on iddue #31004

https://github.com/laravel/framework/issues/31004


taylorotwell:
The formatting here does not match the rest of the style of the docs. I will add a note about this.

PR #5710, opened 2020-01-07
[6.x] Update GitHub Actions example config for Dusk

nikolaynizruhin:
This PR updates GitHub Actions example by replacing `.env.example` with `.env.testing` since at the beginning of the Continuous Integration section, it says: we need to ensure that `.env.testing` file contains an `APP_URL` with a value of `http://127.0.0.1:8000` and also all other ci examples are using `.env.testing`.


browner12:
I don't think everyone uses an `.env.testing` file in their application so I think it would be better to leave as is.

What we *could* do is show an example of how to set an environment variable within your Action file.

PR #5717, opened 2020-01-12
[6.x] Update database-testing.md

DJR-Hermans:
Update static password in user factory example


browner12:
why?

driesvints:
Guessing because the current example is out of date with the file from the skeleton.

PR #5720, opened 2020-01-14
[6.x] Explain the job isolation difference between queue:work and queue:listen

Frederick888:
I was suggested to explain the difference between `queue:work` and `queue:listen` a bit more in laravel/framework#31120.

When `queue:work` is used, since workers are 'long-lived processes', it actually lowers the degree of job isolation and some changes made by a job can affect subsequent ones, e.g. non-idempotent mutations to static variables. This is not by design (I think) but a limitation of PHP.

In such a scenario, users can use `queue:listen` instead at the expense of performance.

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -585,9 +585,9 @@ Laravel includes a queue worker that will process new jobs as they are pushed on
 
 > {tip} To keep the `queue:work` process running permanently in the background, you should use a process monitor such as [Supervisor](#supervisor-configuration) to ensure that the queue worker does not stop running.
 
-Remember, queue workers are long-lived processes and store the booted application state in memory. As a result, they will not notice changes in your code base after they have been started. So, during your deployment process, be sure to [restart your queue workers](#queue-workers-and-deployment).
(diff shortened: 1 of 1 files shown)


queues.md
```diff
@@ -591,6 +591,8 @@ Alternatively, you may run the `queue:listen` command. When using the `queue:lis
 
     php artisan queue:listen
 
+Another benefit of `queue:listen` is that it features a higher degree of job isolation, which prevents changes made by a job from being propagated to subsequent ones in certain scenarios, e.g. static variables mutations.
+
```
mfn: Personally I don't like this "pro argument" for `queue:listen`.

IMHO `queue:work` is absolutely to be preferred (e.g. if you "go horizon", it's not possible to do otherwise AFAIK).

Rather I think the possible pitfall for `queue:work` should be pointed out clearer.

:-)

Frederick888: @mfn Just pushed an update. How about it now?

mfn: > non-idempotent mutations

**I** ❤️ it, thanks!

PR #5729, opened 2020-01-17
[7.x] Updates wording of `Likelihood Of Impact`

nunomaduro:
Hey friends,

What do you think about changing the wording of **Likelihood Of Impact** to **Risk of impacting you**?


taylorotwell:
I think I'm fine with it as is.

PR #5730, opened 2020-01-17
[6.x] Add note about double encoding

browner12:
add some docs to explain how to locally disable double encoding when outputting data in Blade

### DIFF
--- a/blade.md
+++ b/blade.md
@@ -218,7 +218,7 @@ The `@json` directive is also useful for seeding Vue components or `data-*` attr
 
 #### HTML Entity Encoding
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hmm, wow I didn't know this worked. I'm not sure this was 100% intentional on my part so I'm going to hold off on documenting this for now.

browner12:
Yah, it does look a little weird.  But it's just like a component receiving all the data passed to it, just in a little different syntax.

Maybe another "echo" style?

```blade
{## "Laravel &amp; PHP" ##}
```

PR #5736, opened 2020-01-19
[6.x] Named rate limiting

mnaas:
This is an update to the docs related to using named rate limiting, which is so important to be clear for all Laravel users.

This is my first time making a pull request to a public repository, so I'm sorry if it's not a perfect one.

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -413,6 +413,26 @@ Laravel includes a [middleware](/docs/{{version}}/middleware) to rate limit acce
         });
     });
 
+#### Named Rate Limiting
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm confused. Your route does not have any group throttle applied to it anyways so what is it overriding?

taylorotwell:
I guess even after this explanation even I'm still unclear what this does.

mnaas:
Looks like I am missing something here. First I made a mistake appling the `auth:api` middleware for the sub-group.

Correct me please if I am wrong. If we have 2 routes or even a route and a group of routes as in the example, and we applied throttling for both without passing names to them, isn't request counter one for both?
Lets say we have throttle:10,1 for a route and throttle:3,1 for a 2nd route. If we get 3 requests hitting the first route and then one request for the 2nd route. Isn't that 4th request going to be blocked even if it is the first hit for the 2nd route?

taylorotwell:
Yes, maybe you are correct there, for authenticated requests.

PR #5742, opened 2020-01-23
Added composer global installation path for Linux

SamAsEnd:
> If your system uses freedesktop.org standards, then it will first check
XDG_CONFIG_HOME or default to `/home/\<user\>/.config/composer`

From [composer source code](https://github.com/composer/composer/blob/master/src/Composer/Command/GlobalCommand.php#L47-L48)

The doc is misleading developers on Linux leading to questions like [this](https://stackoverflow.com/a/40470979) and [this](https://laracasts.com/discuss/channels/laravel/i-add-homecomposervendorbin-to-my-ubuntu-path-but-it-then-removed-automatically)


SamAsEnd:
Thanks, Taylor :+1:

PR #5746, opened 2020-01-24
Add example table structure for Many To Many

danabrey:
There is sometimes more confusion than necessary when developers new to Laravel first create a Many To Many relationship.

The docs have example table structure for polymorphic relationships, I think it might help to include an example for Many To Many, too, as it's the first move away from a simple one-model-one-table structure in this page of the docs.


danabrey:
On second thoughts, using a different example than `users` might be sensible, as that table will already exist by default, and might cause extra confusion.

PR #5755, opened 2020-01-30
[6.x] Document how to use a custom Passport Client model

jryd:
Added in Passport 7.0 is the ability to extend the Passport Client model to add a skipsAuthorization method. However, no documentation was included on how to tell Passport to use the extended model.

### DIFF
--- a/passport.md
+++ b/passport.md
@@ -384,7 +384,7 @@ If you would like to customize the authorization approval screen, you may publis
 
     php artisan vendor:publish --tag=passport-views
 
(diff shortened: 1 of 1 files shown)


driesvints:
There's already docs for this here: https://laravel.com/docs/6.x/passport#overriding-default-models

driesvints:
I realized that you're actually want to make it obvious that the model needs to be overwritten first. Maybe we can link to the docs above instead?

PR #5764, opened 2020-02-05
Add Cashier Mollie

sandervanhooft:
Adding a separate document for Cashier Mollie.


taylorotwell:
It's empty :)

sandervanhooft:
Fixed here: #5765

PR #5767, opened 2020-02-06
Highlight differences in Feature/Unit ExampleTest

edwardsteven:
After bashing away for hours trying to understand why my Unit tests weren't connecting to the DB I read PR #5169 and realized the tests/Unit/ExampleTest.php has been updated; however, both extend 'TestCase' and it wasn't clear (esp. with PHPStorm imports folding) that the TestCase being extended was different - given the ambiguity (outside of reading the use statements) I wanted to highlight for other users the fact that by default Unit tests don't load the framework TestCase and therefore won't boot Laravel - hoping to save somebody else a bit of forehead smacking.


edwardsteven:
Closing, opened PR on laravel/laravel to alias the PHPUnit/Framework/TestCase to avoid needing to modify documentation

PR #5776, opened 2020-02-13
[6.x] Warn about unintended query duplication when using * in view composers

bertheyman:
Devs new to view composers, sometimes assume `*` only triggers composers for views rendered via the view() function in the controller.
Since @include also triggers `*` view composers, data might be re-rendered much more than the dev meant to - often only a problem when the projects exists for a while and grows.

This PR adds some context to avoid this error.


taylorotwell:
No plans to add this comment right now.

PR #5779, opened 2020-02-14
Add "use" statement for Controller class

tomhatzer:
Hello!

I know this is not a big thing, but while browsing the socialite docs I saw that there's a use statement missing for the Controller.
The code would throw an error as it can't find the Controller class in the App\Http\Controllers\Auth namespace.

Thanks!
Tom


driesvints:
Thanks

PR #5786, opened 2020-02-16
[6.x] Surfacing the overridable authenticated and loggedOut methods

adaviladev:
Adding visibility of undocumented methods for extending the login/logout response without having to alter the core Laravel behavior.

### DIFF
--- a/authentication.md
+++ b/authentication.md
@@ -89,6 +89,24 @@ When a user is successfully authenticated, they will be redirected to the `/home
 
     public const HOME = '/home';
 
+#### Response Customization
+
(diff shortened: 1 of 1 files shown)


adaviladev:
I was working on an API and wanted to leverage the existing authentication behavior and simply return a JSON response with tokens and other custom data. I couldn't find these methods in the documentation and initially was overwriting the `login` method. 

This seemed unnecessarily verbose, and after reading the source code, I discovered the `authenticated` and `loggedOut` methods that were made to short-circuit the original redirect response behavior.

taylorotwell:
The logout example seemed a bit out of place since it doesn't seem to be in context of the scaffolded controllers; however, merged the authenticated example.

adaviladev:
What do you mean by out of place? Is it that the changes should have been made in a different part of the file or that it didn't match the commentary?

PR #5789, opened 2020-02-18
[7.x] Add laravel/installer installation instruction

aesyondu:
Since `laravel new blog` is used as an example under header `park command`, it might be better to add the laravel/installer install instruction to complete the list of instructions.


taylorotwell:
This is discussed in another section of the documentation.

PR #5800, opened 2020-02-22
fix Defining Middleware section bug

Mdhesari:
Here in current Laravel docs example check age middleware will redirect us too many times and it is not usable at all.
In addition it becomes complicated for user in order to test on their local.

So I added an additional && condition and check if it is already redirected to home page and is on the home page so send the next middleware action.


taylorotwell:
It's just a simple example of how to define one it is not meant to be copy and paste.

PR #5801, opened 2020-02-24
[6.x] Add a tip regarding the potential false-positive by scout's batch import command

introwit:
If the value of the `queue` option in the `config/scout.php` configuration file is set to `true`, we aren't notified of any errors that arise during the batch import. The command output at the end gives a sense of false positive saying `[OK] All x-y-z records have been imported`. The errors won't even have to be in the users codebase, it can be algolia rejecting the model/object saying "the object is big, request for a larger quota" if a single random record is few Kb's bigger. Once such error happens, **no records after that are imported to algolia**. This can be a bit confusing error to debug. Thousands of records went missing for us, hence having a tip in the docs would help others like us 🙂


taylorotwell:
No plans to add this to the docs.

PR #5815, opened 2020-02-27
[6.x] Nginx configuration fixed

davidsarkany:
When user try open non exist file with .php extension, laravel don't handle with 404 error page, only show simple "File not found." message.
If you accept it, I fix the previous docs versions.

How to reproduce it?

- download this zip file: [docs-example.zip](https://github.com/laravel/docs/files/4263020/docs-example.zip)

- unzip it
- run `docker-composer up -d`
- Test the `http://old.localhost:8080/test.php` and `http://new.localhost:8080/test.php`

### DIFF
--- a/deployment.md
+++ b/deployment.md
@@ -45,6 +45,7 @@ If you are deploying your application to a server that is running Nginx, you may
         error_page 404 /index.php;
 
         location ~ \.php$ {
+            try_files $uri =404;
             fastcgi_pass unix:/var/run/php/php7.2-fpm.sock;
             fastcgi_index index.php;
             fastcgi_param SCRIPT_FILENAME $realpath_root$fastcgi_script_name;


taylorotwell:
We don't use this on Forge and everything works fine.

davidsarkany:
@taylorotwell 

So it's good work?
Laravel 404 `https://forge.laravel.com/non-exit`
No input file specified message `https://forge.laravel.com/non-exits.php`

browner12:
The default Nginx configuration for Forge (and Homestead) implements what I would call a "loose" front controller where we match any file that ends in `.php` and execute it through the appropriate PHP socket.

For most people, the only public `.php` file they have is `index.php` which obviously serves as their front controller.  However, because of the "loose" Nginx rules, people can execute any `.php` file in their public directory if it exists.

The other strategy would be to implement a more "strict" front controller pattern, which **forces everything** to be executed through the `public/index.php` file.

Rather than:

```
location ~ \.php$ {
}
```

we can be more strict with 

```
location = /index.php {
}
```

This forces **everything** through `index.php`. Then instead of having random PHP files in your public directory you can create routes and move the logic into a controller:

```php
Route::get('mycustomfile.php', 'MyController@custom');
```

You get a little extra overhead of the FW on these requests now, but it allows the organization of your code to be incredibly consistent.

This is the method I actually prefer, and use it on some projects that I am forced to use routes that end in `.php`.

In Homestead we actually have an (undocumented) site type for this exact use case called 'frontcontroller'.  https://github.com/laravel/homestead/blob/master/scripts/site-types/frontcontroller.sh

Then if someone tries to visit a route/php file that doesn't exist, it will get handled like a normal 404 route.

PR #5831, opened 2020-03-02
Blade - change namespace path to directory path

clarkeash:
Should this be the directory rather than the namespace?


clarkeash:
actually I've noticed this has been done in a few places, must be intentional.

PR #5839, opened 2020-03-02
Update verification.md

xyluz:
1. Show how the verification email can be sent from the registercontroller.php
2. Show how verified user can be checked in views

The 2 items above are not compulsory, but, quite useful for a beginner.


taylorotwell:
The controller part should not be needed as there is already an event listener registered to do this.

PR #5850, opened 2020-03-03
Fixed wrong info in frontend page

zupolgec:
Since Laravel 7 and Laravel UI 2, the Auth routes are defined in the latter. So the package should always be installed, not only in dev environments.

This change (https://github.com/laravel/framework/commit/aebd93d5f67de55473f3f0ff5056abbfc5972102) should be mentioned in the upgrade guide.


zupolgec:
In the upgrade guide (https://laravel.com/docs/7.x/upgrade) this part should be updated:

> UI v2.0 (No changes necessary)

PR #5852, opened 2020-03-04
Add tinker to composer dependencies

shpontex:
adding laravel/tinker to composer updating Dependencies


shpontex:
if tinker is not in 2.0 version it will create an error
 laravel/tinker v1.0.10 requires illuminate/support ~5.1|^6.0 -> satisfiable by laravel/framework[v6.14.0]

PR #5855, opened 2020-03-04
Changed artisan make name to reflect example

narwy:
`Users` to `User` to reflect the example class used below.


taylorotwell:
This is not incorrect.

PR #5878, opened 2020-03-11
Update airlock.md

emargareten:


taylorotwell:
Portions of this are not very details or correct.

PR #5884, opened 2020-03-13
Update queries.md

rotorist:
Added an example to show how to pass a variable into the closure to be used in the where clause.


taylorotwell:
That seems more like PHP language documentation.

PR #5885, opened 2020-03-14
Update telescope.md

frizikk:
You should first make migrate then telescope:install otherwise you get error " Base table or view not found: 1146 Table 'database.telescope_entries' doesn't exist".

php artisan telescope:install command don't show any error but bugsnag does.


taylorotwell:
Current documentation works perfectly fine for me.

PR #5886, opened 2020-03-14
[7.x] Fix methods order and add 2 missing ones for `Arr` helper

denis-korchagin95:


taylorotwell:
Q comes before R and S?

denis-korchagin95:
> Q comes before R and S?

Yeah, I don't understand a little logic of placement too, because in original branch Q after R. So, I think what order of methods reflects the order in the source code and I fixed it to be the same.

taylorotwell:
That ordering doesn't make sense. Please just PR your additional only and put them in alphabetical order for now.

denis-korchagin95:
> That ordering doesn't make sense. Please just PR your additional only and put them in alphabetical order for now.

Ok.

PR #5892, opened 2020-03-18
Missing steps

AlexisSerneels:
Without this step, the Mailer [postmark] is not defined occurs.


taylorotwell:
This is already in the Laravel 7.x configuration file.

PR #5896, opened 2020-03-19
[7.x] Added expectsConfirmation docs to console tests page

bensherred:
Added documentation for the recent method `expectsConfirmation` which was added in Laravel 7.2 and can be used to test artisan commands that expect a confirmation such as yes or no.


taylorotwell:
This is already documented.

PR #5899, opened 2020-03-19
[7.x] Document Relation::getMorphedModel and Model::getMorphClass

zack6849:
There doesn't seem to be any documentation on this anywhere else, I sort of had to guess at it and look at the code, I figured adding this tidbit to the docs would help someone else later.


zack6849:
Not too sure if "alias" is the preferred term, feel free to change as necessary

driesvints:
@zack6849 it's the branch version

PR #5901, opened 2020-03-19
Update authentication.md

miagg:
Correct function name override for custom authentication


taylorotwell:
This documentation is referring to doing auth manually. has nothing to do with anything else.

miagg:
I tried to override the login functionality using the docs and I couldn’t get it work. I then saw that inside the trait I needed to override ‘login’ function and not ‘authenticate‘. Many people will do the same. That’s why I proposed to change it.

PR #5903, opened 2020-03-20
[7.x] Added Airlock docs SPA cors configuration

mehmetcansahin:
If you want to use SPA, it will help cors configuration.


taylorotwell:
That's not the only way to do it.

PR #5910, opened 2020-03-21
[7.x] Add transport

antonioribeiro:
I had to add this for my config to work. Not sure if it's a docs bug or framework.


taylorotwell:
You sure you didn't need this in your `mail` config? This is talking about the `services` config file, which has no concept of transports.

PR #5913, opened 2020-03-22
[7.x] Update upgrade.md

leemcd56:
[nunomaduro/collision](https://github.com/nunomaduro/collision) has changed the PHPUnit adapter class to `Printer`, which is a breaking change when upgrading from 6.x.


taylorotwell:
This can be in Collision's upgrade guide.

PR #5921, opened 2020-03-27
[7.x] Add explicit example of the use of extra parameters

martenkoetsier:
The docs were lacking explanation and example of the except and idColumn parameters of the Unique validation rule. The possibilities were explored using a Rule class, but not with the simple rule-string notation.


taylorotwell:
I would like people to always use the Rule class for this.

PR #5923, opened 2020-03-27
[7.x] Add information about broadcast notification type customization

golonix:
This PR adds missing piece saying how to customize notification type of a broadcast notification (used in Echo for example).


taylorotwell:
I think it would be better to just show an example.

PR #5928, opened 2020-03-29
[7.x] Add reference to MailSpons

badeend:
Disclaimer: I am the creator of MailSpons and I sincerely apologize if this is not the proper way to propose this change.

I am proposing to mention [MailSpons](https://mailspons.com/) in the Laravel docs as an alternative to Mailtrap. I've made a "MailSpons vs Mailtrap" page for quick comparison at: https://mailspons.com/vs-mailtrap.io

Kind regards,
Dave Bakker

### DIFF
--- a/mail.md
+++ b/mail.md
@@ -677,9 +677,9 @@ Another solution provided by Laravel is to set a universal recipient of all emai
         'name' => 'Example'
     ],
 
-#### Mailtrap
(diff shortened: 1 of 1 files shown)


taylorotwell:
I feel like this is sort of a never ending precedent to set where we have to mention every possible Mailtrap alternative. Sorry.

PR #5929, opened 2020-03-31
[7.x] Register anonymous components from a package

devfaysal:
Register anonymous components from a package https://github.com/laravel/framework/issues/32154


taylorotwell:
I think this would need more context. Possibly belongs under the package development docs?

PR #5933, opened 2020-04-01
Update validator extend method

KyawNaingTun:
I'm getting the error message
`call_user_func_array() expects parameter 1 to be a valid callback, class 'App\Rules\MyRule' does not have a method 'validate'. `

So, I have fixed to   `Validator::extend('foo', 'FooValidator@passes');`. It is working for me.


taylorotwell:
Because you don't have a validate method on your class.

taylorotwell:
Note that rule objects made via make:rule are not meant to be passed to Validator::extend.

PR #5934, opened 2020-04-01
Add castable docs

brendt:
Docs for https://github.com/laravel/framework/pull/32129

### DIFF
--- a/eloquent-mutators.md
+++ b/eloquent-mutators.md
@@ -7,6 +7,7 @@
 - [Date Mutators](#date-mutators)
 - [Attribute Casting](#attribute-casting)
     - [Custom Casts](#custom-casts)
+    - [Castables](#castables)
     - [Array & JSON Casting](#array-and-json-casting)
     - [Date Casting](#date-casting)
     - [Query Time Casting](#query-time-casting)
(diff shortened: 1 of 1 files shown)


eloquent-mutators.md
```diff
@@ -411,6 +412,43 @@ Once the cast is defined, you may access the `options` attribute and it will aut
...
+By implementing the `Castable` interface, you'll need to specify the caster class in the value object.
+
+    <?php
+    
+    namespace App\ValueObjects;
+    
+    use Illuminate\Contracts\Database\Eloquent\Castable;
+    use App\Casts\Address as AddressCast;
+    
+    class Addess implements Castable
```
mbardelmeijer, line 433: Small typo :)

eloquent-mutators.md
```diff
@@ -411,6 +412,43 @@ Once the cast is defined, you may access the `options` attribute and it will aut
...
+    {
+        /**
+         * Get the caster class for this class.
+         *
+         * @return string
+         */
+        public static function castUsing()
+        {
+            return AddressCast;
```
mbardelmeijer, line 442:
```suggestion
            return AddressCast::class;
```

eloquent-mutators.md
brendt: Solved

eloquent-mutators.md
brendt: Solved

taylorotwell:
In hindsight it probably would have been nice to pass `array $arguments` to the Castable object too, so that you from within `castUsing` you could use an anonymous class...

```php
public static function castUsing(array $arguments)
{
     return new class(...$arguments) implements CastsAttributes {
        get;
        set;
     }
}

```

PR #5935, opened 2020-04-01
[7.x] Fix grammar in Queues.md

rdarcy1:


taylorotwell:
middleware is plural here so left "allow" but kept the to change. Thanks.

PR #5948, opened 2020-04-06
[7.x] Update valet.md

crydotsnake:
I thought `Make sure you have export PATH=$PATH:~/.composer/vendor/bin in your .zshrc (OhMyZsh) or in your .bash_profile file.` would be better to understand then:

`Make sure the ~/.composer/vendor/bin directory is in your system's "PATH".`


taylorotwell:
I think the old version was simpler.

crydotsnake:
Not really.

PR #5957, opened 2020-04-09
[7.x] Mention that validation rule names are case-sensitive

Calinou:
This might be obvious to some, but I guess it doesn't hurt to mention it :slightly_smiling_face:

In the future, we could consider being more strict about validation names. This would likely be a breaking change though.


taylorotwell:
I don't think we ever document using them any other way.

PR #5958, opened 2020-04-09
Change the naming of the BeforeMiddleware class to be consistent with…

shutchinson:
… the example given above to make it clear that the classes simply need to have Before or After before their names.


taylorotwell:
Unnecessary IMO.

taylorotwell:
They don't need Before or After in their names.

PR #5960, opened 2020-04-11
Added 'directory' property to Supervisor config

pultho:
The runtime of the Supervisor and PHP can be set with the 'directory' property. No need to specify the Artisan path in the 'command' property.

http://supervisord.org/configuration.html?highlight=directory#program-x-section-example


taylorotwell:
I think it's OK as is.

PR #5962, opened 2020-04-12
Cashier: Minor fixes

Xenonym:
- Cashier: Remove locale from Stripe guide link

  Some users may be using Stripe in other languages, and we should not
  override the language they are currently using Stripe in.

- Fix typo: API's -> APIs


driesvints:
Thanks!

PR #5966, opened 2020-04-14
[7.x] Document component name property

driesvints:
Document https://github.com/laravel/framework/pull/32346


taylorotwell:
This feels like such a unique use case I may leave it out for now because it's hard to explain why you would ever need it except for your specific library 😆

driesvints:
Haha, okay fair enough.

PR #5975, opened 2020-04-15
[7.x] Normalize style to PSR-2

voyula:
Normalize style to PSR-2.

https://www.php-fig.org/psr/psr-2/

> Control structure keywords MUST have one space after them; method and function calls MUST NOT.

### DIFF
--- a/blade.md
+++ b/blade.md
@@ -269,12 +269,12 @@ You may check if a section has content using the `@hasSection` directive:
 
 Switch statements can be constructed using the `@switch`, `@case`, `@break`, `@default` and `@endswitch` directives:
 
-    @switch($i)
-        @case(1)
+    @switch ($i)
+        @case (1)
             First case...
             @break
 
-        @case(2)
+        @case (2)
             Second case...
             @break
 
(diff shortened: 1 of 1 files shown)


nunomaduro:
I prefer the current style in the docs.

voyula:
@nunomaduro But this style is not consistent.
We can force single code style in Laravel.
This prevents mental complexity.

voyula:
See: https://github.com/laravel/docs/blob/7.x/blade.md#if-statements

And see: https://github.com/laravel/docs/blob/7.x/blade.md#section-directives

This really not consistent.

Why two control structure keywords style different from each other.

taylorotwell:
Looks bad. I literally never put spaces like that.

taylorotwell:
Think about how you actually use PHP.

If I use the `count` function I don't do `count ($foo)`... I do `count($foo)`.

If I write an if statement I DO use a space `if ($foo)`

voyula:
@taylorotwell Yeah, I too, i DO use a space when I use a control structure(like if, elseif etc.).

Documentation using space in some places, like "unless, isset, empty, etc.".

This documentation why not using space on all control structures(elseenv('testing') etc.)?

Is it not control structure? 😕

PR #5976, opened 2020-04-15
[7.x] Normalize style

voyula:
See: https://github.com/laravel/docs/blob/master/blade.md#if-statements

Currently blade documentation uses space after "if, elseif and unless" statement.

Why this is using different style??? Why using different styles on same control structures.
No method calls here, have only control structures.

I'm so confused.


taylorotwell:
Stop. Please.

taylorotwell:
I DEFINE the style. Me. What I say is the style is the style. This is how I like it.

voyula:
Ok.

PR #5978, opened 2020-04-16
[7.x] Clearer HasOneThrough example

DarkGhostHunter:
The new example:

* Explains clearly the relation between the three models in association order.
* Changes the table order to mirror the verbal explanation.
* Uses the _User perspective_, meaning, an active participant (mechanic) is who access far data (car's owner).

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -431,20 +431,24 @@ If you have defined a many-to-many relationship that uses a custom pivot model, 
 ### Has One Through
 
 The "has-one-through" relationship links models through a single intermediate relation.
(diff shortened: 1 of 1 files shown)


taylorotwell:
Merged, but I want to note I had to fix several obvious problems with this. The spacing was all wrong on the table definition. And, the last example was also totally wrong and referenced a "Supplier" model that wasn't even in the example. I've noticed an uptick of PRs where people aren't paying attention to what they are doing so I want to note that please make sure your PR is well done before sending otherwise I have to pull it down and fix these things manually.

DarkGhostHunter:
Sure, I'll re-read more thoroughly next time.

PR #5984, opened 2020-04-17
Specified the option numbers.

anandhrushikesh:
Added 'Option 1' and 'Option 1' in front of the homestead installation options.
Will make it less confusing for people just getting started.


taylorotwell:
These aren't separate options?

PR #5986, opened 2020-04-17
[7.x] Update docs to reflect Redis prefix being present.

SketchNI:
When using Socket.IO to run your own websocket server with redis and Laravel, pushing to a channel results in a prefix being added because of a setting in `config/database.php:redis.options.prefix`


taylorotwell:
I thought the Laravel Socket.io server package removes this prefix for you?

SketchNI:
Are you referring to [this package](https://www.npmjs.com/package/laravel-echo-server)? It does not. Whether that is a recent change, I do not know.

taylorotwell:
image


https://github.com/tlaverdure/laravel-echo-server/blob/fbc1c9ebdb3de84d50d8f66d8f27caca9c2459c2/src/subscribers/redis-subscriber.ts#L48

PR #5995, opened 2020-04-20
[6.x] Explain how to disable requests throttling in tests

ziabice:
In this PR I explain how to use the `withoutMiddleware` method to disable the `Illuminate\Routing\Middleware\ThrottleRequests` middleware. This is useful when doing API testing.

Please note that I'm not a native english speaker so there could be some typos.


taylorotwell:
Feels like it would be better to just make throttling automatically disabled during tests.

PR #6003, opened 2020-04-22
[6.x] Document in Dusk scrolling an element into view

u01jmg3:
Documents https://github.com/laravel/dusk/pull/766

---

It's also worth noting that there is an example under 'Browser Macros' using `scrollToElement()`. This is misleading as there is no reason to macro this method as we have the undocumented `scrollTo()` as well as `scrollIntoView()`. Perhaps this example should be updated to something else. Equally it might also be worth documenting `scrollTo()`.


driesvints:
Heya, can you send this to 7.x instead? Thanks.

PR #6012, opened 2020-04-25
[7.x] Remove type-hinting in explicit binding

AliN11:
Type-hinting isn't needed anymore if model is bound explicitly in RouteServiceProvider


AliN11:
A little explanations would be appreciated.

PR #6013, opened 2020-04-26
[7.x] Added information about the `firstOr` method

andrey-helldar:
Source: https://laravel-news.com/firstornew-firstorcreate-firstor-updateorcreate

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -598,6 +598,33 @@ You may also come across situations where you want to update an existing model o
         ['price' => 99, 'discounted' => 1]
     );
 
+#### `firstOr`
+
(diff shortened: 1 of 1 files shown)


andrey-helldar:
I just found a mention of this method through a file search: https://laravel.com/docs/7.x/eloquent#retrieving-single-models

At the same time, the site search does not show information about this method:

taylorotwell:
If this is already documented it seems odd to document it in two places?

andrey-helldar:
@taylorotwell, I agree. It’s seems.
In this case, is it possible to correct the site search mechanism so that it finds this method?
I think many before that article simply did not even pay attention to him, given the fact that phpStorm sees it:

PR #6017, opened 2020-04-28
[7.x] Add note about increments event behavior

driesvints:
We've had quite a bit of people who were thrown off by this behavior so I tink it's best we should document it.


taylorotwell:
Maybe something worth fixing tbh.

PR #6021, opened 2020-05-02
[7.x] Trusted certificate for ssl

marcwieland95:
Add info to add the Homestead's CA to your local machine to make SSL work


taylorotwell:
I don't think this needs to be in the Homestead documentation.

marcwieland95:
@taylorotwell Why not? It's a crucial step to make SSL working. This part is pretty bad documented

PR #6022, opened 2020-05-03
[7.x] Updated Fluent matchAll Anchor

alex-yark:
The link in the documentation pointed to matchAll when it should have been match-all - corrected this in line with other anchor tags.


driesvints:
Thanks!

PR #6023, opened 2020-05-03
[7.x] Update upgrade.md - remove null in session config

ashgibson:
The upgrade docs from version 6 to 7 mention this:

> Next, please update your `session` configuration file's `secure` option to have a fallback value of `null`:
> 
>     'secure' => env('SESSION_SECURE_COOKIE', null),

However when viewing the version 7 session.php field, the null is not there. Is the documentation incorrect?


taylorotwell:
It's fine as-is. `null` is the default if the environment variable is missing so new 7.x applications don't exactly need it... but old applications may have been using `false`.

PR #6035, opened 2020-05-08
[7.x] Method order fixed

andrey-helldar:
Now the order of displaying methods is as follows:
```
$response->ok() : bool;
$response->failed() : bool;
$response->successful() : bool;
$response->serverError() : bool;
$response->clientError() : bool;
```
It looks nice, but the failed method breaks the combination of methods `ok` and `successful`.

I think it would be better to make it like this:
```
$response->ok() : bool;
$response->successful() : bool;
$response->failed() : bool;
$response->serverError() : bool;
$response->clientError() : bool;
```
This code looks logically correct.


driesvints:
Thanks

PR #6038, opened 2020-05-09
Added middleware note to Cookie::queue

tamilps2:
Just a little note for the Cookie queue method.


taylorotwell:
It is there by default.

PR #6039, opened 2020-05-09
Update testing.md

chrisijoyah:
This needs to extend the TestCase in the application.


driesvints:
It doesn't. Only feature tests.

PR #6041, opened 2020-05-11
[7.x] Facade singleton behaviour explanation

forrestedw:
Propose adding a note on Facades returning singleton even if `bind` is used in the packages service provider, as discussed in [this thread](https://github.com/laravel/ideas/issues/1088#issuecomment-535872736).


taylorotwell:
Probably should be documented under facades.

lloricode:
Hi, seems this is not already in live in laravel.com in both 7 and 8
https://laravel.com/docs/7.x/packages
https://laravel.com/docs/8.x/packages

PR #6043, opened 2020-05-11
[7.x] New blade if example

driesvints:
Since https://github.com/laravel/framework/pull/32742 got merged it's best that we update this example.


mikemand:
The closing tags need fixing also. They are still `@endenv`.

PR #6050, opened 2020-05-13
Add note to Eager loading documentation

infernobass7:
Added a note to Eager Loading section to make user's aware that if you do not set the $keyType for a model, using non-integer keys, the default behavior will retrieve all models from the database.


taylorotwell:
I think I will just leave this as-is. Reason being is that probably lots of things go wrong if you do not set the keyType properly based on the documentation and if we added a note everywhere it would be overwhelming.

PR #6053, opened 2020-05-14
Fix incorrect doc block in Blade Templates under casing

CreepPork:
There was an unnecessary ` * @param  string  $message` entry in the docblock which is incorrect.


driesvints:
Thanks

PR #6055, opened 2020-05-18
Remark about preserved keys in the filter() method

chimit:


taylorotwell:
I'm not sure this needs to be noted here. Most collection methods do.

PR #6063, opened 2020-05-19
Add a note for Elasticsearch version format on Homestead v10.3.0 and later

milepanic:
Add a note for Elasticsearch version format so it can be installed successfully by Homestead v10.3.0 or later


taylorotwell:
I don't like filling up the docs with notes like this.

PR #6074, opened 2020-05-24
[7.x] Update path customization in Authentication document

pranayaryal:
Include path for `AuthenticatesUsers` trait


pranayaryal:
Even if you plan to not include path, please mention that the `authenticated()` method is in `AuthenticatesUsers` trait. thanks.

PR #6076, opened 2020-05-26
[7.x] Clarify authorizeResource method usage

bomshteyn:
authorizeResource doesn't require the second argument. So having it there in the documentation is confusing especially in a case where the snake_case  of the given Model name would be the same as 2nd argument. 

Potentially, if the point of having it there is to highlight that this snake_case guessing can be overridden, we should add another example where the Model name and the request parameter are not the same.


taylorotwell:
I think the existing documentation is fine.

PR #6077, opened 2020-05-26
Update eloquent.md

joelmellon:
As far as I can tell, there's never been a `booted()` method on `Illuminate\Database\Eloquent\Model`. `Illuminate/Foundation/Application::booted()` exists though, so maybe that's the source of this documentation error.

There are lots of code samples out there that list my changes as the correct way to add global scopes. I've also tested and have it working this way in my own code. Since there's no `booted()` method, I assume registering event listeners should be done similarly, so I've updated that here as well.

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -767,7 +767,7 @@ Writing a global scope is simple. Define a class that implements the `Illuminate
 
 #### Applying Global Scopes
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Yes, there is on 7.x.

image

joelmellon:
🤦🏻‍♂️ Thought I was working in a 7.x project when it was actually running 6.x.

PR #6079, opened 2020-05-28
[7.x] State how to upload to the public directory directly

CreepPork:
I've always had trouble understanding the `File upload` documentation and these changes make things less ambiguous when it comes to `avatar` and `avatars`. Also by adding the `public/` this makes it far more useful in the real world because you want to upload the files in the `storage/app/public` directory so you can access it from the `public/storage` directory symlink.

It now explicitly states if you don't want to upload to the `public` directory but in my view, this should be a secondary item.


taylorotwell:
That's not really what I would recommend. 

I would recommend using this and specify the `public` disk:

image

CreepPork:
It seems like I wasn't aware that the default driver was set to `local`. This would then move the file to the `storage/app/avatars` directory. Maybe a better description to use the `public` disk in the `File Upload` section so it makes more sense in reality? Anyways, learned something :smile:

PR #6083, opened 2020-06-01
[7.x] Cashier Paddle docs

driesvints:


taylorotwell:
I wonder if there is any benefit to making `@paddleJS` be `@paddleScripts` to have consistency with how Livewire does that in the ecosystem.

driesvints:
@taylorotwell I would leave this as all the Paddle docs constantly refer to Paddle JS which this actually represents. It's best to keep the terminology the same imo.

cashier-paddle.md
```diff
@@ -0,0 +1,859 @@
...
+
+If you don't want to make use of the overlay checkout widget, Paddle also has an option to display an inline checkout. While you cannot properly adjust any of HTML fields (it's basically the checkout widget but inline) it allows you to embed it within your app's look and feel.
+
+To make it easy for you to get started with inline checkout we've shipped a `paddle-checkout` Blade component for you to use. You can generate a pay link just like you do with the overlay widget and pass that to the component's `override` attribute:
+
+    <x-paddle-checkout :override="$payLink" class="w-full" />
+
+Since `height` is a preset value for inline checkouts you'll need to pass it explicitly as an attribute if you want to adjust it:
+
+    <x-paddle-checkout :override="$payLink" class="w-full" height="500" />
```
taylorotwell, line 167: What is the meaning of "override" here?

cashier-paddle.md
```diff
@@ -0,0 +1,859 @@
...
+You can check out [the API documentation on pay link generation](https://developer.paddle.com/api-reference/product-api/pay-links/createpaylink) for more info.
+
+<a name="inline-checkout"></a>
+### Inline Checkout
+
+If you don't want to make use of the overlay checkout widget, Paddle also has an option to display an inline checkout. While you cannot properly adjust any of HTML fields (it's basically the checkout widget but inline) it allows you to embed it within your app's look and feel.
+
+To make it easy for you to get started with inline checkout we've shipped a `paddle-checkout` Blade component for you to use. You can generate a pay link just like you do with the overlay widget and pass that to the component's `override` attribute:
+
+    <x-paddle-checkout :override="$payLink" class="w-full" />
```
taylorotwell, line 163: What is the meaning of "override" here?

driesvints: It conforms to the option "override" on checkout to basically set up the checkout widget using a pay link.

https://developer.paddle.com/reference/paddle-js/parameters#checkout-properties

Screenshot 2020-06-02 at 17 03 32

cashier-paddle.md
```diff
@@ -0,0 +1,861 @@
...
+
+In light of this behavior, there are some important things to keep in mind when using Cashier and Paddle. First, when a new subscription is created we will save the `customer_email` value in the database as the user's `paddle_email` column value. **It is extremely important that you do not modify this value.** Cashier will use the `paddle_email` value for every new subscription and [pay link](#pay-links) to ensure all transactions are linked to the same customer within Paddle. Furthermore, this will ensure that you can assign multiple Paddle subscriptions to a single user using Cashier. A caveat of this is that any [override of the `customerEmail` method](#customer-defaults) on a billable user won't work anymore after the initial subscription has been created.
+
```
taylorotwell, line 192: What do you mean by "override of customerEmail method on a billable user won't work anymore"?

driesvints: In the section of "Customer Defaults" I explain that you can set a couple of methods that'll pre-fill the checkout widget for a customer so they can immediately move to the payment information part.

What I explain above is that that default will be ignored after the first subscription has been created. It's vital that the original email address on the user is used to create any subsequent subscriptions and payments so they're all attached to the same user within Paddle.

PR #6095, opened 2020-06-08
Fixed GlobalScope applying section in docs

2hamed:
The correct method is `boot` not `booted`.


driesvints:
No it's not.

2hamed:
Well the `booted` method is not working. Nothing happens when I use the `booted` method.

PR #6103, opened 2020-06-11
[7.x] Documenting the Queue Job retryAfter method

Farrow:
Adding documentation for the Queue retryAfter method that was (re)-added for 7.x as mentioned in https://github.com/laravel/framework/pull/32370 but is not mentioned in the docs


driesvints:
@Farrow the example you shared here is incorrect. You can't call `->attempts()` here since the `retryAfter` method will only be invoked for the first job.

PR #6106, opened 2020-06-11
[7.X] Use foreignId() instead of unsignedInteger()

EGA-SUPREMO:
I know this change is unnecessary but I think using the alias is more appropriate as it makes it easier to read.


taylorotwell:
It wouldn't match the actual stub created by the framework. So, we may want to update that first.

EGA-SUPREMO:
@taylorotwell ok, I want to contribute, so where should I look at?

PR #6109, opened 2020-06-12
[7.x] Clarify usage of pay link

driesvints:
Since quite a few people have already reported that they were confused with how pay links worked I'm hoping to make things clearer by added these snippets to a few places.


alzaabi98:
> Since quite a few people have already reported that they were confused with how pay links worked I'm hoping to make things clearer by added these snippets to a few places.

much clear now

PR #6110, opened 2020-06-12
Update mail.md with correct markdown button colors

xar:


driesvints:
They're aliases.

xar:
I don't think so, `primary` / `success` / `red` doesn't work on fresh l7 installation. Also when you open `mail/html/button.blade.php` there is `{{ $color ?? 'blue' }}`

PR #6112, opened 2020-06-14
[7.x] Mention laravel-websockets on the broadcasting page

ldiebold:
It's probably worth mentioning [beyondcode/laravel-websockets](https://github.com/beyondcode/laravel-websockets) on the broadcasting page. It's written in PHP, seems to be stable and feels like a more "Laravel" solution than laravel-echo-server.


taylorotwell:
Thanks for the idea. I have added this to a few places in the docs.

PR #6117, opened 2020-06-16
[7.x] Warn about possible cookie problems

bertheyman:
Follow up for https://github.com/laravel/framework/issues/33215

TL;DR: Using dump sets headers, which might silent fail for setting cookies. (as they can't be set when headers are already sent)

This PR hopes to make possible cookie problems more clear and prevent people getting stuck:
https://laracasts.com/discuss/channels/laravel/cookie-setting-doesnt-work-if-i-use-dump-function
https://stackoverflow.com/questions/45207485/how-to-set-and-get-cookie-in-laravel

Feel free to request other wording or other tweaks, happy to adjust.

### DIFF
--- a/responses.md
+++ b/responses.md
@@ -87,6 +87,8 @@ The `cookie` method on response instances allows you to easily attach cookies to
     return response($content)
                     ->header('Content-Type', $type)
                     ->cookie('name', 'value', $minutes);
+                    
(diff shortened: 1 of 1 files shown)


taylorotwell:
Even cookies set before a dump function won't be sent either.

bertheyman:
Would you like me to update the warning, or do you prefer to not have this inside the docs?

PR #6118, opened 2020-06-16
[7.0] Cashier Paddle: Use amount() method instead of total()

stefanzweifel:
In the Cashier Paddle docs the not available `total()` method is referenced.
Switching to `amount()` works as expected.


driesvints:
Thanks!

PR #6125, opened 2020-06-19
[7.x] Add note about Tinker error handling

PaulBehrendtVentoro:
Sorry my english is quite german.

See https://github.com/laravel/tinker/issues/107


taylorotwell:
Hmm, I'm just not sure if we should document this. I'm not sure if Tinker is really the place you should be "noticing bugs" and that that is better done in tests? With this comment it's just not clear what the reader is supposed take from it.

PaulBehrendtVentoro:
This works the other way around.
My tests don't prove anything for Tinker.
So for me this means, i can't use Tinker in Production. But this is not stated in the docs.

```
class ArgumentationHelper
{
    public static function handle(): bool
    {
        try {
            static::getWarning();
        } catch (\ErrorException $e) {
            return true; // WHEN USING PHPUNIT
        }
        return false; // WHEN USING TINKER
    }

    private static function getWarning()
    {
        \trigger_error('wanna handle?', E_USER_WARNING);
    }
}
```

PR #6127, opened 2020-06-22
[7.x] Replace plain string class names with actual references.

royvanv:


taylorotwell:
It is longer in these cases.

royvanv:
It's slightly longer (5 characters), but why is that the issue you have with it? I believe writing the class references this way makes it easier to maintain the code.

driesvints:
@royvanv for docs themselves this reads easier as it's fewer code to parse.

PR #6129, opened 2020-06-23
Added missing available method

chinleung:
The `add` method was missing from the list of available methods.


taylorotwell:
The `push` method is already documented.

chinleung:
@taylorotwell But this is the `add` method though. The `push` allows you to add more than one item at a time, so there's a slight difference. 🧐

PR #6136, opened 2020-06-27
Add instruction to register Horizon provider

shalvah:
The package's provider gets auto-registered, but the one published in the user's `app/Providers` doesn't. Easy to forget that.


driesvints:
It gets added automatically to your `app.php` file. Not sure why it doesn't works for you.

PR #6140, opened 2020-06-28
TelescopeServiceProvider registered twice in example

jackcruden:
As per title, I've removed one of the mentions and left in the more descriptive reference.


driesvints:
The docs are correct. The example registers the provider from the package and the other one in your `App\Providers` directory.

PR #6143, opened 2020-06-29
[7.x] Add method toQuery

ErickZH:
`toQuery` method added at https://github.com/laravel/framework/pull/33356


taylorotwell:
Fixed example since it was only using one model for example and not returning a collection at all?

PR #6146, opened 2020-06-30
[7.x] Add Explicit Warning of Exposed Endpoint in Wrong Environment

martbock:
Since it was previously pointed out that the default environment in the `.env.example` files will not change in the near future, I think we should at least warn users more explicitly about potential exposure of the Telescope endpoint in production environments if they set their environment incorrectly.

The previous discussions are located here:
- https://github.com/laravel/laravel/pull/5331
- https://github.com/laravel/framework/pull/33384

I hope this is somewhat of a compromise between what @lupinitylabs and I were proposing and what the position of @driesvints and the Laravel team on this issue is.

If you disagree with the exact wording, I'm open to suggestions, of course.

### DIFF
--- a/telescope.md
+++ b/telescope.md
@@ -115,7 +115,11 @@ By default, all entries older than 24 hours will be pruned. You may use the `hou
 <a name="dashboard-authorization"></a>
 ## Dashboard Authorization
 
-Telescope exposes a dashboard at `/telescope`. By default, you will only be able to access this dashboard in the `local` environment. Within your `app/Providers/TelescopeServiceProvider.php` file, there is a `gate` method. This authorization gate controls access to Telescope in **non-local** environments. You are free to modify this gate as needed to restrict access to your Telescope installation:
+Telescope exposes a dashboard at `/telescope`. By default, you will only be able to access this dashboard in the `local` environment. 
+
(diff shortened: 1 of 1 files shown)


lupinitylabs:
While I had a different concern in mind (safe defaults vs. misconfiguration), a more explicit documentation may be helpful. Although I really think that the root cause is the user/developer not being thoughtful about deploying their app/.env in the first place and it is my belief that they won't read the documentation right when deploying if they are not even reading through their .env...

driesvints:
This PR/note generally seems fine by me. 

About people not reading the docs: you can't expect to start driving a car without reading and learning traffic laws first. And we can't protect developers from themselves if they blindly deploy their apps into production. Just my 2 cents.

martbock:
> Although I really think that the root cause is the user/developer not being thoughtful about deploying their app/.env in the first place 

I get your point, but especially if one is an inexperienced developer and maybe hasn't deployed any of their apps before, I can see how they can miss this just because they didn't know this was something they should pay attention to. Confusion doesn't have to be thoughtlessness.

> it is my belief that they won't read the documentation right when deploying if they are not even reading through their .env...

Well, maybe that's true for most cases. But even if this change only helps one person out there to prevent a severe security issue in their app because they saw the red box and reflected for a moment, it's a success in my book.

PR #6147, opened 2020-06-30
UploadedFile reference to Symfony docs updated from 3.0 to 5.0 

stemis:


driesvints:
Adjusted this to point to master so it's always up to date.

PR #6160, opened 2020-07-04
fix route parameter names

AliSaleem27:
The function arguments do not match the route parameters.


taylorotwell:
They don't have to match.

PR #6169, opened 2020-07-09
[7.x] Fix Notification::assertToSent method arguments.

brian-dupont:
According to source code the method signature is: assertSentTo(mixed $notifiable, string $notification, callable $callback = null).

In the mocking documentation it is being used with a callable as the second argument and no third argument.


taylorotwell:
Second argument isn't necessary. Type will be read from the Closure's type hint.

PR #6177, opened 2020-07-14
Corrected wrong comparison operator in docs

mbklnd:
Updated the docs to show the correct way of setting a new attribute in the appends section.


driesvints:
The example is correct. It needs to return a boolean.

PR #6179, opened 2020-07-17
[8.x] Broad directory permission to bootstrap

gmsantos:
This may be necessary to avoid issues with new artisan down command introduced by https://github.com/laravel/framework/pull/33560

From the original PR:

> In addition, a new bootstrap/maintenance.php file is written. This file is loaded before Composer in the framework's index.php file and is able to render the template and exit with zero dependencies. Composer and Laravel are never loaded.

Another option would be to move `bootstrap/maintenance.php` to `bootstrap/cache/maintenance.php` (no permissions changes required here) or include a note in [Maintenance Mode](https://laravel.com/docs/master/configuration#maintenance-mode) docs.

### DIFF
--- a/installation.md
+++ b/installation.md
@@ -83,7 +83,7 @@ All of the configuration files for the Laravel framework are stored in the `conf
 
 #### Directory Permissions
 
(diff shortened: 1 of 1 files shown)


driesvints:
Think we probably want to indeed move this to `bootstrap/cache` or even better to `storage/framework` like the previous one.

taylorotwell:
I'll move it to the cache directory.

driesvints:
FYI: this was moved to `storage/framework/maintenance.php` eventually.

PR #6180, opened 2020-07-17
Update notifications.md

futuremeng:
Add use App\Invoice and __construct to avoid "undefined $invoice"


driesvints:
The "..." hints at more code being present.

PR #6181, opened 2020-07-17
[7.x] Add: updating/updated only fired if model is dirty

joostvanhoof:
I recently found out that `updating`/`updated` are only fired if `save()` is used _and_ the model is dirty, as can be seen [here](https://github.com/laravel/framework/blob/7.x/src/Illuminate/Database/Eloquent/Model.php#L805).

Currently the docs only mention that they're fired if the `save()` is used, which I found confusing as I did use `save()` but `updating`/`updated` were still not fired.

I found out after posting this on SO, [see here](https://stackoverflow.com/questions/62936862/laravel-deleting-cache-working-in-controller-but-not-in-model-closure).


taylorotwell:
Fixed typos.

joostvanhoof:
I think something went wrong, the docs currently say (highlight mine):

> The updating / updated events will fire when an existing model is modified **_and and_** the save method is called

On https://laravel.com/docs/7.x/eloquent#events

PR #6186, opened 2020-07-18
[7.x] Update mail.md

CiiDyR:
Add some missing DocBlock parameters.

### DIFF
--- a/mail.md
+++ b/mail.md
@@ -202,7 +202,7 @@ Typically, you will want to pass some data to your view that you can utilize whe
         /**
          * Create a new message instance.
          *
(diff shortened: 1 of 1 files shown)


driesvints:
Please use FQN, thanks.

CiiDyR:
Shouldn't @var also use FQN then? This is why I didn't write the @param's as FQN

driesvints:
@CiiDyR it should yes. All docblocks across all Laravel projects should use FQN.

CiiDyR:
I've created a new PR (#6189) that updates the docs where it was missing FQN.

PR #6189, opened 2020-07-20
[7.x] Add FQN to DocBlocks

CiiDyR:
This pull request would change a lot of DocBlock definitions to use FQN instead of relative namespaces.
It also adds a lot of `use` statements to code snippets.

When I need to reference the Laravel docs, I often find myself searching for the namespace of, for example, facades used in a tiny snippet. This would allow someone to just instantly see the namespace for importing in a project.

I still have to go through the official packages docs, but wanted to get thoughts on this first.


driesvints:
Heya, thanks for your work. All the extra use statements for the facades probably aren't necessary. Also: this PR is way too large to review so it's probably best that you send this in batches.

CiiDyR:
@driesvints I could break it down into the sidebar hierarchy, how about that?
I think the added convenience is worth it. But if you think it's overkill I can just leave them out in the next PR(s)

driesvints:
I'll let @taylorotwell decide on that.

taylorotwell:
Yeah, we don't necessarily blanket use statements like this. Usually in the flow of documentation I will show the import in the first example but not on subsequent examples.

driesvints:
@CiiDyR so definitely feel free to send in the FQN fixes but without the import statements. Maybe best in separate PRs that aren't too big to review.

CiiDyR:
@driesvints Sure thing. Good to know for future contributions. I'll chunk them into a reviewable size :)

PR #6191, opened 2020-07-21
[7.x] Update docs for beta 2

driesvints:
In draft until beta 2 is released.

### DIFF
--- a/cashier-paddle.md
+++ b/cashier-paddle.md
@@ -36,7 +36,7 @@
     - [Simple Charge](#simple-charge)
     - [Charging Products](#charging-products)
     - [Refunding Orders](#refunding-orders)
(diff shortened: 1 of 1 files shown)


cashier-paddle.md
```diff
@@ -94,7 +94,7 @@ Before using Cashier, you must add the `Billable` trait to your model definition
         use Billable;
     }
 
-The trait can be implemented on other types of models in the same app as well:
+If you have billable entities that are not users, you may add the trait to those classes:
```
driesvints, line 97: Think the following keyword indicates more that the trait can be used in combination with other models

```suggestion
If you have billable entities that are not users, you may also add the trait to those classes:
```

PR #6196, opened 2020-07-24
[7.x] Correcting description for whereDoesntHave

DelightedD0D:
The description of the example using whereDoesntHave with a nested relationship did not correctly describe the behavior of the function. Updated with correct description.

Originally brought to attention by 0xN0 via this [SO Question](https://stackoverflow.com/questions/63064693/laravel-eloquent-wheredoesnthave-documentation-confuses-me)

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -977,7 +977,7 @@ If you need even more power, you may use the `whereDoesntHave` and `orWhereDoesn
         $query->where('content', 'like', 'foo%');
     })->get();
 
-You may use "dot" notation to execute a query against a nested relationship. For example, the following query will retrieve all posts with comments from authors that are not banned:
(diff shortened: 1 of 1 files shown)


MitchTalmadge:
Am I misunderstanding something? The SO question and the original change to the wording (24cc0f0) say that this query will:
- Fetch all posts with 0 comments AND,
- Fetch all posts where ALL comment authors are banned.

> ...the following query will retrieve all posts that have no comments, or where all comments are from banned authors:

But commit 50303d1 overwrites that wording change to suggest that the query will:
- Fetch all posts with 0 comments AND,
- Fetch all posts where AT LEAST ONE comment has an author that IS NOT banned.

> ...the following query will retrieve all posts that do not have comments and posts that have comments from authors that are not banned:

These are two completely different explanations. Which is correct? Why did this get changed & merged without discussion?

Tangent to this -- what is the query trying to practically show, anyway? Why would you want to fetch posts with 0 comments or posts where a comment has not been banned? I feel like this entire situation could be resolved with a better example.

PR #6197, opened 2020-07-24
[8.x] Bump minimum supported sqlite version

driesvints:
Bumping the minimum supported version because https://github.com/laravel/framework/pull/33618 was merged.

We'll also need to note this in the upgrade guide.


driesvints:
This is only needed for the feature in the linked PR.

PR #6199, opened 2020-07-25
[7.x] Document appending a further eager load on a subquery

turbo124:


taylorotwell:
This doesn't seem necessary to add here.

PR #6201, opened 2020-07-25
fixed typo in blade.md

bhushangaykawad:
It is like this 

```blade
@section('sidebar')
            This is the master sidebar.
@show
```

should be 

```blade
@section('sidebar')
            This is the master sidebar.
@endsection
```


driesvints:
Thanks!

PR #6203, opened 2020-07-26
[7.x] Added missing docs for CORS header 'Access-Control-Allow-Origin'

techieni3:
I have create SPA outside laravel project and want to communicate for SPA authentication. I am getting ` The Same Origin Policy disallows reading the remote resource` error. 

image

To solve it issue we need to add routes that we need want to access from SPA for authentication in `cors` config file as `['api/*','sanctum/csrf-cookie','login','logout']`.


taylorotwell:
I don't think this is necessary in all cases. I know I have not needed to add it before. Maybe we need a little more info clarifying when it is needed?

PR #6204, opened 2020-07-27
[7.x] date serialization bit removed as no longer relevant

tanerkay:
removed bit about `$dateFormat` being for date serialization as it is now dictated by `serializeDate()`

disclaimer: Not 100% about this


taylorotwell:
Missing word off the front?

PR #6206, opened 2020-07-27
Fix code example for custom casts

AlexVanderbist:
This typo was already fixed on the 7.0 branch via #6182. Not sure if it needs to be on master as well, but here it is.


driesvints:
@AlexVanderbist 7.x gets merged into master regularly but apparently this time it's been a while.

PR #6207, opened 2020-07-28
Fix incorrect Artisan schedule method name

dmlogic:
This PR corrects the schedule `monthlyOnLastDay` to `lastDayOfMonth`. Attempting to use the schedule as documented results in an exception

Screenshot from 2020-07-28 07-21-08


driesvints:
Thanks!

PR #6211, opened 2020-07-29
Add saved to the model observer

connor11528:
There is a saved observer that isn't referenced in the docs. This PR adds the saved function so people are aware of the `saved` observer function :)


taylorotwell:
That isn't in the file by default.

PR #6213, opened 2020-08-01
Described the REDIS_PREFIX

Back1ng:
I think it can be described better.


taylorotwell:
I'm not sure the flow of this totally makes sense. It reads a bit confusing.

PR #6217, opened 2020-08-03
[7.x] Update eloquent.md

driesvints:
Warns about overriding `query` and `newQuery`. I wasn't sure of either adding a note or a new section on the usage of `query` & `newQuery` so feel free to provide feedback about that.

See https://github.com/laravel/framework/issues/33695

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -237,6 +237,8 @@ Once you have created a model and [its associated database table](/docs/{{versio
     foreach ($flights as $flight) {
         echo $flight->name;
     }
+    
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think I would maybe rather just mark the methods with some @internal notification.

jonnott:
> I think I would maybe rather just mark the methods with some @internal notification.

@taylorotwell  Should we address the possibility of marking @internal on https://github.com/laravel/framework/issues/33695 then?

driesvints:
@jonnott we haven't had time yet to look at that further but we want to find the root cause of that first before we do any change.

PR #6219, opened 2020-08-03
[7.x] Offer grammar alternative in hasOneThrough section

yabbou:
old: the `Mechanic` can access to the `Owner` _through_ the `Car` itself
new: the `Mechanic` can access the `Owner` _through_ the `Car` itself

There are other ways to rephrase to sentence, but currently seems off to me... best of luck!


driesvints:
Thanks!

PR #6220, opened 2020-08-04
Remove Extending trials and past-due-status

Perafan18:
- Remove `Extending Trials` from the table of contents because the section does not exist, that method does not exist in the package.
- Remove the reference to `past-due-status` in the table of contents that does not show H4 titles.


driesvints:
Thanks!

PR #6224, opened 2020-08-04
[7.x] Update wording for Cashier Paddle docs

crnkovic:
You had a missing word in the first sentence:

> Sometimes, payments for subscriptions.

I've added just a bit of explanation on when subscription payments can fail.


driesvints:
Thanks! :)

PR #6229, opened 2020-08-05
Update middleware.md

nmokkenstorm:
This is the only h3 on the page that's currently not targetable by a hash in the url


taylorotwell:
I'll just change it to a h4 instead.

PR #6230, opened 2020-08-06
[7.x] Suggest asset publishing Composer hook

driesvints:
This PR adds the suggestion of hooking up the `telescope:publish` and `horizon:publish` commands to the app's `composer.json`'s `post-autoload-dump` list. This will make sure that assets are always kept up to date and prevents breakage of the dashboards. This was inspired by Livewire's docs: https://laravel-livewire.com/docs/installation

Added a new "Upgrading Telescope" section to be in line with other first party docs.

I've also moved "Dashboard Authorization" in the Telescope docs one level down to be in sync with Horizon's docs.

### DIFF
--- a/horizon.md
+++ b/horizon.md
@@ -102,7 +102,7 @@ Horizon exposes a dashboard at `/horizon`. By default, you will only be able to 
 
 When upgrading to a new major version of Horizon, it's important that you carefully review [the upgrade guide](https://github.com/laravel/horizon/blob/master/UPGRADE.md).
 
-In addition, you should re-publish Horizon's assets:
+In addition, when upgrading to any new Horizon version, you should re-publish Horizon's assets:
 
     php artisan horizon:publish
     
--- a/horizon.md
+++ b/horizon.md
@@ -110,9 +110,7 @@ To keep the assets up-to-date and avoid issues in future updates, we highly reco
 
     {
         "scripts": {
-            "post-autoload-dump": [
-                "Illuminate\\Foundation\\ComposerScripts::postAutoloadDump",
-                "@php artisan package:discover --ansi",
+            "post-update-cmd": [
                 "@php artisan horizon:publish --ansi"
             ]
         }
--- a/horizon.md
+++ b/horizon.md
@@ -105,6 +105,18 @@ When upgrading to a new major version of Horizon, it's important that you carefu
 In addition, you should re-publish Horizon's assets:
 
     php artisan horizon:publish
+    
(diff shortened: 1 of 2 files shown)


crynobone:
I normally set it under `post-update-cmd` since assets would just changed on `composer update`

Screenshot 2020-08-06 at 6 19 41 PM

I would understand that Livewire might need to republish if a new component is created etc (guessing here).

driesvints:
@crynobone does that also work if you, for example, use `composer require` with the `--update-with-dependencies` and --update-with-all-dependencies` flags?

crynobone:
`--update-with-dependencies`

Screenshot 2020-08-06 at 6 50 20 PM

`--update-with-all-dependencies`

Screenshot 2020-08-06 at 6 51 20 PM

Yup, both triggered `post-update-cmd`.

driesvints:
@crynobone Cool, thanks for verifying 🙂 

Update the PR.

horizon.md
```diff
@@ -102,9 +102,19 @@ Horizon exposes a dashboard at `/horizon`. By default, you will only be able to
 
 When upgrading to a new major version of Horizon, it's important that you carefully review [the upgrade guide](https://github.com/laravel/horizon/blob/master/UPGRADE.md).
 
-In addition, you should re-publish Horizon's assets:
+In addition, when upgrading to any new Horizon version, you should re-publish Horizon's assets:
 
     php artisan horizon:publish
+    
+To keep the assets up-to-date and avoid issues in future updates, we highly recommend adding the command to the `post-autoload-dump` scripts in your `composer.json` file:
```
crynobone, line 109: `post-autoload-dump` should be updated to `post-update-cmd`

telescope.md
```diff
@@ -134,6 +129,23 @@ Telescope exposes a dashboard at `/telescope`. By default, you will only be able
...
 > {note} You should ensure you change your `APP_ENV` environment variable to `production` in your production environment. Otherwise, your Telescope installation will be publicly available.
+    
+<a name="upgrading-telescope"></a>
+## Upgrading Telescope
+
+When upgrading to a new version of Telescope, you should re-publish Telescope's assets:
+
+    php artisan telescope:publish
+
+To keep the assets up-to-date and avoid issues in future updates, we highly recommend adding the command to the `post-autoload-dump` scripts in your `composer.json` file:
```
crynobone, line 140: Same here

horizon.md
driesvints: Good catch. Updated for both.

PR #6233, opened 2020-08-07
[7.x] Remove use AgeScope that isn't used in example

dinandmentink:
Very small thingy. The `use App\Scopes\AgeScope;` is not used in the example. The statement is probably there because it was copied over from another example.


driesvints:
Thanks!

PR #6234, opened 2020-08-07
[7.x] Added Zip PHP Extension to requirements list

mnaderian:
To install Laravel installer the ZIP PHP Extension is needed. It was not mentioned in the list of requirements.


driesvints:
The installer needs this. Not the framework.

mnaderian:
So shouldn't this be mentioned in the installer section?

driesvints:
It's already required in its composer.json: https://github.com/laravel/installer/blob/master/composer.json#L14

You shouldn't be able to install it unless you're circumventing it somehow.

PR #6235, opened 2020-08-09
[7.x] Add Laravel Echo to Contribution Guide

peeley:
The current [list of GitHub repositories](https://laravel.com/docs/7.x/contributions#bug-reports) for Laravel projects is missing an entry for Laravel Echo.


driesvints:
Thanks

PR #6236, opened 2020-08-09
[7.x] Updated Eloquent address cast key.

davidrushton:
The custom `\App\Address` castable was using the `options` key, but I'm assuming this should be something like `address` for clarity?


driesvints:
Thanks!

PR #6242, opened 2020-08-12
Add `creating`, `updating`, `deleting` methods to sample observer

exzachlyvv:
Right now, the sample methods on the Eloquent model observer are sparse. I think we should build that out a bit more. There is no other place these are documented (unless I'm missing it).

There are also a few more niche ones we can add like `booting`, `booted`, `retrieved`, `saving`, `saved`, `replicating`


taylorotwell:
It matches the stub observer.

PR #6247, opened 2020-08-13
[7.x] Improve Horizon 'waits' example

hivokas:
Because "queue combination" is not quite easy to understand.


taylorotwell:
Does it actually work to use a comma in this spot?

hivokas:
@taylorotwell yes, which I actually don't like a lot, because this way I need to concatenate the list of queues for each supervisor.

image

image

PR #6254, opened 2020-08-19
fixed unrealistic age restriction in Defining Middleware example

WaiHtikeLin:
I fixed unrealistic age restriction in example at 'Defining Middleware' section in 'Middleware' topic because age > 200 is unrealistic to give an example.


taylorotwell:
🤷

PR #6262, opened 2020-08-23
[7.x]  Documentation for Route::resource() named routes

tauseefsshah:
There was no documentation for using named routes with resourceful routes and controller **[ Route::resource() ]**. This is a very useful feature to know, hope this helps aware developers and gives them the adequate knowledge about this feature.


taylorotwell:
This should probably be documented with resource controllers?

tauseefsshah:
Yes found them with the resource controller docs. I was looking for these in the url and route docs since others were there.
Thanks though 👍

PR #6263, opened 2020-08-24
Updating 'use Exception' to 'use Throwable' in queues

spekulatius:
Hello @driesvints,

just a super minor fix: In the queue documentation an example still included `Exception` while `Throwable` was used otherwise.

Cheers,
Peter


driesvints:
Thanks!

PR #6268, opened 2020-08-24
[5.7] Fix non-active cloudinary link for the Telescope screenshot

andresayej:
This PR depends on the following PR to be merged first:
laravel/laravel.com-next#128


andresayej:
@taylorotwell I think you need to trigger a new build for the 5.7 docs as the merged changes are not reflected on the live site (tried clearing the cache, as well as different browsers)

PR #6270, opened 2020-08-25
[7.x] Add methods cascadeOnDelete and cascadeOnUpdate

ErickZH:


taylorotwell:
I think the existing examples are OK.

ErickZH:
@taylorotwell I think that new users who come to the framework should know about the existence of these methods.

PR #6271, opened 2020-08-25
[7.x] Added example for unique rule validation

Rutaba:
Added a single line example for unique rule validation with ignoring value and multiple where conditions


taylorotwell:
Not needed since we show how to do it above.

Rutaba:
The example demonstrated under **Forcing A Unique Rule To Ignore A Given ID** section is with the use of **Rule Class**. Forcing to ignore a given ID/value along with the multiple where conditions, without using the **Rule class**, is not covered in the docs, and a simple user couldn't know if he also has an option do it in a single line.

PR #6282, opened 2020-08-28
[7.x] Replace "a third argument" with "a second argument"

paulandroshchuk:
I suppose we need to update the master branch as well?
I can do that once somebody approves this PR.


driesvints:
Thanks! We merge 7.x into master regularly so that's not necessary.

PR #6285, opened 2020-08-30
fixed the meaning of the sentence for Conditionally Queueing Listeners

siarheipashkevich:


taylorotwell:
It is not executed at all.

siarheipashkevich:
In this case the name of the method `shouldQueue` is not correct for this logic. IMHO: `shouldQueue` tells me that the job can be executed in the queue or right now.

PR #6290, opened 2020-09-01
Update comparison tool to show difference from 7.x.

Bradez:
Fixes the link in the Miscellaneous section for GitHub comparison tool to use a comparison between `7.x...master` rather than `6.x...master`.


driesvints:
Thanks!

PR #6292, opened 2020-09-01
[8.x] Set release and fixes until dates.

Bradez:
These are the same dates as defined in the 7.x branch but were left omitted from master.


driesvints:
Thanks

PR #6296, opened 2020-09-02
[7.x] Update eloquent-relationships.md

lsmj:
Updated the example with a pivot value that is available by default (created_at is not)


taylorotwell:
It's explained how to make things available directly below this example.

PR #6297, opened 2020-09-02
Use App\Models\User in the release notes examples

iruoy:
The examples for Model Factory Classes shouldn't use App\User since they're only available in Laravel 8 and the default directory for Models is App\Models in Laravel 8.


driesvints:
Thanks

PR #6309, opened 2020-09-07
[8.x] Add upgrade step for routes

jimgwhit:
Probably should be in upgrade guide.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -168,16 +168,13 @@ The paginator now uses the [Tailwind CSS framework](https://tailwindcss.com) for
 
 Routing now requires a use statement and class name resolution:
 
-```
     use App\Http\Controllers\UserController;
-```
+
 And
-```
+
     Route::get('/user', [UserController::class, 'index']);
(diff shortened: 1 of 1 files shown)


AegirLeet:
I think you're wrong - routes defined in Laravel 7 using `FooController@bar` syntax with `RouteServiceProvider::$namespace` set to something like `'App\Http\Controllers` will continue working fine in Laravel 8.

New Laravel 8 projects won't have `RouteServiceProvider::$namespace` set, but that's irrelevant for the upgrade.

taylorotwell:
If you're upgrading your RouteServiceProvider will still have the namespace.

snapey:
@jimgwhit Your policy of upgrading Laravel/Laravel as well as the framework means that you have to adopt the new approach since your upgrade will remove the namespacing of the controllers in the RouteServiceProvider.php

Anyone upgrading the framework only will retain their existing RouteServiceProvider, including the namespace statement, and their routes will continue to work.

PR #6318, opened 2020-09-08
explain concept of tags at the start before "binding typed variadics"

bilogic:
I was trying to read the docs head to tail and realized that ```tags``` it seemed a bit out of order


driesvints:
Please send this to 8.x

PR #6319, opened 2020-09-08
[8.x] explain concept of tags at the start before "binding typed variadics"

bilogic:
https://github.com/laravel/docs/pull/6318
better ordering to explain tags first


taylorotwell:
To me tagging is a bit of an advanced and lesser used concept. I think it's fine where it is.

PR #6330, opened 2020-09-09
Model name incorrect on some of the examples

colinhall17:
Post model name was missing from some of the examples


driesvints:
Thanks

PR #6331, opened 2020-09-09
[8.x] Temporary URLs in S3: Update request parameters example

MohannadNaj:
In Temporary URLs section, while explaining the third additional S3 Request parameters, The current example educates the user in a quick glance how to customize the `Content-Type` response to a different MIME type.

While this illustrates the customization ability, I believe an example of `Specifying A File Name` (which is already documented for file uploads) might be more useful.


taylorotwell:
Why not both?

MohannadNaj:
@taylorotwell Yes it would be better!

I thought it's intentionally meant to be one-line array.

Updated now.

PR #6335, opened 2020-09-10
Updated api array inside auth config

TechieForFun:
Added hash index


taylorotwell:
Unclear what this is for? Need some explanation.

TechieForFun:
When you create a new project it has a config option inside auth config which is `'hash' => false` by default. I added to document to prevent confusions.

PR #6337, opened 2020-09-10
[8.x] Use WithFaker

Perafan18:
👀  `WithFaker`


taylorotwell:
This documentation isn't correct. You don't need to do this.

PR #6341, opened 2020-09-10
[8.x] restore Password Confirmation docs

browner12:
I'm not sure why this was removed. Does Jetstream handle this some other way?


browner12:
please only merge this or #6342, depending on which direction you want to go.

PR #6342, opened 2020-09-10
[8.x] remove unused link to Password Confirmation

browner12:
does this not exist anymore?


browner12:
please only merge this or #6341, depending on which direction you want to go.

taylorotwell:
It exists but needs to be redocumented and updated.

PR #6352, opened 2020-09-11
[8.x] Added note regarding using in-memory database for dusk

agjmills:
Spent half a day trying to work this out, stumbled across this stackoverflow answer https://stackoverflow.com/a/50376781 - felt it should be inclued in the docs


driesvints:
Given that this pops up regularly I think a note could be useful yeah somewhere.

PR #6354, opened 2020-09-11
Added save method call when storing data to session

Vladimir5577:


manekshms:
yes without save method the data won't be store in the session, I also got stuck for sometime

taylorotwell:
Unnecessary. Your application has another problem.

PR #6356, opened 2020-09-12
add link to the public/index file for reference

gauravmak:


driesvints:
This references a specific commit which might show an outdated file when it gets updated in the future.

PR #6370, opened 2020-09-14
Fix sintaxe for Single Action Controllers in Laravel 8

henryavila:
Fix sintaxe for Single Action Controllers in Laravel 8


taylorotwell:
Unnecessary.

PR #6373, opened 2020-09-15
Update upgrade.md to link to tag 7.x of mail config

ShawnCorrigan:
While currently the difference between config/mail.php in Laravel 7 and the develop branch are the same, this could change in the future and could be easily forgotten about. This could prove confusing for users trying to upgrade from 6 to 7 if it is changed to something not compatible in the future.


driesvints:
Thanks!

PR #6374, opened 2020-09-15
Correct facade name in dispatching batches example

cyrrill:
The dispatching batches example incorrectly imports the facade as: `Illuminate\Support\Facades\Batch`, however, the correct name is `Illuminate\Support\Facades\Bus`.

This typo also causes a duplicate symbol exception with line immediately preceding it: `Illuminate\Bus\Batch`.


driesvints:
Thanks!

PR #6375, opened 2020-09-15
[8.x] Note about default url values and implicit binding

driesvints:
Clarify that the middleware for url defaults should be prioritized over the `SubstituteBindings` middleware.

See https://github.com/laravel/framework/issues/34342

### DIFF
--- a/urls.md
+++ b/urls.md
@@ -171,3 +171,19 @@ It is cumbersome to always pass the `locale` every time you call the `route` hel
     }
 
 Once the default value for the `locale` parameter has been set, you are no longer required to pass its value when generating URLs via the `route` helper.
+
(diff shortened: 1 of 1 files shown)


eballeste:
Hey, per my comments in the original issue, might it be better to do something like this in the middleware so that we don't have to force the middleware to be global and keep it as a route middleware?

This is what I did in my middleware to fix the issue but I feel like it would have been better to somehow handle this in the RouteUrlGenerator file since their seems to be an obvious conflict between using default route parameters that are also used with route model binding. 

```php
public function handle($request, Closure $next)
{
     $tenant_slug = is_string($request->tenant)
        ? $request->tenant : $request->tenant->slug;

     URL::defaults(['tenant' => $tenant_slug]);

     return $next($request);
}
```

driesvints:
Hey @eballeste. As I said in the other issue, the `$middlewarePriority` also works with route middleware. It only orders middleware, it doesn't adds it as a global one.

eballeste:
gotcha sorry about that, I've been using Laravel for about 2 months now and just read up on `$middlewarePriority`. This indeed solved my issue, thanks for your help! ✌️✌️✌️

PR #6388, opened 2020-09-17
Fix typo

kiiya:


yoeriboven:
It's the word `provide` which initially was `provider`. Kinda hard to see the difference since it's on one line. :)

taylorotwell:
Fixed.

PR #6391, opened 2020-09-18
[6.x] Fixes broken Passport API cookie security fix handling

HenkPoley:
Recently Passport 7.x, 8.x & 9.x added a CookieValuePrefix as a security measure. This broke cookie token handling deep inside `auth:api`.

Fix is in Passport 9.3.2. Please merge this documentation fix so people who upgrade to LTS get informed.

Ref: https://github.com/laravel/framework/issues/34227

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -46,6 +46,8 @@ PHP 7.1 will no longer be actively maintained as of December 2019. Therefore, La
 
 Update your `laravel/framework` dependency to `^6.0` in your `composer.json` file.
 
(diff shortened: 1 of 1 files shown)


HenkPoley:
For your information, this is now broken back to at least Laravel 5.4. Probably anybody with a pretty standard _authenticated API_ will bump into this once they run `composer update`, due to recent downstream dependency upgrades.

I won't try to fix more, since by Laravel policy these fixes will not be merged, so it would probably be useless.

PR #6397, opened 2020-09-21
[8.x] Add max-time to example Supervisor config

paras-malhotra:
Many of us simply copy paste the example Supervisor config mentioned in the Laravel docs. I think we should add the `max-time` option in the example Supervisor config as it's a good practice to avoid memory leaks.


themsaid:
Not sure really. People will need to customize the options in all cases so I don't think we should put all the options in this example. Leaving it to Taylor to decide.

PR #6399, opened 2020-09-21
[8.x] Clarify combining old and new model factories in upgrade guide

driesvints:
This isn't immediately clear it seems: https://github.com/laravel/framework/issues/34437


taylorotwell:
I'm not sure this is entirely accurate. You can put your new factories in a different location. They are just classes and will be autoloaded however you wish.

PR #6405, opened 2020-09-22
Fix $this->release docs

damcclean:
I've just spent a good 20 minutes trying to figure out why `$this->release(10)` wasn't working then I figured out I have to do `$this->job->release(10)` instead which solved the problem.

This documentation should probably be updated to help people like other developers.


taylorotwell:
Unnecessary. Queued jobs are generated with the `InteractsWithJob` trait included. That trait has a release method.

PR #6409, opened 2020-09-23
[8.x] Improve allOnQueue() / allOnConnection() upgrade notes

ramirovarandas:
Hi,

During the upgrade process, we've followed the guide, however we removed the references to the `allOnQueue()` and `allOnConnection()` from our code, but it that stopped the chain to process the next joib in the chain. 

Using these methods with the helper `dispatch()` returned the previous - correct - behaviour.

I don't know if this is going to be the way in the future, but just suggesting a note in the docs to help other people and avoid problems.

Thanks!

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -241,7 +241,7 @@ For consistency with other dispatching methods, the `allOnQueue()` and `allOnCon
         new ReleasePodcast
     ])->onConnection('redis')->onQueue('podcasts')->dispatch();
 
-The `allOnQueue()` and `allOnConnection()` still exists when using `dispatch()` helper, which will create an instance of `PendingDispatch`.
+Note that this change only affects code using the `withChain` method. The `allOnQueue()` and `allOnConnection()` are still available when using the global `dispatch()` helper.
 
 <a name="failed-jobs-table-batch-support"></a>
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can you provide more information on what you mean? The note you added isn't super clear to me.

ramirovarandas:
Hi Taylor,

Our application was using Laravel 7.x and we create a job class dynamically - user inputs params through a form - and based on these params we choose a specific job class, however some of these process require a chain, so we were using the `dispatch()` to dispatch all the jobs on the same queue / connection, like below:

```php
$job = new MyJobClass($params);

dispatch($job)->chain([
    function() {
         (new DoOtherThing())->handle();
    },
])->allOnQueue('my-queue');
```

But when we read the upgrade guide, we replaced the `allOnQueue()` call with `onQueue()` as described in the guide, however these made the application stop running the chain - even the tests passed - and when we reverted everything behavoured normally.

I apologize if the place / way I putted the note in the docs was not ok, but as I've never contributed to the guide I didn't know how / where to add it. Also I don't know if this should be in the guide / docs, we just suggested opening a PR to help / prevent others to go through the same problem.

It's Laravel teams call to accept these of course, but it was just a suggestion.

Thanks!

PR #6411, opened 2020-09-24
[8.x] Clarify that all types but enums can be renamed

perifer:
This changed in https://github.com/laravel/framework/pull/32205


perifer:
Just realized that this should probably change in 7.x as well, I can create another PR for that.

PR #6412, opened 2020-09-24
[8.x] Factory states: Add tip for using dynamic data

danabrey:
This is a gotcha with the new factories in 8.x. When generating dynamic data within a state transformation, e.g.:

```
public function anonymous()
    {
        $uuid = Str::uuid();
        return $this->state([
            'name' => 'anonymous_user_' . $uuid,
            'email' => $uuid . '@anonymous.com',
            'is_anonymous' => true,
            'password' => Hash::make(Str::random(10)),
        ]);
    }
```

The method only runs once however many models you're creating. So generating 100 instances of that state will create 100 models each with the same uuid.

Using a callback fixes that.

### DIFF
--- a/database-testing.md
+++ b/database-testing.md
@@ -158,7 +158,7 @@ If your state transformation requires access to the other attributes defined by 
         });
     }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Better we just show an example. I'll fix it.

danabrey:
That's fine too, just anything that points out the gotcha.

PR #6413, opened 2020-09-24
[8.x] Add Guzzle ^6.5 version constraint to upgrade

gtjamesa:
A tip should be added clarifying that Guzzle 6 is not unsupported as the updated dependency implies. As [pointed out by Graham](https://github.com/guzzle/command/pull/40#issuecomment-695216164), the docs are missleading and Laravel suggests the [`^6.5.5|^7.0.1`](https://github.com/laravel/framework/blob/8.x/composer.json#L135) constraint.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -66,6 +66,8 @@ In addition, the Laravel installer has been updated to support `composer create-
 
 Finally, examine any other third-party packages consumed by your application and verify you are using the proper version for Laravel 8 support.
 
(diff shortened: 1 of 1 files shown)


driesvints:
I don't recommend adding this to the upgrade guide as we're trying to encourage people to use newer Guzzle versions. You can adjust this in your own app if you need it.

decadence:
+1 for this because some of libraries still use Guzzle 6

driesvints:
@decadence you can simply change this in your composer.json. No one's restricting you there.

PR #6417, opened 2020-09-25
[8.x] Adds note about phpredis class conflict

stebogit:
The default `Redis` Facade name conflicts with the class of the phpredis PHP extension (ref https://github.com/laravel/framework/issues/32007#issuecomment-600441581). I encountered this issue updating Homestead recently and I felt this note could save somebody else some time.


taylorotwell:
The exception message we throw already tells you this is the case.

PR #6421, opened 2020-09-25
[8.x] Update homestead.md

xyluz:
bash init.sh work on windows too.


browner12:
on all versions? or just ones with their new WSL?

xyluz:
> on all versions? or just ones with their new WSL?

I'm not sure, I only have access to the new WSL.

PR #6422, opened 2020-09-25
Specify difference in update on Model and Builder.

sdawka:
Update on the builder (using `where`) instead of on the model (using `find`) bypasses the model convenience features such as guarded attributes, etc. The proposed change indirectly brings this fact to light, and can help avoid inadverdent security lapses.


taylorotwell:
I think this should be added to the body text somewhere - not as a note.

PR #6423, opened 2020-09-26
Additional clarity on create and update

sdawka:
Existence of Model's `create` method and caveats of two different `update` functions (on Model and Builder) made clearer without needing to refer to the API specs and source.


taylorotwell:
Wording here needs some work. I'll take a look.

PR #6429, opened 2020-09-29
[8.x] Add mention of undocumented `$response->object()` method

n1c:
This method exists but is not mentioned in the documentation.


taylorotwell:
I really think people should just use ->json()

n1c:
Converting a json string to an assoc array isn't even the default behaviour for `json_decode`, surely exposing the option is useful?

JohnRoux:
Surely if we support the method, we should document it?

PR #6432, opened 2020-09-29
[8.x] Document Artisan command events

codegefluester:
Long time listener, first time caller here. I was using those events the other day but couldn't find any docs on them so I figured I'll send this PR throough to add it to the docs.

I added as much detail as I have from glancing over the code and from using the events, happy to make changes in terms of tone/level of detail if needed.

### DIFF
--- a/artisan.md
+++ b/artisan.md
@@ -18,6 +18,7 @@
 - [Registering Commands](#registering-commands)
 - [Programmatically Executing Commands](#programmatically-executing-commands)
     - [Calling Commands From Other Commands](#calling-commands-from-other-commands)
+- [Events](#events)
 - [Stub Customization](#stub-customization)
 
 <a name="introduction"></a>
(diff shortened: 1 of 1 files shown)


taylorotwell:
Why does it look like you removed unrelated code?

codegefluester:
@taylorotwell no idea to be honest, All I did was adding the text between the last two sections, maybe that makes it look like content has been removed? The diff says I only added 6 lines and removed 0

codegefluester:
@taylorotwell Just had another look with the "Rich Diff View" and it shows only additions and no removals. Anything you want me to check/do to get this merged? It's my first time submitting a PR (for Laravel anyways) so I'm not quite certain what the next steps are right now.

https://github.com/laravel/docs/pull/6432/files?short_path=7af53ca#diff-7af53ca4def61eb29ea73de8ff108cb4

codegefluester:
Friendly ping, @taylorotwell

codegefluester:
I did some testing using different editors (just in the off chance that it is caused by this) and also tried just adding the bit about Events to the end of the page rather than between the last two sections and the result is always the same. 

The lines appear red and green in the diff view. The only place where it shows up correctly is in the rich diff view (see https://github.com/laravel/docs/pull/6432/files?short_path=7af53ca#diff-7af53ca4def61eb29ea73de8ff108cb4).

@taylorotwell if you want we can either re-open this pull request to get it merged, I submit a new one (incl. the changes you made) or we just leave it.

Screenshot 2020-10-01 at 17 29 24

PR #6440, opened 2020-10-01
fix broken image

driade:


driesvints:
Thanks!

driade:
FYI Now https://github.com/driade/laravel-book with be able to compile these changes, that image broke the workflow!

PR #6447, opened 2020-10-02
[8.x] Removed -repo GitHub search filter from the  Contribution Guide -> Bug Reports section

codegefluester:
Looks like this filter is unnecessary


taylorotwell:
I think it is.

PR #6451, opened 2020-10-03
[8.x] Add warning to `migrate:fresh`

tontonsb:
For one reason or another you might have tables in the database that you don't want to be managed by Laravel migrations even when developing. In those cases `migrate:fresh` should not be used.

Adding a warning  as suggested in [this comment](https://github.com/laravel/framework/issues/27851#issuecomment-693351362).

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -154,6 +154,8 @@ The `migrate:fresh` command will drop all tables from the database and then exec
 
     php artisan migrate:fresh --seed
 
+> {note} The `migrate:fresh` command does not respect the database prefix. Consider this carefully when developing on a database which is shared by other applications.
+
 <a name="tables"></a>
 ## Tables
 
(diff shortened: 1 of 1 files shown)


ahinkle:
Would argue that this command should only be used in dev/local. Not sure why you would want to use the prefix (for multiple apps) unless in production.

tontonsb:
> Would argue that this command should only be used in dev/local. Not sure why you would want to use the prefix (for multiple apps) unless in production.

There are two cases (that I've both come across a lot and first is my everyday reality) that requires using a prefix:

- Using (or having to use) a single database for multiple (all) projects on the development/staging server.
- Developing an addon app that's sharing a database of another app, for example phpBB or wordpress or moodle or whatever. In that case you don't want to redo the  import of that database on every refresh when you simply need to change a field on a table that you're adding with the migrations.

PR #6456, opened 2020-10-05
Fix very small typo

enderandpeter:
* The word "script" in this sentence is out of place because the sentence is referring to `bootstrap/app.php` as a distinct thing,
     but it would be alright to say "the `bootstrap/app.php` script".
  * Removing the word "script" seems like the better choice because now it matches the phrasing of the next paragraph that says,
    "let's just focus on the HTTP kernel, which is located in `app/Http/Kernel.php`"


driesvints:
Thanks!

PR #6459, opened 2020-10-05
captalize `Extension` word in a item in installation.md required extensions

parsampsh:
an item in installation required extensions list is not sync with anothers. this changed from 'extension' to 'Extension' to be sync with another items in list.


driesvints:
Seems like something went wrong with branching here

parsampsh:
oh yes i'm sorry, i will fix this and send again :)

PR #6464, opened 2020-10-06
[8.x] Removed the ' after class

royduin:
Don't know if this is a spelling thing or it was just a search replace issue in the past. But I don't see a `'` everywhere, for example I don't see it here: https://laravel.com/docs/8.x/blade (search for `class constructor`).

If this one gets merged I can create PR's for the other branches?


taylorotwell:
It's possessive.

PR #6467, opened 2020-10-06
Use the default traits of User in the example

treyssatvincent:
In 8.x the default User model have the `HasFactory` trait by default. 
It should be reflected in the example given by the documentation in the same way as the `Notifiable` one.


treyssatvincent:
The commit name may be inappropriate

PR #6470, opened 2020-10-07
Adding the correct return values for missing options

baminc:
If a option is missing for a command the return value is not null but false. This updates the documentation.


taylorotwell:
Think it depends on the option... I get `null`:

image

baminc:
A option as a bool flag returns false if not present. For example if --dry is missing $this->option('dry') returns false not null. Not sure thats reflected in the current documentation?

PR #6476, opened 2020-10-08
Remove unnecessary line of Telescope documentation

r2luna:


driesvints:
This isn't unnecessary. If you're only installing telescope locally you should add the TelescopeServiceProvider manually. Note the line above where it says to remove it from the config file.

PR #6477, opened 2020-10-08
Add version to SES config example code

hsalem7:
Given that the version is required in AWS SDK configuration, maybe it makes sense to have it in the config example code. That would save some time for the developer to check what key should add to the config array.

Ref:
https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/getting-started_migration.html


taylorotwell:
We already set this for you.

PR #6478, opened 2020-10-08
[8.x] More indefinite articles before vowel sounds

enderandpeter:
Found more as I read individual pages and by using the command: `grep -E '\W[aA] \W?[AEIOU].+' *.md`


driesvints:
@enderandpeter thanks for all these prs btw. Appreciated 👍

PR #6482, opened 2020-10-08
update code sample to match parent method

macbookandrew:


taylorotwell:
Service provider boot is not static.

macbookandrew:
Oh, I’m sorry…I misread the description above the code block and added this in my **model’s** `boot` method, not the app service provider.

BTW, thanks for all your work on the Laravel ecosystem—I love working with it!

PR #6492, opened 2020-10-13
Add note on using dropForeign method

paulocuambe:
Reading the [docs](https://laravel.com/docs/8.x/migrations#foreign-key-constraints) I was confused why after rolling back a migration using `dropForeign` method I had success but the field in the database was still there.

After experimenting a bit I noticed that we **must** always use `dropColumn` alongside this method to properly delete the foreign constraint and the column.

**Note:** as I don't have a good English vocabulary I would appreciate if you would review this address any points related to this PR.


taylorotwell:
The docs already state is drops the foreign key.

PR #6504, opened 2020-10-17
[8.x] Fix php artisan command

mrozbarry:
`schedule:work` does not exist, but `schedule:run` does and seems to be what was intended.

### DIFF
--- a/scheduling.md
+++ b/scheduling.md
@@ -31,9 +31,9 @@ This Cron will call the Laravel command scheduler every minute. When the `schedu
 
 ### Starting The Scheduler Locally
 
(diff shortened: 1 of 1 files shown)


browner12:
https://github.com/laravel/framework/blob/8.x/src/Illuminate/Console/Scheduling/ScheduleWorkCommand.php

mrozbarry:
Oh, that's pretty recent. Should the documentation mention both, since 8.7.1 does not seem to have `schedule:work`?

browner12:
I believe the assumption in the docs is we always assume the latest release of a major version, since you can upgrade to it with no breaking changes.

mrozbarry:
Okay, I can accept that. Thanks for the quick feedback!

PR #6505, opened 2020-10-19
[8.x] Document RateLimited and WithoutOverlapping Job Middleware

paras-malhotra:
Document the 2 new job middleware: `RateLimited` (https://github.com/laravel/framework/pull/34829) and `WithoutOverlapping` (https://github.com/laravel/framework/pull/34794).

There was a section on rate limiting using `Redis::throttle` and `Redis::funnel`. This section is removed because:
1. The `RateLimitedWithRedis` class does the same thing as `Redis::throttle`. The `Redis::throttle` is also already documented in the job middleware section. 
2. The `Redis::funnel` method is now moved to job middleware as an example to limit concurrency.

This way, all 3 types of middleware (rate limiting, preventing overlapping and concurrency limiting) are together documented in the job middleware section.

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -6,14 +6,16 @@
 - [Creating Jobs](#creating-jobs)
     - [Generating Job Classes](#generating-job-classes)
     - [Class Structure](#class-structure)
-    - [Job Middleware](#job-middleware)
+- [Job Middleware](#job-middleware)
+    - [Rate Limiting](#rate-limiting)
+    - [Preventing Job Overlaps](#preventing-job-overlaps)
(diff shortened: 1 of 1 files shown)


sansebi:
**WithoutOverlapping** works like a charm! 👍 
I run into a dead end with my `CACHE_DRIVER` being set to `file`

> Call to undefined method Illuminate\Cache\FileStore::lock()

Would be nice to extend the docs that it only works for specific cache drivers.

driesvints:
@sansebi can you send in a PR for that?

paras-malhotra:
Thanks @sansebi, I sent a PR to document this: https://github.com/laravel/docs/pull/6520

PR #6506, opened 2020-10-20
[8.x] Improve linkability of docs by adding named anchors to h4 tags

jessarcher:
Some h4 tags already had named anchors, but the majority didn't. This made it hard to link to specific subsections of the documentation in blog posts, peer reviews, chats, etc. I have often found myself linking to a h3 tag and then telling the recipient to "scroll until you reach X". This PR aims to fix that by:

* Adding named anchors to the h4 tags that were missing them.
* Adding named anchors to the handful of h3 tags that were also missing them.
* Fixing an existing duplicate anchor name.

This PR does *not* change the TOC or visual appearance. The only UX change is that the h4 tags are now focusable and have a pointer cursor as they are links, but they do not have the red "#" that the other headings have (I can PR this to the website repo if that is a welcome addition though).

I have also only done this for the 8.x branch.

**The process**

I mostly automated this by writing a script that searched for lines beginning with `#### ` that weren't already preceded by a named anchor, and then automatically prepended a named anchor with a slugified version of the heading text. Thanks `Str::slug()`!

This created a few duplicate anchor names within the same page, so I wrote another script to find any duplicates within the same file so that I could manually add some extra context to uniquely distinguish them. In doing so I found one preexisting duplicate that I also fixed.

It's not a perfect process, but given there ended up being 674 new anchors, it was too much to do by hand. I have had a quick glance over every added anchor and they all seem good to me.


timacdonald:
I have wanted this sooooooo many times. Thank you 🙌

helpers.md
```diff
@@ -128,7 +131,7 @@ Laravel includes a variety of global "helper" PHP functions. Many of these funct
 
 </div>
 
-<a name="fluent-strings"></a>
+<a name="fluent-strings-method-list"></a>
```
timacdonald, line 134: Are we worried that this will break existing links to this header in existing blog posts and what not

artisan.md
```diff
@@ -497,6 +509,7 @@ If your command defines an option that accepts an array, you may pass an array o
         ]);
     });
 
+<a name="passing-boolean-values"></a>
 #### Passing Boolean Values
 
 If you need to specify the value of an option that does not accept string values, such as the `--force` flag on the `migrate:refresh` command, you should pass `true` or `false`:
```
timacdonald, line 515: Is this just GitHub doing some weird highlighting or something? The text is highlighted like it's being deleted or something, but it isn't showing as lines deleted in the sidebar

timacdonald: Screenshot in case it is just me

Screen Shot 2020-10-20 at 4 10 38 pm

timacdonald: 🤷‍♂️ The rich-diff looks normal. Doesn't show any content missing. Maybe just something I've not seen on GitHub before. Ignore me.

helpers.md
jessarcher: This was an existing duplicate anchor name, so it will have already been broken to some extent, depending on which of the two "Fluent Strings" headings they were attempting to link to.

One was a h2 and one was a h3, so I changed the h3, because the h2 outranked it (:rofl:). However, this h3 comes first in the document, and it seems like that one "wins" when there are duplicates on the page, at least with Firefox. So I may have been better off changing the h2, although to what I'm not sure.

The TOC on this page is also missing a lot of the h3 links and could probably do with a bit of attention.

artisan.md
jessarcher: Yeah I noticed that as well. Seems to be a syntax highlighting glitch not directly related to this change. Perhaps there is some invalid Markdown elsewhere in the file throwing it off though...

driesvints: Yeah there's a long outstanding problem with Github breaking over some diffs. Feel free to ignore.

driesvints (approved):
This lgtm. Thanks Jess!

ashbakernz (approved):
This is definitely going to help me link sections to other team mates! Thanks Jess

taylorotwell:
Nice work!

HiroKws:
To be honest, I wonder if these are manually maintained. It makes more sense to leave it to the converter that converts markdown to HTML, or generate them on load with JavaScript.

PR #6509, opened 2020-10-21
Update queues.md

ronnievisser:
Instantiate the WithoutOverlapping class so that methods can be called on it.


driesvints:
Thanks

PR #6510, opened 2020-10-21
Order of parameters is swapped

bilogic:
https://github.com/laravel/framework/blob/9c047344505da509c2b893eed7aac369a73e07eb/src/Illuminate/View/Compilers/BladeCompiler.php#L546


taylorotwell:
Classes can be passed in either order.

PR #6514, opened 2020-10-22
[8.x] Added a semicolon after assertJsonPath method

tatsuroooK:
Added a semicolon after `assertJsonPath()` method because seems necessary.


driesvints:
Thanks!

PR #6521, opened 2020-10-24
[6.x] Add information on Dusk usage with Homestead

lupinitylabs:
This PR proposes an additional section with information on how to install Dusk on Laravel Homestead. 

Neither the Dusk nor the Homestead documentation explicitly states that it is now needed to include the webdriver feature in order to use Dusk. 

This info would have saved me a couple of hours of debugging and research.
It would probably be useful to also specify the version at which Homestead dropped Dusk support, which I could not determine.

### DIFF
--- a/dusk.md
+++ b/dusk.md
@@ -2,6 +2,7 @@
 
 - [Introduction](#introduction)
 - [Installation](#installation)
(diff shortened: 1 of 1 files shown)


taylorotwell:
I've made these changes on the 8.x branch.

PR #6525, opened 2020-10-26
[8.x] Note for limitation of 72 characters for bcrypt driver

laundy:
Add a note that for the bcrypt driver no plain texts longer than 72 characters can be hashed. PHP's `password_hash()` and Laravel won't output a warning in this case but several plain texts longer than 72 characters will then yield the same hash and produce collisions.

### DIFF
--- a/hashing.md
+++ b/hashing.md
@@ -58,6 +58,8 @@ If you are using the Bcrypt algorithm, the `make` method allows you to manage th
         'rounds' => 12,
     ]);
 
+> {note} As Laravel `Hash` facade utilizes PHP's `password_hash` function, in combination with the `bcrypt` driver only the first 72 characters of the plain text are hashed. For more information, check out the [official PHP documentation](https://www.php.net/manual/en/function.password-hash.php).
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
Unclear what the take-away from this sentence is supposed to be.

laundy:
Takeaway is that you shouldn't use `Hash` facade with `bcrypt` for other things than passwords (and in my opinion even not for those). Example: I used `Hash::make()` with default configuration for getting a checksum for a [Microsoft Graph API](https://docs.microsoft.com/en-us/graph/overview) ID will create a lot of "collisions" as it is longer than 72 characters. Explained a possible use case in detail on [StackOverflow](https://stackoverflow.com/questions/64526355/laravel-hashcheck-returns-true-for-two-different-messages).

This special behavior is not obvious from the current documentation as it looks like a "multi-purpose hash facade". Even the [PHP documentation](https://www.php.net/manual/en/function.password-hash.php) contains a warning for this. Hope this make things clearer.

PR #6534, opened 2020-10-29
[8.x] Added snippet on how to generate UUID for existing failed jobs

lasselehtinen:
When adding the UUID during upgrade, existing failed jobs will be left as null and will cause issues. So I added a small snippet you can run in Tinker to generate UUID for existing rows.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -282,10 +282,12 @@ If you plan to use the [job batching](/docs/{{version}}/queues#job-batching) fea
 
 Next, the `failed.driver` configuration option within your `queue` configuration file should be updated to `database-uuids`.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
What issues will it cause if you don't do this?

lasselehtinen:
You cannot retry individual jobs since UUID is now the key. artisan queue:failed returns the ID column as empty. Haven't tested, but `queue:retry all` probably will also fail, since the id would be empty?

https://github.com/laravel/framework/blob/43bea00fd27c76c01fd009e46725a54885f4d2a5/src/Illuminate/Queue/Console/RetryCommand.php#L58

PR #6545, opened 2020-11-05
[8.x] Clarify route group name prefix when array argument

edh649:
Using 'name' as the array argument for the route group name prefix parameters does not work, 'as' does.


taylorotwell:
We don't document array usage anymore. Only method usage.

edh649:
Okay. Shall submit another pull request to remove

> Shared attributes are specified in an array format as the first parameter to the Route::group method.

At the top of the route group section

PR #6549, opened 2020-11-06
[8.x] Make it clearer that `union()` can be used to join 2 or more queries

u01jmg3:
- Make it clearer that `union()` can be used to join 2 or more queries 🤝


taylorotwell:
I have updated the wording without changing the example.

PR #6553, opened 2020-11-06
Add information about several database connections to database testing section

AntonKucheriavyi:
I ran into the problem when I was testing code with multiple database connections and couldn't find it in the documentation. It would be nice to add this to save people time


driesvints:
This should be sent to 8.x and tbh I don't think this is necessary.

PR #6554, opened 2020-11-08
[8.x] Add information about several database connections to database testing section

AntonKucheriavyi:
I ran into the problem when I was testing code with multiple database connections and couldn't find it in the documentation. It would be nice to add this to save people time


taylorotwell:
This doesn't feel complete.

AntonKucheriavyi:
What else can I add? I'll try

PR #6557, opened 2020-11-09
[8.x] Fix missing Token Class

muhammadcahya:
Fix missing Token Class usage on overriding section, use `PersonalAccessToken` directly when use custom `PersonalAccessToken` model on boot


driesvints:
Thanks!

PR #6562, opened 2020-11-10
[8.x] Add undocumented api change to Paginator

iBotPeaches:
All length paginated endpoints were failing tests because of the meta property change for a project. Probably most don't do that level of testing, so marked it as Low.

Tracked back to this [commit](https://github.com/laravel/framework/commit/13751a187834fabe515c14fb3ac1dc008fd23f37) - looks like it was probably for one of the new frontend components.


taylorotwell:
Since there is really no actionable upgrade path for the developer here I'm not sure it should be included.

PR #6564, opened 2020-11-11
[8.x] Adds section for migrations setup, status tip

emjayess:
Adds a Migration Setup doc section indicating the use of 'migrate:install' for setup, and a tip for 'migrate:status' usage

---

There is no mention of either `migrate:install` nor `migrate:status` in the database migrations documentation page, and the absence of documentation for these commands may trip up newcomers or anyone unaware of this necessary first step.

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -1,6 +1,7 @@
 # Database: Migrations
 
 - [Introduction](#introduction)
+- [Migrations Setup and Status](#migrations-repository)
 - [Generating Migrations](#generating-migrations)
     - [Squashing Migrations](#squashing-migrations)
 - [Migration Structure](#migration-structure)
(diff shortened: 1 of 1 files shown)


driesvints:
This isn't necessary as the framework will install this table automatically when the first migrations are run.

emjayess:
> _This isn't necessary as the framework will install this table automatically when the first migrations are run._

Well, perhaps the docs could minimally mentioned as much?

---

The point stands that the documentation is lacking a more thoughtful and complete explanation of migration internals, and available commands (or artisan migrate api surface area), which can be vexing for people desiring to understand more quickly how it works.

And I have personally observed this trip up contributors in the process of trying to ramp up on existing projects.

driesvints:
We can't possible explain all internals in user facing docs. That would be overwhelming. If you feel anything else's missing feel free to attempt a new PR.

PR #6585, opened 2020-11-23
[8.x] Show how to define queues directly on the Job

browner12:
maybe I'm the odd duck here, but I like to define queues directly on my Job, rather than in my calling code, since I want it to default to the same queue every time.


taylorotwell:
I've pulled in these changes to a local branch where I have a lot of changes going on.

PR #6587, opened 2020-11-24
[8.x] Document required headers for Sanctum token authorization

crzzano:
No mention of this in the docs but it is required for the auth:sanctum guard to work

### DIFF
--- a/sanctum.md
+++ b/sanctum.md
@@ -160,6 +160,8 @@ To protect routes so that all incoming requests must be authenticated, you shoul
         return $request->user();
     });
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
We would need more details than this. When calling which route? Under what circumstances? What error do you get if you don't?

crzzano:
It's for all API routes that use auth:sanctum. Without these headers the system redirects you to the homepage (or wherever your guest middleware takes you). As this is in the API Token Authentication section of the Sanctum docs I think it is important to mention which headers are required for the sanctum auth middleware to function as expected.

PR #6588, opened 2020-11-24
[8.x] add a mention that you don't need separate queue worker with horizon

smknstd:
I made the mistake running both. I found out about it by reading @driesvints [blog post](https://medium.com/@driesvints/laravel-horizon-with-forge-and-envoyer-82a7e819d69f) and noticed it's not explained anywhere else.

> Don’t forget that you don’t need to define queues through Forge as Horizon completely takes this over.


taylorotwell:
This doesn't fit in with the flow of the docs. I'll try to clarify this.

PR #6597, opened 2020-11-27
[8.x] Update homebrew default php version

MammutAlex:
Now homebrew install [php 8 by default](https://formulae.brew.sh/formula/php)


driesvints:
Is Homebrew already completely up to date? With extensions like Memcached and Redis?

PR #6598, opened 2020-11-27
[8.x] Update cache.md

raftx24:
This PR addresses a potential problem when using as default a custom cache driver.
By registering drivers in the boot method of an ApplicationServiceProvider as currently recommended in the docs, the driver may not available to other services that need retrieving values from cache, services that run in boot methods from other service providers. This can happen due to load order of the providers.

### DIFF
--- a/cache.md
+++ b/cache.md
@@ -397,7 +397,7 @@ We just need to implement each of these methods using a MongoDB connection. For 
 <a name="registering-the-driver"></a>
 ### Registering The Driver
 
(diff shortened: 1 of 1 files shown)


driesvints:
FYI, @taylorotwell, this is related to https://github.com/laravel/telescope/issues/988. I'm not sure that this is the best place to document it but I agree with @raftx24 & @aocneanu that we should update the docs here a little.

taylorotwell:
I've made these changes on my local branch that I'm about to send up to laravel/docs next week. Thanks.

PR #6600, opened 2020-11-27
[8.x] Minor update on database docs page

tuathail:
If the DB configurations are set at runtime with the Config facade / config helper then the following statement isn't strictly correct: 
"The name passed to the connection method should correspond to one of the connections listed in your config/database.php configuration file".
ie it could happen that the connection name isn't in the config/database.php file and everything could still work, while the above line suggests that this isn't possible!


taylorotwell:
Made this change on a local branch I'm working on.

PR #6616, opened 2020-12-06
[8.x] Update CSRF documentation to show expected field name

weaversam8:
Nowhere in the documentation is `_token` listed as the correct field name for the CSRF token hidden field. This illustrates that with an example, also helping people understand what exactly the `@csrf` macro does.


taylorotwell:
This is already fixed in my documentation updates shipping tomorrow.

PR #6618, opened 2020-12-06
[8.x] Add note about URL generation for controller actions

thinkverse:
This section was a bit confusing so added a note that routes need to be registered for the `action()` function to generate a URL for a given controller action.


taylorotwell:
Hmm, I would think that would be the obvious expectation but maybe I'm missing something? It feels redundant to say it here.

PR #6619, opened 2020-12-07
[8.x] Validation Error

jsaberon:
Since Laravel 8 is using Tailwindcss as a default css. I change the display validation error example to tailwind classes because it is still in bootstrap


taylorotwell:
I think it's fine as it is. It's just supposed to be for example purposes and not necessarily any particular CSS framework.

PR #6620, opened 2020-12-07
[8.x] Reference to Pivot Tables

MimisK13:
Reading the docs, I did not find references and instructions about pivot tables. And that's not enough, but it's a start for new Laravel users.


taylorotwell:
The Eloquent relationship documentation clearly defines the table structure that should be in pivot tables?

PR #6630, opened 2020-12-08
Clarify required software for Homestead

LukeTowers:
Minor change to clarify the software required for Homestead to function. The existing text is somewhat ambiguous and can lead to thinking that all of those provided options are required to be installed when in reality only one of them needs to be. Thanks to @petro-lazurko for bringing this to my attention.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -129,7 +129,7 @@ Homestead runs on any Windows, macOS, or Linux system and includes Nginx, PHP, M
 <a name="first-steps"></a>
 ### First Steps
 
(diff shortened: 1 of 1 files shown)


browner12:
I kind of like the original wording better.  AFAIK, these are the only 3 Virtual machines we build boxes for, so saying "such as" implies that it would work on others as well. You explicitly need one of these 3 options.

Also, you either added an extra "(" or forgot to add a closing one.

LukeTowers:
@browner12 how about
```md
Before launching your Homestead environment, you must install [Vagrant](https://www.vagrantup.com/downloads.html) as well as one of these supported virtualizers:
- [VirtualBox 6.x](https://www.virtualbox.org/wiki/Downloads)
- [VMWare](https://www.vmware.com)
- [Parallels](https://www.parallels.com/products/desktop/)
- [Hyper-V](https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/quick-start/enable-hyper-v)
```

browner12:
I'd be more onboard with this list layout. Obviously Taylor has final say.

I agree, it could be hard to catch that "or" buried in the sentence.

LukeTowers:
Sounds good, I'll resubmit the PR then.

PR #6639, opened 2020-12-09
[8.x] Update blade.md

richardkeep:
The shorthand function statement requires PHP 7.4 but Laravel 8 still supports PHP 7.3. Whoever will copy this code snippet will need to refactor on PHP 7.3.


taylorotwell:
I think it's OK.

PR #6648, opened 2020-12-10
[8.x] Update migrations.md

nbayramberdiyev:
Documentation for [`foreignIdFor`](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Schema/Blueprint.php#L856) method.


taylorotwell:
Not sure I really recommend using this method because model names and locations can change.

PR #6650, opened 2020-12-10
Fix composer command

elvendor:


driesvints:
Another PR already fixed this. Thanks

PR #6659, opened 2020-12-12
[8.x] Improve Example ModelObserver Registration

jlaswell:
This adds `parent::boot()` call to example of registering a model observer within the EventServiceProvider. I know this is a somewhat obvious element to registering things in the `boot()` method, but it caught me today and want to pitch this clarity as a suggestion.

Happy to submit this change to `7.x` branch as well if it's a +1.

cc: @roberto-aguilar

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -1216,6 +1216,8 @@ To register an observer, you need to call the `observe` method on the model you 
      */
     public function boot()
     {
+        parent::boot();
+        
         User::observe(UserObserver::class);
     }
 


crynobone:
Why not use the new `booted()` method instead.

jlaswell:
This is to change the example within the docs related to the EventServiceProvider rather than provide a different suggestion to how to implement the `observe` call.

That seems like a valid enough implementation change, but the existing example works well enough with the suggested change.

crynobone:
> This is to change the example within the docs related to the EventServiceProvider rather than to provide a different suggestion on how to implement the observe call.

:+1:

taylorotwell:
It is no longer necessary to call parent::boot() in the `EventServiceProvider` in recent releases of Laravel.

PR #6660, opened 2020-12-12
Added section for customising Sail from the .env file.

dbilovd:
Some sail configurations like port numbers, DB credentials, etc. can be updated directly from the .env file.


taylorotwell:
I'll get this added.

dbilovd:
Okay

PR #6671, opened 2020-12-16
[8.x] Document broker method

nathanpurcell:
I couldn't find any reference to the broker method anywhere in the 8.x documentation. Goes hand in hand with the guard method on the Auth facade when adding multiple entry points so thought it was important.


taylorotwell:
This will need to be more fleshed out. I'll try to document it.

nathanpurcell:
Related, would you be interested in a PR to Breeze that allows the user to scaffold auth for specific guards? As it stands all of Breeze appears to be for the default guard. It’s simple enough to extend but I was thinking of something along the lines of: 

`php artisan breeze:install —guard=members`

PR #6673, opened 2020-12-17
[8.x] Compare v7 to v8 branch

wjonkerhulst:
We need to compare te changes from v7 to the v8 branch when upgrading.


driesvints:
Thanks!

PR #6676, opened 2020-12-17
Update sanctum.md

lucianotonet:
I think that was wrong


taylorotwell:
Doesn't look wrong to me. Creating something would normally be a POST request.

PR #6683, opened 2020-12-22
[8.x] Timezone configuration.

mellester:
Not all Databases are out off the box UTC. There is a undocumnted setting for myslq and postgress that should help streamline this

### DIFF
--- a/database.md
+++ b/database.md
@@ -315,3 +315,15 @@ If you would like to connect to your database's CLI, you may use the `db` Artisa
 If needed, you may specify a database connection name to connect to a database connection that is not the default connection:
 
     php artisan db mysql
+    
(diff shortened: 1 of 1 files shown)


driesvints:
You should store your timestamps as UTC and then do the conversion to whatever you want in userland.

mellester:
> You should store your timestamps as UTC and then do the conversion to whatever you want in userland.

Yeah you should. This setting is the preferred way to me to ensuring that happens.  Some people like to to use and send 'DB::raw(NOW())'  this setting ensures that said timestamp is utc otherwise its might use the database timezone

taylorotwell:
Some grammar problems that don't fit the rest of the Laravel documentation.

PR #6687, opened 2020-12-24
Fix typo artisan call silent

PanjiNamjaElf:
In docs said when want suppress all of artisan output may use `callSilently` method
But when i see it on laravel 8.x the method name is `callSilent` not call `callSilently`


taylorotwell:
I see both methods.

PR #6689, opened 2020-12-25
Mention that dynamodb is supported

stancl:


taylorotwell:
I don't think I would lump Dynamo in with Redis or Memcached. I think I will list it on its own line.

PR #6698, opened 2020-12-29
Remove mail sender name

MarcelWeidum:
This does not work see: https://stackoverflow.com/questions/41972333/how-to-change-from-name-in-laravel-mail-notification


taylorotwell:
image


Looks like it uses these keys to me?

PR #6699, opened 2020-12-29
Redundant Outline List

EricMcWinNer:
The additional line sending us to the Inverse is redundant and misleading, I started thinking there is a brand new relationship I didn't know of


EricMcWinNer:
Hi @taylorotwell, could you please let me know why the issue was closed? I read through the docs and the third link leads to the inverse of One to Many which isn't a relationship of itself? Please let me know if I'm misunderstanding anything. Like why didn't we have a similar link for the inverse of a One to One relationship?

PR #6700, opened 2020-12-30
Removed TelescopeServiceProvider without qualified namespace

moazam1:


bonzai:
These are not the same service providers.

The one you want to remove extends `Laravel\Telescope\TelescopeApplicationServiceProvider`.

paras-malhotra:
Neither should be removed. Both are needed for Telescope to work.

paras-malhotra:
`Laravel\Telescope\TelescopeServiceProvider::class` needs to be removed as it does not require manual registration (is auto discovered in local).

driesvints:
This is incorrect. It needs to be explicitly added in the service provider because these are instructions for a local only installation. With a local only installation the server provider isn't auto loaded because it shouldn't be available in production.

paras-malhotra:
@driesvints with a local installation, the service provider is auto-loaded imo. Remember there are no instructions to add telescope to the `dont-discover` list and the Telescope package does register auto-loading as below:

https://github.com/laravel/telescope/blob/535889a7585c331854dbdf8ca5e66209588408dd/composer.json#L37-L43

PR #6702, opened 2020-12-30
Typo and grammar

Netflyerr:


Netflyerr:
I think the "if" can be omitted, though.

PR #6708, opened 2020-12-30
Windows Terminal/PowerShell Alias

nasirkhan:
On the installation doc it is mentioned to use Windows Terminal on windows, Bash and Power Shell Alias is different. While following the doc from a windows machine I found this should be included.


taylorotwell:
You should be within WSL.

PR #6709, opened 2020-12-31
Remove duplicate text in blade.md

simasbut:
The section on Manually Registering Components was duplicated. This PR deletes the duplicate which did not correspond to Table of Contents. And leaves the same text at line 1070 untouched.


taylorotwell:
Intentional.

PR #6714, opened 2021-01-04
[8.x] Change User to UserResource

zuzana-kunckova:
I would like to propose to change `User` in the Generating Resources section example to `UserResource` to avoid conflict with the `User` model namespace when importing the `User` resource in the `UserController`.

### DIFF
--- a/eloquent-resources.md
+++ b/eloquent-resources.md
@@ -166,7 +166,7 @@ When the `preserveKeys` property is set to `true`, collection keys will be prese
 <a name="customizing-the-underlying-resource-class"></a>
 #### Customizing The Underlying Resource Class
 
-Typically, the `$this->collection` property of a resource collection is automatically populated with the result of mapping each item of the collection to its singular resource class. The singular resource class is assumed to be the collection's class name without the trailing `Collection` portion of the class name.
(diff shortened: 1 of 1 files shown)


taylorotwell:
If we do this there may be a few other spots in this piece of documentation that need to be updated. Are you able to check for those? Thanks.

zuzana-kunckova:
Yes, I'll go over the whole section and update the PR.

taylorotwell:
While reading this over this section gave me a bit of pause:

image

I think we will need to make a slight change to the framework if we make this documentation change since people will use the `Resource` suffix when generating resources which means the default resource collections will no longer be able to find the resource they map to.

I think I'll need to update this method in the framework to also look for FooResource:

image

driesvints:
Isn't the more simpler solution to alias the User resource import in the controller?

zuzana-kunckova:
I didn't realise it would mean making more extensive changes, sorry, Taylor.

If you don't make the change and continue to alias the User resource import in the controller, it might be good to mention in the docs that this is what we are doing and why. I didn't notice the alias when I first started working with API resources. Also, when I started creating resources by giving them more descriptive names, e.g. `UserResource` instead of just `User`, it wasn't clear whether that wouldn't break something somewhere else, whether the resource HAD to be same as the model name.

taylorotwell:
Personally I think it would be nice to add the $collects property to the resource collection stub class and make that connection a bit more explicit and de-emphasize this magical convention. I may look into that today.

zuzana-kunckova:
Awesome, thank you, Taylor!

PR #6718, opened 2021-01-06
Update upgrade.md

georanma:
Update to model factories changes to advise about `HasFactory` trait. Also includes a note for custom Model structures with respect to the namespace updates. 
Not 100% on the location or wording, so open to suggestions.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -139,6 +139,25 @@ Laravel's [model factories](/docs/{{version}}/database-testing#creating-factorie
 
     composer require laravel/legacy-factories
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't think this guide should cover how to transition your factories. It deals with keeping your existing code compatible.

georanma:
@taylorotwell maybe it feels more relevant to mention this detail with the factories and seeders namespace change then? I was following that part of the document, along with the additional details at the actual Factories documentation, but found not mention of the trait.

PR #6720, opened 2021-01-06
Explain Sail Bash alias further

ccchapman:
This change provides a little more detail into how a user may configure their Bash alias.

### DIFF
--- a/sail.md
+++ b/sail.md
@@ -59,10 +59,16 @@ By default, Sail commands are invoked using the `vendor/bin/sail` script that is
 ./vendor/bin/sail up
 ```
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
It does not have to go in that file and it depends on your operating system.

ccchapman:
It is not operating system specific, this file is outlined in `man bash`. While it is not the only way, it is written as a suggestion, it is the most common place a Bash user will place this file. For someone that is reading this that has never configured an alias, it is useful to provide a little more detail. More advance users that maintain dotfiles already or use a different shell (e.g. fish, zsh) will know where they may like to place it.

PR #6726, opened 2021-01-08
[8.x] Fix mobile horizontal scroll on Sanctum page

TimothePearce:
## Issue
The current Sanctum page has a horizontal scroll on mobile devices as you can see on the screenshot below:

horizontal-scroll

## Fix
I replaced the `<code>` tag with a `<pre>` one to avoid the overflow.

### DIFF
--- a/sanctum.md
+++ b/sanctum.md
@@ -75,7 +75,9 @@ Next, if you plan to utilize Sanctum to authenticate an SPA, you should add Sanc
 <a name="migration-customization"></a>
 #### Migration Customization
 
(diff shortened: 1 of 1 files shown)


driesvints:
I'm thinking we should rather fix the breakage itself rather than adjusting the content. Although this is a good intermediary fix.

TimothePearce:
Thanks, Dries! You're right, I'll make a PR in the [website repo](https://github.com/laravel/laravel.com-next).

TimothePearce:
@driesvints Here is the fix: https://github.com/laravel/laravel.com-next/pull/153

PR #6743, opened 2021-01-14
[8.x] Update the docs to reflect the merge of reduceWithKeys into the existing reduce method

mokhosh:
This updates Collection docs to reflect latest changes in `reduce`

As discussed here: https://github.com/laravel/framework/pull/35839
The new pull request to framework here: https://github.com/laravel/framework/pull/35878
Previous change to the docs that was merged and needs to be updated because the underlying changed was reverted here: https://github.com/laravel/docs/pull/6740

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -164,7 +164,6 @@ For the majority of the remaining collection documentation, we'll discuss each m
 [put](#method-put)
 [random](#method-random)
 [reduce](#method-reduce)
-[reduceWithKeys](#method-reducewithkeys)
 [reject](#method-reject)
 [replace](#method-replace)
 [replaceRecursive](#method-replacerecursive)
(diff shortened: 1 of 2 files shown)


collections.md
```diff
@@ -1725,7 +1721,7 @@ The `reduceWithKeys` method reduces an associative collection to a single value,
 
     $collection->reduceWithKeys(function ($carry, $value, $key) use ($ratio) {
         return $carry + ($value * $ratio[$key]);
-    }, 0);
+    });
```
mokhosh, line 1724: I previously added 0 as `$initialValue` to the example to make it clear that `reduceWithKeys` also accepts initial value. But now that it's merged into `reduce` and we already show that in the previous example there is no need for that and we can remove it to simplify and make it more readable.

Hornet-Wing:
I got confused by this one, should this one not be something like this?

"The `reduceWithKeys` method also passes array keys in associative collections to the given callback"

PR #6747, opened 2021-01-16
updated docs to reflect correct command to run schedule locally

pavan168:
In Running The Scheduler Locally

the command should be queue:work not queue:work


taylorotwell:
The current documentation is correct.

PR #6752, opened 2021-01-19
Missing `return` in an example function

Nyleor:


taylorotwell:
Not actually necessary but will add.

PR #6757, opened 2021-01-21
[8.x] Note about Horizon and Telescope UI code

driesvints:
Add a note that the UI from Horizon and Telescope does not fall under the public API and isn't part of Laravel's SemVer policy.


GrahamCampbell (approved):

taylorotwell:
I'm honestly not sure we need a special note about this.

PR #6760, opened 2021-01-21
[8.x] Improve dynamic relationship example

MarceauKa:
Add an example to the dynamic relationship using an anonymous PHP class.

In real case usage, this feature was useful when migrating a payment method provided by a third-party package where models and relations were not available anymore.


taylorotwell:
I'm not sure we need this example in the documentation.

MarceauKa:
@taylorotwell I agree with you about the example. But I think it can be useful to note that an anonymous class extending `Model` is possible. It saved me hours of work today.

PR #6761, opened 2021-01-21
[8.x] Add note about SqlServer and DBAL v3

driesvints:
The current implementation of DBAL v2 on Laravel 8 is semi-broken and I don't see a clean way to fix this properly. The DBAL v3 implementation should work however.


GrahamCampbell (approved):

driesvints:
Follow up note that this is only for sql server.

PR #6762, opened 2021-01-21
Added command for running a single test

SomePlayer:


taylorotwell:
This is more of a PHPUnit documentation thing. We note that any PHPUnit option can be passed already.

PR #6763, opened 2021-01-21
[8.x] Added informations about ->fallback() method on routes

wijourdil:
- This PR updates the documentation on fallback routes, adding informations on `->fallback()` method

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -520,6 +520,18 @@ Using the `Route::fallback` method, you may define a route that will be executed
 
 > {note} The fallback route should always be the last route registered by your application.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
We already show how to register with Route::fallback() so I'm not sure showing another way is needed.

wijourdil:
Yes but the two things are not exactly the same things, are they ?

The way it is written, we understand we can only define one route using the `Route::fallback()` method.

Maybe it can be useful to add at least one sentence to talk about the `->fallback()` method ?
I lost literally hours to do something because I didn't know this method, maybe it could help some people ?

PR #6771, opened 2021-01-24
Add some missing helpers

bakerkretzmar:
Adds entries on the Helpers page for `Str::parseCallback`, `Str::substrCount`, `encrypt`, `decrypt`, and `queueable`.


taylorotwell:
I don't want people to use encrypt / decrypt... I don't document those anymore intentionally. parseCallback is more for internal use.

PR #6772, opened 2021-01-25
Valet: Add instructions for using MailHog to intercept email from Valet sites

ChrisHardie:
This addition may be too opinionated for all of the scenarios that Valet supports. But think of the pain we'll save some poor developer who didn't think to check if their Valet-hosted WordPress site with production data was sending email to real users. (Friends, that poor developer was me today.)


taylorotwell:
I'm not sure I plan to include a section on this in these docs.

PR #6780, opened 2021-01-26
[8x] Added link to laravelversions.com

pascalbaljet:


browner12:
I think our current table is just fine.

PR #6781, opened 2021-01-26
[8.x] Add valet TLD command

stephenjude:
I found that that the valet command for changing the default TLD is not documented. I have updated the readme to reflect this command.

Command: `valet tld wip`


taylorotwell:
I don't really think it's worthwhile to change the TLD so I left it out. It can be discovered via artisan list valet.

stephenjude:
Okay

PR #6789, opened 2021-01-28
[8.x] The CreatesApplication trait

driesvints:
This adds a small section to the docs to make it clear that the original location of the `CreatesApplication` should not be changed as some functionality depends on it.

See https://github.com/laravel/framework/issues/36053


crynobone:
Also Laravel Dusk.

PR #6794, opened 2021-01-28
Correctly hyphenate links

u01jmg3:
- As per the rest of this file the method `Arr::sortRecursive` has the link `#method-array-sort-recursive` (and not `#method-array-sortrecursive`)


driesvints:
This will unfortunately break all existing links that are targeting these.

u01jmg3:
People who have previously bookmarked? Yes - the tidy up may therefore not be worth it 😿

PR #6796, opened 2021-01-29
[8.x] Making more sense for Laravel Rate Limit example code

thihakyaw:
Laravel configureRateLimiting() method didn't call in boot() method which makes the example code is not showing the example enough well.


taylorotwell:
I don't know why you added spaces to every code example.

PR #6802, opened 2021-01-31
Add note about domain approval and 404 error

marcomessa:
I had some issues with Paddle checkout, due to my test domain (both the local one with valet and the remote one). Then I found out that Paddle accept only localhost / 127.0.0.1, and all other domain must be approved. I know this is not strictly related to Cashier, but maybe this alert can help some other people.

### DIFF
--- a/cashier-paddle.md
+++ b/cashier-paddle.md
@@ -9,6 +9,7 @@
     - [API Keys](#api-keys)
     - [Paddle JS](#paddle-js)
     - [Currency Configuration](#currency-configuration)
+    - [Approve your domains](#approve-domains)
 - [Core Concepts](#core-concepts)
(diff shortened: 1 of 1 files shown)


driesvints:
I can use an unapproved domain to test locally just fine:

Screenshot 2021-02-01 at 10 10 53

marcomessa:
Hi @driesvints thank you for your answer. I also tried with expose but I've got a different behaviour:

expose
Schermata 2021-02-01 alle 10 14 48

artisan serve
Schermata 2021-02-01 alle 10 15 41

I think you're right and maybe the problem is elsewhere, I'll continue search and write here with new news. I apologize for the false allarm!

driesvints:
No problem 👍

PR #6803, opened 2021-01-31
[8.x] Update Validation.md to include sometimes section

Fijvect:
Currently the "Sometimes" rule sends users to a section that is not directly what they're expecting to see. This change adds the "sometimes" rule where it would be expected, while still referencing the "Validating when present" section for users to see examples and read further on.


taylorotwell:
I think it was fine how it was.

PR #6806, opened 2021-02-02
[8.x] Add note about uniqueId and uniqueFor for job batching

driesvints:
These aren't available when using job batching. See https://github.com/laravel/framework/issues/36118


taylorotwell:
I have added a note to the top of this section about batches in general.

PR #6808, opened 2021-02-02
[8.x] Updated Homestead for the v12.0.0 release

svpernova09:
- [X] There is no more `18.04`, there is only `22.04` 🚀 
- [X] 99.9999% of users should never `vagrant box add ...` 👎 Homestead manages this for users.
- [X] PostgreSQL is better than MySQL, at least we've added v13 now.
- [X] MySQL 8 is now default.
- [X] Testing out dropping support for Hyper-V and VMware providers ([no one has yelled yet](https://twitter.com/JoePFerguson/status/1356435248687554561))
- [X] PHP 8 is the current default.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -57,8 +57,7 @@ Homestead runs on any Windows, macOS, or Linux system and includes Nginx, PHP, M
 </style>
 
 <div id="software-list" markdown="1">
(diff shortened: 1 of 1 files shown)


taylorotwell:
Thanks I'll give this a review.

PR #6811, opened 2021-02-03
Update queries.md

m6u:
fix closure usage. (With reference to Anonymous functions "Example #3" https://www.php.net/manual/en/functions.anonymous.php )

### DIFF
--- a/queries.md
+++ b/queries.md
@@ -739,7 +739,7 @@ Sometimes you may want certain query clauses to apply to a query based on anothe
     $role = $request->input('role');
 
     $users = DB::table('users')
-                    ->when($role, function ($query, $role) {
+                    ->when($role, function ($query) use ($role) {
                         return $query->where('role_id', $role);
                     })
                     ->get();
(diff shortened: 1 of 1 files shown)


taylorotwell:
Laravel passes the first argument given to `when` to your callback.

kyoya0819:
I think that His Pull Request is right.
Please check His Pull Request again.

m6u:
First, I thought about the usage of when() in this way.
```
->when($city !== '', function($q, $city) {
    return $q->where("addr2", $city);
})
```
It works, but could not get the correct execution result.
Next, I wrote this.
```
->when($city !== '', function($q) use($city) {
    return $q->where("addr2", $city);
})
```
This worked and I got the correct execution results.
With [this point](https://github.com/laravel/docs/pull/6811#issuecomment-772508416), I reconsidered.
```
->when($city, function($q, $city) {
    if ($city !== '') {
        return $q->where("addr2", $city);
    }
    else {
        return $q;
    }
})
```
It good works, taylorotwell's point is correct.
By the way, if I want to pass more than one value to the closure, how do I write it?
Possible examples:
```
->when([$state, $city, $town, $section], function($q, $state, $city, $town, $section) {  // Maybe it won't work.
    if (...) {
        return $q->where(...);
    }
    else {
        return $q;
    }
})
```
or
```
->when(true, function($q) use ($state, $city, $town, $section) {  // It might work.
    if (...) {
        return $q->where(...);
    }
    else {
        return $q;
    }
})
```
Or something else.
I would like to include this kind of writing in the document as an example.

PR #6821, opened 2021-02-08
[8.x] Remove unused variable

hosseinmoghaddam:


driesvints:
This is to indicate the variable is passed down. It's useful in the docs even though it's not used.

hosseinmoghaddam:
you're right.

PR #6826, opened 2021-02-08
[8.x] Creating Payment method and Retrieving Default payment method as Stripe

yemenpoint:
Creating Payment method and Retrieving Default payment method as Stripe


driesvints:
This shouldn't be part of the Cashier docs since you're using purely a Stripe PHP SDK object. I also don't really know what you'd gain from documenting the `asStripePaymentMethod` call.

PR #6828, opened 2021-02-09
[6.x] Localisation namespace

tommym9:
Clocked an issue where a package uses a different namespace to the package name. In this case the namespace in the package service provider is used.

Added an addition to the docs for this edge case. Not sure if it something that needs to be fixed in the framework or such an edge case it needs to be looked out for.

The example I'm running off is: https://github.com/ubient/laravel-pwned-passwords/blob/master/src/PwnedPasswordsServiceProvider.php using `laravel-pwned-passwords` or `laravelPwnedPasswords` didn't work. It had to be `PwnedPasswords`.


driesvints:
Please send this to 8.x

PR #6832, opened 2021-02-12
Update sail.md

chrisk-7777:
Small correction on the stop procedure for containers running in detached mode. 
`down` (current docs) will stop **and** remove containers. The equivalent of `Control + C` would be `stop` which only stops (and doesn't remove). This can be seen in the terminal output when comparing `Control + C`, `stop` and `down`. 

The issue with running `down` is if any pruning is performed, those database volumes are now unlinked to containers and the data will be lost.

References:
 - https://docs.docker.com/compose/reference/down/
 - https://docs.docker.com/compose/reference/stop/

### DIFF
--- a/sail.md
+++ b/sail.md
@@ -90,10 +90,10 @@ sail up -d
 
 Once the application's containers have been started, you may access the project in your web browser at: http://localhost.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
No plans to change this right now.

mmanzano:
I agree with @chrisk-7777. Also it's unpleasant if you have some things configured in your local container like the vapor login.

PR #6834, opened 2021-02-15
[8.x] Adds disclaimer to breeze install command

fylzero:
Suggesting to add this since it seems to have caused someone some grief recently, per this post.

[https://laracasts.com/discuss/channels/laravel/careful-laravel-commands-are-not-careful](https://laracasts.com/discuss/channels/laravel/careful-laravel-commands-are-not-careful)

There isn't an explicit warning against adding this to existing projects, which may be useful.

### DIFF
--- a/starter-kits.md
+++ b/starter-kits.md
@@ -38,7 +38,7 @@ composer require laravel/breeze --dev
 
 After Composer has installed the Laravel Breeze package, you may run the `breeze:install` Artisan command. This command publishes the authentication views, routes, controllers, and other resources to your application. Laravel Breeze publishes all of its code to your application so that you have full control and visibility over its features and implementation. After Breeze is installed, you should also compile your assets so that your application's CSS file is available:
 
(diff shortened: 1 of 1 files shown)


driesvints:
I'm not sure if this is needed. It's documented under "starter kits" which should be indicative enough that this is for new installs only. The first sentence of the introduction also states:

> To give you a head start building **your new Laravel application,** we are happy to offer authentication and application starter kits.

fylzero:
@driesvints I think you're right but also, being that this particular kit is geared towards the absolute beginner maybe stating that explicitly isn't the worst idea.  There might be a less abrupt/verbose way of adding that than what I am presenting here.

fylzero:
Thinking about this more and looking at the docs it should be sufficient to state that it is a starter kit, also the instructions clearly say the first step of installing is to create a fresh application.  Not to mention a big warning on a beginner starter kit page claiming death and destruction is probably not ideal...  closing.

driesvints:
@fylzero still appreciate you trying to help out here with improving these docs :)

PR #6839, opened 2021-02-16
[8.x] Document wildcard subdomains with Valet

driesvints:
I spent quite a bit of time figuring out how to get subdomains to work with Valet but turns out Valet automatically allows you to use any wildcard subdomain with the base registered site 😅 

@mattstauffer @drbyte Hope I got this correct?

### DIFF
--- a/valet.md
+++ b/valet.md
@@ -126,7 +126,7 @@ The `park` command registers a directory on your machine that contains your appl
 
     valet park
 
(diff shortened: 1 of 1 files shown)


drbyte:
@driesvints Yes indeed it is supposed to work that way. I agree with how you've updated the docs.

There's an open issue 1032 reporting that some recent changes may have broken subdomain detection in certain cases.

PR #6840, opened 2021-02-16
Update http-client.md

DairoPortillo:
The request can be converted to object


taylorotwell:
I don't really find this method that helpful vs just using an array.

PR #6844, opened 2021-02-18
[8.x] Add descriptions for Envoy notifications.

akikgt:
Hi, I added descriptions about how to pass a message to Envoy notification directives.
I think it would be useful to make it explicit.

### DIFF
--- a/envoy.md
+++ b/envoy.md
@@ -198,12 +198,12 @@ If you would like to be prompted for confirmation before running a given task on
 <a name="slack"></a>
 ### Slack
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
What happens if you don't pass this argument? If you pass it - does it overwrite any default message that Envoy would generate?

akikgt:
If you don't pass this argument, it shows a default message.
For instance, the @slack's default message is `Runner ran a task.`
And yes, if you pass a message argument, it overwrites the default message.

taylorotwell:
This would make people think they have to pass a message because it doesn't talk about that behavior of overwriting the default message.

akikgt:
Thank you. I should've mentioned that a message is optional. I'll update the descriptions.

PR #6846, opened 2021-02-18
[8.x] Remove duplicate provider registration

abenettt:
Remove duplicate provider registration for Telescope.


driesvints:
This isn't duplicate. This is needed for local only installs.

abenettt:
But they are both inside the if statement and the only difference is the namespace that is being specified or am I getting this wrong?

driesvints:
If you read the paragraph above you'll notice that this is about two different `TelescopeServiceProvider` classes.

abenettt:
Ah indeed, I got that wrong, my bad.

PR #6855, opened 2021-02-22
[8.x] Missing extension docs for custom validation

vstyler96:
I included this section that was omitted from 7.x to 8.x, this is still working and not marked as obsolete in the Validator Facade.


taylorotwell:
Use custom rule objects.

PR #6857, opened 2021-02-22
[8.x] Only register passport routes if not cached

GrahamCampbell:
// cc @driesvints


driesvints:
This lgtm but it reminds me that Passport is our only package that handles routes this way. I might send in a PR soon to make them more conform our other packages.

PR #6872, opened 2021-02-28
Grammar correction

rochdiNassah:


taylorotwell:
Some of these corrections don't seem correct.

PR #6874, opened 2021-03-01
Update horizon.md

pramanikriju:
Added registering the Horizon Service provider as one of the installation steps. 
This is a crucial step to make the Horizon gate work and is a standard part of the installation, I feel it should be a part of the docs. 
This references [#563](https://github.com/laravel/horizon/issues/563#issuecomment-480882947) and makes it clearer for end users and beginners as well

### DIFF
--- a/horizon.md
+++ b/horizon.md
@@ -38,6 +38,16 @@ After installing Horizon, publish its assets using the `horizon:install` Artisan
 
     php artisan horizon:install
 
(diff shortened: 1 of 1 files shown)


driesvints:
This is already done by the install command: https://github.com/laravel/horizon/blob/5.x/src/Console/InstallCommand.php#L60-L64

PR #6878, opened 2021-03-02
[8.x] Add the `truncate` method to the Eloquent page

jeroenvanrensen:


driesvints:
Might need a linkback to the Query builder's truncate method that highlights some of the caveats: https://laravel.com/docs/8.x/queries#delete-statements

jeroenvanrensen:
Yeah that sounds like a good idea. I'll add it later today

PR #6892, opened 2021-03-09
[8.x] Document new ThrottlesExceptions job middleware

paras-malhotra:
Documentation for https://github.com/laravel/framework/pull/36473 and https://github.com/laravel/framework/pull/36518 for the upcoming release.

If https://github.com/laravel/framework/pull/36518 isn't merged before the upcoming release, we can just delete the last "tip" line from this PR.

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -10,6 +10,7 @@
 - [Job Middleware](#job-middleware)
     - [Rate Limiting](#rate-limiting)
     - [Preventing Job Overlaps](#preventing-job-overlaps)
+    - [Throttling Exceptions](#throttling-exceptions)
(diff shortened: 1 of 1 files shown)


paras-malhotra:
@taylorotwell wouldn't setting the cache key as the job UUID mean that subsequent jobs of the same class would not be short circuited? So, say I have a job that hits an API for each user. The API is down and I need all subsequent jobs (of the same class but maybe for different users) to be short circuited to give enough time for the service to recover.

I think it might be better to have this set to the job class.

taylorotwell:
I've updated it to use the class name by default and added a `byJob` method to quickly indicate it should use the UUID:

```php
new ThrottlesException(...)->byJob()
```

PR #6899, opened 2021-03-10
Grammar corrections

rochdiNassah:
Pronouns should be corrected depending on what is used in the examples `users` or `user`.


timacdonald:
I _really_ hope we are not going to replace “they” with “him” 👀😳

PR #6900, opened 2021-03-10
Add note about phpunit.xml server variables when running Artisan test command

PHLAK:
Added a note to clarify that the Artisan `test` command cannot access `phpunit.xml` server variables from service providers.

See https://github.com/laravel/framework/issues/36530 for the reason behind this note.


taylorotwell:
See @nunomaduro's latest comment on your issue.

PR #6905, opened 2021-03-11
Resolve typo on Redis section

MajidAlaeinia:
- Rename `names` to `name`


taylorotwell:
The two commands are not actually related.

PR #6907, opened 2021-03-14
Grammar correction

rochdiNassah:
"was" should be used instead of "were" with singular nouns.


rochdiNassah:
"Was" should be used instead of "Were" in singular noun.

PR #6919, opened 2021-03-16
Format correction

rochdiNassah:
It was the last missing Full Stop :).


driesvints:
Thanks! Next time: feel free to just send in one pr with all of these 👍

PR #6922, opened 2021-03-17
Fixed public file permission example to 0644

TTFerreira:
The code example for file and folder permissions had public files set to 0664 when it should be 0644


TTFerreira:
Realised now the example showed differences :-)

PR #6943, opened 2021-03-24
[8.x] Document the query builder's new `lazy()` method

JosephSilber:
Ref https://github.com/laravel/framework/pull/36699

---

I'll let you rewrite it in your own voice, but this should give you a good starting point.

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -11,6 +11,7 @@
 - [Retrieving Models](#retrieving-models)
     - [Collections](#collections)
     - [Chunking Results](#chunking-results)
+    - [Streaming Results Lazily](#streaming-results-lazily)
     - [Cursors](#cursors)
     - [Advanced Subqueries](#advanced-subqueries)
 - [Retrieving Single Models / Aggregates](#retrieving-single-models)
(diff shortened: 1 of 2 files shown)


taylorotwell:
Thank you ❤️

eloquent.md
```diff
@@ -329,47 +332,84 @@ Your application may run out of memory if you attempt to load tens of thousands
...
+### Streaming Results Lazily
+
+The `lazy()` method works similarly to [the `chunk` method](#chunking-results), executing the query in chunks. However, instead of passing each chunk into a callback, the `lazy()` method returns [a `LazyCollection`](/docs/{{version}}/collections#lazy-collections) of Eloquent models, which lets you interact with the results as a single stream:
+
+```php
+use App\Models\Flight;
+
+foreach (Flight::lazy() as $flight) {
+    //
+}
```
crynobone, line 366: @JosephSilber I'm wondering if `$flight` is an instance of Eloquent or a chunk set of Eloquent?

JosephSilber: > instead of passing each chunk into a callback, the `lazy()` method returns [...] results as a single stream

So `$flight` is a single Eloquent model.

JosephSilber: 1. I thought that was clear enough. Is it not?

2. If it isn't, would this help?

    image

crynobone: ability to pass `chunkCount` throws me off there. Understood now

JosephSilber: The `$chunkCount` determines how many records should be requested per chunk behind the scenes.

Bigger chunks use more memory, why smaller chunks means more queries (potentially exacerbated by eager loading relationships).

PR #6945, opened 2021-03-24
[8.x] Clarify that `release()` does not halt execution

jwcobb:
The documentation does not make it clear that using `release()` does not halt execution of the rest of the Job. This should be clearer.

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -1038,6 +1038,23 @@ By default, the `release` method will release the job back onto the queue for im
 
     $this->release(10)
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hmm - I'm not sure we ever indicate that it does halt execution?

jwcobb:
> Hmm - I'm not sure we ever indicate that it does halt execution?

The docs don't indicate it halts execution, but that was my assumption because it didn't make sense to me to put the job back on the queue if you're able to finish it. I spent a few hours debugging only to realize what the issue so I figured I would try and help the next person from making a wrong assumption.

PR #6950, opened 2021-03-26
Fix expose share command

mpociot:
There is no `valet expose` command available. In order to share a site through expose, you only have to call `expose` after its installed globally.


driesvints:
Thanks!

PR #6951, opened 2021-03-26
[8.x] Dynamodb table

driesvints:
Documents https://github.com/laravel/framework/pull/36776

I've also added a part to document on how people can use something like CloudFormation to set this up. Taken from the AWS docs here: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/SQLtoNoSQL.CreateTable.html


BrandonSurowiec:
Documentation will need to be updated since it references a non-existing command `cache:dynamodb` due to the related PR being rejected.

PR #6959, opened 2021-03-31
Add Str::startsWith array example

brysonreece:
Adds example showing the ability to pass multiple needles to Str::startWith.


brysonreece:
🎉 Thanks!

PR #6968, opened 2021-04-06
Schema Facade Correction

matthewscalf:
Corrected the path to the Schema facade under the soft deletes section on the Eloquent page.


driesvints:
thanks!

PR #6972, opened 2021-04-08
Update sail.md

judgej:
The `--ignore-platform-reqs` is needed to bootstrap the vendor directory on Windows because the `laravelsail/php80-composer` image does not have the common PHP extensions that many popular packages need, such as `ext-intl`.

Once the sail containers are built, it would be a good idea to run a `composer update` through sail (without the `--ignore-platform-reqs` option), just to make sure all the PHP extensions that the dependencies need really are available. Not sure if that should be added as a note too?

### DIFF
--- a/sail.md
+++ b/sail.md
@@ -145,7 +145,7 @@ docker run --rm \
     -v $(pwd):/opt \
     -w /opt \
     laravelsail/php80-composer:latest \
-    composer install
+    composer install --ignore-platform-reqs
 ```
 
 <a name="executing-artisan-commands"></a>


judgej:
Without this option, an application cloned from a repository that uses packages such as Spatie Media Library, will fail at this point; the vendor directory will not be built due to PHP extensions not being available.

driesvints:
If you need extensions that the docker image doesn't ship with it's probably best that you update the runtime file that's published with Sail: https://github.com/laravel/sail/blob/1.x/runtimes/8.0/Dockerfile

driesvints:
Also not checking platform requirements isn't wanted because packages won't actually work without these.

judgej:
That's not what this is about.

The composer container used in the updated command is for bootstrapping the vendor folder *before* anything else on Windows. The Docker file you describe is simply *not present* at this stage in the installation. The Docker file also *already* contains the extensions needed in most popular Laravel packages. Running `composer update` after this bootstrapping command again to ensure the required extensions are present in the *runtime* container will certainly be necessary to make sure the extensions that all packages the application needs are present.

I covered all that in the two posts above, but probably wasn't clear enough.

judgej:
Thank you. Hopefully, this will save some people some time getting started with Sail on Windows.

PR #6977, opened 2021-04-11
docs: add a note about unique jobs and queue clearing

joostdebruijn:
Based on the discussion in laravel/framework#36927.


taylorotwell:
I don't generally like to keep stacking "notes" in the docs. We either need to let it go or find a way to integrate it into the documentation more naturally.

PR #6978, opened 2021-04-12
Update filesystem.md

ngunyimacharia:
Document option for injecting SSH private key file contents into the configuration file. Current config file only documents adding file path to private key.

Configuration option documented in https://github.com/thephpleague/flysystem-sftp


taylorotwell:
I'm not sure this is the best way to document that.

PR #6986, opened 2021-04-14
[8.x] Fix checkout docs

yibr:
The quantity should be passed as the second argument in an array


driesvints:
Seems like this was an old implementation indeed.

driesvints (approved):

PR #6989, opened 2021-04-16
[8.x] Document that Laravel Sail doesn't work on M1 Macs

rosswintle:
Laravel Sail is not supported on M1/Apple Silicon Macs (see [this issue on Sail](https://github.com/laravel/sail/issues/104#issuecomment-821042352)).

While the ecosystem catches up, it should be noted that the recommend "quick start" method does not work on this hardware, and an alternative should be presented.

Thanks.

### DIFF
--- a/installation.md
+++ b/installation.md
@@ -59,6 +59,8 @@ Laravel Sail is a light-weight command-line interface for interacting with Larav
 <a name="getting-started-on-macos"></a>
 ### Getting Started On macOS
 
+> {tip} This method of getting started relies on Laravel Sail, which is not supported on M1/Apple Silicon-based Macs.  If you are using this hardware then please use another method for local development, such as [Laravel Valet](https://laravel.com/docs/8.x/valet)
+
 If you're developing on a Mac and [Docker Desktop](https://www.docker.com/products/docker-desktop) is already installed, you can use a simple terminal command to create a new Laravel project. For example, to create a new Laravel application in a directory named "example-app", you may run the following command in your terminal:
 
 ```nothing


driesvints:
Thanks @rosswintle. But this isn't 100% true. In some cases Sail on macOS does work. See [the tweet by Taylor](https://twitter.com/taylorotwell/status/1382803176575836162) that you shared and the replies. It's just that a couple of the tooling isn't fully compatible yet.

rosswintle:
I'm trying hard to be helpful here...

The docs also say "Sail provides a great starting point for building a Laravel application using PHP, MySQL, and Redis __without requiring prior Docker experience__." and I would argue that to fix this for Apple Silicon you do actually need to understand Docker.

Currently the quick-start throws up a bunch of errors and dies when you try to do certain things (e.g. `sail shell`).

I'm trying to remove the barrier to entry for people so that they don't throw their hands in the air and walk away frustrated on their first encounter with the framework.

What should the docs say, if not that Sail is not M1 compatible? What __is__ true?

rosswintle:
This is why I didn't PR the docs. They're your docs. I'll leave it to you to fix.

I've pointed out that the quick start doesn't work and is a barrier to entry to people on M1 who don't know Docker. That's all I can do here.

Thanks.

antja0:
Sad to see this not merged.
Thank you for trying @rosswintle.

This little notion could've saved me some time...

antja0:
Even some reference to Valet or alternative installation tools would've been great, in case sail doesn't work.
Please consider this 😄

PR #6993, opened 2021-04-17
Update information under `accepted` rule

ahangarha:


taylorotwell:
Unclear to me exactly what this means?

ahangarha:
If one uses `nullable|accepted`, the `accepted` rule overrides `nullable`. The difference between `nullable|boolean` and `nullable|accepted` is in the purpose of the second rule not their values. Boolean can be null but accpeted cannot be null. So `accepted` is also `required` in its nature so it overrides `nullable`.

I made this mistake once. I thought `nullable|accepted` means either null or some values equivalent to true. But I was wrong. So I propose this addition to the documentation for `accepted` rule so that it becomes clear what would happen.

PR #6996, opened 2021-04-19
Update sanctum.md

francoism90:
It's needed to specify the token column instead of record id.


taylorotwell:
Current code is correct.

PR #7000, opened 2021-04-20
[8.x] Remove sanctum middleware configuration duplication

KevinYobeth:
The docs mention the sanctum middleware installation for SPA authentication twice.


taylorotwell:
Documentation doesn't necessarily have to be DRY like code. We mention things multiple times occasionally in order to help people out in the flow of what they are reading.

PR #7006, opened 2021-04-20
[Sail] Add documentation about WWWUSER & WWWGROUP params

franck-grenier:
Hello Laravel,
I had troubles with permissions between my host and my sail-8.0/app container. So I dived into Sail Docker files and saw the commands aiming to change `sail` user and group IDs according to `WWWUSER` and `WWWGROUP` params.

It looks like other people had the same issues : https://laracasts.com/discuss/channels/laravel/groupadd-invalid-group-id-sail

So I decided to propose a small update of Sail documentation since the commands to handle this are already implemented.

Feel free to edit.

Hope you'll find it useful.

### DIFF
--- a/sail.md
+++ b/sail.md
@@ -21,6 +21,7 @@
 - [PHP Versions](#sail-php-versions)
 - [Sharing Your Site](#sharing-your-site)
 - [Customization](#sail-customization)
+- [Permissions and user mapping](#sail-user-permissions)
 
 <a name="introduction"></a>
 ## Introduction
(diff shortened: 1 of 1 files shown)


taylorotwell:
Eh - it's unclear when this should be used or under what circumstances? I definitely have never had to use it on Mac or Windows. Are you on Linux? If so - what distribution?

franck-grenier:
I'm on Linux, Ubuntu.

If I don't use WWWUSER and WWWGROUP to update the `sail` user/group IDs, I don't have write permissions on files generated by the container (ie `php artisan make:controller` or even log files).

If you don't like to chmod 777 all your files, I guess mapping the IDs is the good strategy in a Docker context.

And it seems that WWWUSER and WWWGROUP vars are made for this purpose.

You already have the functionality into Sail, it just needs to be documented.

PR #7011, opened 2021-04-21
[8.x] Fix docs according to the recently added `queueForget()` method

dmitry-ivanov:
This PR fixes the docs according to the [recently added `Cookie::queueForget()` method](https://github.com/laravel/framework/pull/37072).


taylorotwell:
I will make a note to document this on Tuesday when the feature is actually released.

PR #7018, opened 2021-04-22
 [8.x] Add clarification about  save() and  create() for saving on relations

Afrowson:
I encountered the the Issue that the specific types of relations where you can save a related model to via the save() and create() methods is not mentioned in the documentation. Calling ->save on a belongsTo relation will even create a new entry in to the database but wont set the id of the new entry to the child. 
These 8 words should clarify that the described way to save a model on a relation is only applicable for 'hasMany' and 'hasOne' relations.

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -1511,7 +1511,7 @@ Using these model definitions and relationships, we may retrieve `ActivityFeed` 
 <a name="the-save-method"></a>
 ### The `save` Method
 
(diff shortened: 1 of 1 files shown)


Afrowson:
@taylorotwell I dont know how to write tests for this. If the explanation is not sufficient, please explain what is missing.

taylorotwell:
This isn't accurate - like I said last time. morphMany - belongsToMany - morphOne, etc. all support save.

PR #7024, opened 2021-04-23
Add note about MaxAttemptsExceededException not releasing the lock on WithoutOverlapping middleware

lasselehtinen:
This was triggered by the discussion in [this issue](https://github.com/laravel/framework/issues/37060). It makes sense to clarify that if you do not define the expiresAfter, certain exceptions will not release the lock automatically.

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -346,6 +346,8 @@ Any overlapping jobs will be released back to the queue. You may also specify th
         return [(new WithoutOverlapping($this->order->id))->releaseAfter(60)];
     }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't really understand this paragraph so I'm not sure everyday users will. @paras-malhotra can you maybe take a shot at it?

paras-malhotra:
Sure, I've opened #7025. @lasselehtinen you need to use `expireAfter` and not `releaseAfter`.

PR #7032, opened 2021-04-26
Added missing modifiers for columns in the Database

Danil42Russia:
Updated based on the code: [ColumnDefinition](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Schema/ColumnDefinition.php#L7-L31)


taylorotwell:
Much of this is covered under a separate section on indexes. These example don't always do a great job of actually telling the user how to use these features.

Danil42Russia:
@taylorotwell, thank you, I realized my mistake. If I remove the mention of indexes, it will somehow help to merge the PR?

PR #7045, opened 2021-05-04
[8.x] Fix confirmation prompt second argument's default is inverted

Tofandel:
See https://github.com/symfony/console/blob/5.x/Style/SymfonyStyle.php#L291


Tofandel:
Hmm after reviewing it seems there is a difference between laravel and symfony which sometimes lead to confusion when using 
 `$this->confirm()` and `$this->output->confirm()`
 
 Should probably add a warning instead
 
 https://github.com/laravel/framework/blob/8.x/src/Illuminate/Console/Concerns/InteractsWithIO.php#L131

PR #7046, opened 2021-05-04
[8.x] Add a note about Symfony's different default attribute for confirm()

Tofandel:
See

https://github.com/laravel/framework/blob/8.x/src/Illuminate/Console/Concerns/InteractsWithIO.php#L131
https://github.com/symfony/console/blame/5.x/Style/SymfonyStyle.php#L291

Which is confusing

Superseeds #7045


Tofandel:
?

PR #7048, opened 2021-05-06
[8.x] Document rendering signed route failures

driesvints:
After experiencing https://freek.dev/1977-dealing-with-expired-signed-urls-in-laravel?preview_secret=I9e6ZvUlTU I felt it might be worth to document this. Thanks to @freekmurze 

Although I feel like we could do better out of the box with a better message for the 403 exception? 🤔


taylorotwell:
@driesvints there is no `render` method in the Laravel 8 exception handler.

PR #7057, opened 2021-05-11
add missing return statement

wi-wissen:


taylorotwell:
Unnecessary to return here in this case.

PR #7063, opened 2021-05-14
Add Fortify logout redirection info

trbsi:


taylorotwell:
For something so simple maybe it would be better to PR Fortify an easier way to change this?

taylorotwell:
I have pushed up my own commit to document this. Thank you.

PR #7064, opened 2021-05-14

generalov-viktar:
The example is initially wrong.
Must be included in the receipt of book_id


generalov-viktar:
This is necessary for a hasMany relationship

PR #7066, opened 2021-05-15
Add info on Sail setup for broadcasting

n8udd:


taylorotwell:
This depends entirely on what you are using as your websocket service.

PR #7076, opened 2021-05-21
cashier: Update stripe webhook url

ankurk91:
Hi, @driesvints 

We should only whitelist `stripe/webhook/*` not the whole `stripe/*`, as many of users would be having routes starting with `stripe`

Here are some examples
```
PUT stripe/subscriptions/1
POST stripe/subscriptions/create
DELETE stripe/subscriptions/1
GET stripe/connect-callback
```

Laravel cashier only has one endpoint to be whitelisted.


driesvints:
Well, not really. There's also the payment page: `/stripe/payment`. I don't think many people are using `stripe/` as their base url and you're still free to customize this so if you need to change this feel free to do so in your app 👍

PR #7078, opened 2021-05-21
curl must using `-S` flag, not `-s`

albasyir:
i have try on my mac big sur, curl version 7.71.1 (x86_64-apple-darwin13.4.0), use `-s` but not work, after that i try use `-S`, it's work perfectly


taylorotwell:
image

PR #7083, opened 2021-05-24
Add mailto tag to email

radosslav:


taylorotwell:
We should really setup something like a Google Form for this. 😓

PR #7087, opened 2021-05-25
Replace the term attribute into relation for describing the pivot model

ajcastro:
…el because it is actually a relation in the model


taylorotwell:
It isn't a full relationship object.

PR #7096, opened 2021-05-29
Update "Preventing Jobs Overlapping" - queue.md

back2Lobby:
It wasn't clear how jobs can overlap when they are being processed one by one. I have never used multiple workers soo far (because I am still learning). So, I couldn't understand the logic behind "Jobs Overlap". Then I found out from stack overflow that it occurs when you use multiple queue workers.


taylorotwell:
Jobs don't come "from" queue workers. I think the existing wording is pretty satisfactory.

PR #7097, opened 2021-05-30
[8.x] Add existsOr and doesntExistsOr methods

kevinb1989:


taylorotwell:
I think we would need a good example of what you might actually do within those closures.

PR #7102, opened 2021-06-01
[8.x] Explain timezones for custom datetime format

driesvints:
After investigating https://github.com/laravel/framework/issues/37465 we came to the conclusion that there's a discrepancy between a regular datetime cast and one with a custom format. The one with the custom format will have the `app.timezone` applied when serialized while the other one will have UTC applied (default Carbon behavior). 

I feel this is an important gotcha that we need to note about in the docs.

### DIFF
--- a/eloquent-mutators.md
+++ b/eloquent-mutators.md
@@ -267,7 +267,7 @@ When defining a `date` or `datetime` cast, you may also specify the date's forma
         'created_at' => 'datetime:Y-m-d',
     ];
 
-When a column is cast as a date, you may set its value to a UNIX timestamp, date string (`Y-m-d`), date-time string, or a `DateTime` / `Carbon` instance. The date's value will be correctly converted and stored in your database.
+When a column is cast as a date, you may set the corresponding model attribute value to a UNIX timestamp, date string (`Y-m-d`), date-time string, or a `DateTime` / `Carbon` instance. The date's value will be correctly converted and stored in your database.
 
 You may customize the default serialization format for all of your model's dates by defining a `serializeDate` method on your model. This method does not affect how your dates are formatted for storage in the database:
 
--- a/eloquent-mutators.md
+++ b/eloquent-mutators.md
@@ -267,7 +267,7 @@ When defining a `date` or `datetime` cast, you may also specify the date's forma
         'created_at' => 'datetime:Y-m-d',
     ];
 
-> {note} It's important to note that when using a custom date format using the `date` or `datetime` cast that the `app.timezone` will be applied.
+> {note} It's important to note that when using a custom date format using the `date` or `datetime` cast that the time will not be converted to UTC when the date is serialized.
 
 When a column is cast as a date, you may set its value to a UNIX timestamp, date string (`Y-m-d`), date-time string, or a `DateTime` / `Carbon` instance. The date's value will be correctly converted and stored in your database.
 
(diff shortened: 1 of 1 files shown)


kylekatarnls (approved):

eloquent-mutators.md
```diff
@@ -267,7 +267,9 @@ When defining a `date` or `datetime` cast, you may also specify the date's forma
         'created_at' => 'datetime:Y-m-d',
     ];
 
-When a column is cast as a date, you may set its value to a UNIX timestamp, date string (`Y-m-d`), date-time string, or a `DateTime` / `Carbon` instance. The date's value will be correctly converted and stored in your database:
+> {note} It's important to note that when using a custom date format using the `date` or `datetime` cast that the `app.timezone` will be applied.
```
kylekatarnls, line 270: Not sure we have to insist on that point but, in ISO-8601 strings, only the timezone offset of the date will appear without keeping the whole (`app.timezone`) information in the string.

It's correct regarding the datetime but it means guessing what was the exact timezone from it accurately is not possible (multiple timezone names can result into the same offset, and a named timezone can have multiple possible offsets due to DST).

eloquent-mutators.md
```diff
@@ -267,7 +267,9 @@ When defining a `date` or `datetime` cast, you may also specify the date's forma
         'created_at' => 'datetime:Y-m-d',
     ];
 
-When a column is cast as a date, you may set its value to a UNIX timestamp, date string (`Y-m-d`), date-time string, or a `DateTime` / `Carbon` instance. The date's value will be correctly converted and stored in your database:
+> {note} It's important to note that when using a custom date format using the `date` or `datetime` cast that the time will not be converted to UTC when the date is serialized.
```
kylekatarnls, line 270: Nope, it's the contrary:
- With a custom format: app.timezone is applied
- Without a custom format: UTC is enforced as per ISO-8601 norm (as it's the format used by default for JSON and maximizing interoperability)

driesvints: @kylekatarnls I don't think that's true. Taylor and I both verified that `datetime` cast with a custom format ignores the UTC conversion. That's the real explanation.

kylekatarnls (changes requested):
Please precise concrete difference between custom format vs. default.

eloquent-mutators.md
kylekatarnls: Sorry, but this technical explanation is helpless for the reader. It does not tell anything about the concrete results and difference you'll get.

kylekatarnls: Some first other topic to document would be how Laravel creates date objects from DB datetime strings and what timezone it picks to read those strings (e.g. `config('app.timezone')`).

Then for the custom date format, I think both scenario should be mentioned (with sentence that does not use negation), so the reader is not left into the doubt about *what happen in the other cases*:
```suggestion
> {note} It's important to note that `date` or `datetime` properties are output by default as UTC ISO-8601 string, then as soon as a custom date format is used, UTC will no longer be enforced, so inner timezone of the `Carbon` instance applies (refer to [How DB datetime fields are converted to Carbon instances section])
```

taylorotwell:
@kylekatarnls what do you think of this PR now after my latest change?

kylekatarnls:
I didn't mean the first version was bad. Thanks guys for your work there. I'll try to give a hand to get timezone in and out from/to model more consistent in Laravel 9.

PR #7104, opened 2021-06-02
Fix for Octane Port with Sail

arhea:
When spinning up Octane with Sail the command needs to use port `80` instead of port `8000` for the roadrunner server.


taylorotwell:
Please explain why?

PR #7109, opened 2021-06-03
Removed the markup HTML that spoiled the display Markdown

Sergey-Shabanov:
The document contained incorrect markup that messed up the display of the Included Software and Optional Software list


Sergey-Shabanov:
Found out that remote markup is necessary for correct display of a multi-column list on the documentation site

PR #7111, opened 2021-06-03
[8.x] add foreignIdFor to list of collection methods

tcampbPPU:
I saw this method being used in a migration example on Laravel news [article](https://laravel-news.com/disable-eloquent-lazy-loading-during-development) but could not find example of this in current docs.


taylorotwell:
I generally am not a *huge* fan of recommending this method. I think you should just be explicit about the column names you want. Thanks though!

PR #7118, opened 2021-06-06
[8.x] Sail Queue Workers

andriichuk:
Hi there! Added queue workers installation manual for Sail environment.


taylorotwell:
Just run `sail artisan queue:work`? You don't need an entirely separate container.

PR #7119, opened 2021-06-06
[8.x] Valet - Added  "tld" to "other commands"

dansysanalyst:
- I did read the doc a couple of times before having to resort to CLI valet list.
- One doc line may improve user experience with valet.


taylorotwell:
I don't like to document this. I prefer people use the default TLD.

PR #7120, opened 2021-06-07
Added version numbers for Password class

apokryfos:
Added the version number that the `Password` class and `Password::defaults()` method were added. There seems to be confusion as to whether the method is available otherwise as seen in https://stackoverflow.com/questions/67875161/does-defaults-method-exist-in-password


taylorotwell:
All docs should be assumed to require the latest Laravel 8 version if you're on the Laravel 8 branch.

PR #7121, opened 2021-06-07
[8.x] Indicate correct service provider method

driesvints:
The `loadViewComponentsAs` call, just like the `callAfterResolving` call belongs in the `register` method of a service provider since it wires dependency resolution.

### DIFF
--- a/packages.md
+++ b/packages.md
@@ -246,17 +246,17 @@ Now, when users of your package execute Laravel's `vendor:publish` Artisan comma
 <a name="view-components"></a>
 ### View Components
 
-If your package contains [view components](/docs/{{version}}/blade#components), you may use the `loadViewComponentsAs` method to inform Laravel how to load them. The `loadViewComponentsAs` method accepts two arguments: the tag prefix for your view components and an array of your view component class names. For example, if your package's prefix is `courier` and you have `Alert` and `Button` view components, you would add the following to your service provider's `boot` method:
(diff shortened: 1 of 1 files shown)


xel1045:
Since they are all using `callAfterResolving`, I guess this change should also be applied to:

- [loadMigrationsFrom](https://laravel.com/docs/8.x/packages#migrations)
- [loadTranslationsFrom](https://laravel.com/docs/8.x/packages#translations)
- [loadViewsFrom](https://laravel.com/docs/8.x/packages#views)
- [loadMigrationsFrom](https://laravel.com/docs/8.x/packages#migrations)

driesvints:
@xel1045 hmmm, that's interesting. And I see we call those methods from the boot method in our own packages. I do believe that these should be called in the register method. I'll discuss this with the team.

driesvints:
@xel1045 okay so we got to the bottom of this: in general it should actually not matter wether `callAfterResolving` is placed in `register` or `boot`. It should maybe be better to place the methods in `register` to make sure the services are wired into the container as soon as possible to reduce the chance of conflicts. 

It turns out that the Blade Icons one is a special case. There, the `callAfterResolving` needs to be in `register` because we need to make sure components are loaded into the framework as soon as possible before anyone tries to resolve `view` in a `boot` method of their service provider and potentially render a view with icon components in them. So in this situation we *need* to place it in register. As said before, the reason why it was causing issues for people was because they were incorrectly instantiating the `view` service already in their `register` method of their service provider and that should always be avoided. 

I hope that's clear. I reverted this PR to let the docs stay consistent.

xel1045:
Yes, this is clearer.

Thanks a lot for all your research!

PR #7124, opened 2021-06-10
Update database-testing.md

fridzema:


taylorotwell:
This doesn't match what is actually in the factory. This is done to greatly increase performance when creating a lot of users via factories.

PR #7131, opened 2021-06-13
[8.x] Document Homestead "paravirtualization" setting

ktross:
The `paravirtualization` setting, added in [this commit](https://github.com/laravel/homestead/commit/af3a5f1c521f493893762371d894ffe12801c37f), is missing from the docs.

I was trying to get Docker and Vagrant/VirtualBox to run alongside each other on Windows 10 with WSL2/Hyper-V and had to dig through the source to find this option. Documenting it might save users some time and frustration.

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -175,6 +175,15 @@ The `provider` key in your `Homestead.yaml` file indicates which Vagrant provide
 
     provider: virtualbox
 
+<a name="setting-your-paravirtualization-interface"></a>
+#### Setting Your Paravirtualization Interface
+
+The `paravirtualization` key in your `Homestead.yaml` file indicates which paravirtualization interface should be provided to the guest operating system:
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not sure this documentation explains when and why to use the option?

ktross:
@taylorotwell My intent was more to let users that are already searching for this option know that Homestead supports it. I don't think most users will ever need to use this, and if they do, it's likely they already know why. The when and why is probably outside the scope of Homestead docs, and better left to the VirtualBox/VBoxManage docs.

If this setting was documented (even lightly) in the Homestead docs, it might have saved me an hour or two.

https://www.virtualbox.org/manual/ch10.html#gimproviders
https://www.virtualbox.org/manual/ch08.html#vboxmanage-modifyvm

PR #7135, opened 2021-06-16
[8.x] Document scout eager loading

edgarsn:
As discussed in laravel/scout#481 about eager loading in scout, it's good to mention to use lazy eager loading instead of

```
->query(function ($query) {
     $query->with('user');
})
```


taylorotwell:
Not much context here.

PR #7140, opened 2021-06-21
Include a Note that Breeze Should Be Run on New Application

dev-talha-akbar:
I noted that running breeze install command on an existing application overwrites the existing files which will be quite undesirable if the developer doesn't have any backup of the existing project. I personally could have lost some progress on the project because I was not careful enough to read the instruction in the docs that it should be run on a new Laravel project however, it should also be considered that the documentation didn't emphasize this limitation enough. With the addition of the new note, it is hoped that other developers will take note of such disastrous consequence.

### DIFF
--- a/starter-kits.md
+++ b/starter-kits.md
@@ -47,6 +47,8 @@ npm run dev
 php artisan migrate
 ```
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
This first sentence of the installation instructions says to create a new Laravel application.

dev-talha-akbar:
I had the following concern:

> the documentation didn't emphasize this limitation enough.

Anyway, you know best 👍

PR #7150, opened 2021-06-23
Fix typo in Fortify docs

tnorthcutt:
`All of Fortify's view's rendering logic` --> `All of Fortify's view rendering logic`


driesvints:
Thanks!

PR #7153, opened 2021-06-24
[8.x] Clarify dispatchSync behavior

driesvints:
Clarifies some gotchas when using `dispatchSync`. See https://github.com/laravel/framework/issues/37778

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -687,6 +687,8 @@ If you would like to dispatch a job immediately (synchronously), you may use the
             ProcessPodcast::dispatchSync($podcast);
         }
     }
+    
+When using the `dispatchSync` method, no value is returned from the job's `handle` method and no job middleware is executed.
 
 <a name="jobs-and-database-transactions"></a>
 ### Jobs & Database Transactions


sharifzadesina:
This doesn't explain it well,

`dispatchSync` runs the middleware if you have `ShouldQueue` interface.

and the actual value will be returned if you don't use `ShouldQueue` interface.

Simple, the whole behavior of the method depends on `ShouldQueue` interface marker.

Or are you planning to unify the behavior of the method in the future? (I hope this is the case).

sharifzadesina:
As the `dispatchNow` method is deprecated now,

And the jobs are for running something on a separate process (async or sync), It is logical that it doesn't return actual value.  I think the doc should be changed like this:

> When using the `dispatchSync` method, no value is returned from the job's `handle` method.

Maybe we also need this:

> Jobs must always implement the `ShouldQueue` interface.

This way, the middleware always runs, and never the actual value will be returned, and people will create a new **service layer** if they want to run something and get the value.

taylorotwell:
I'm going to comment on the issue.

PR #7161, opened 2021-06-29
[8.x] Clarify that empty query string variables are ignored in pagination

driesvints:
Because of the usage of `http_build_query` to build up the query string, empty query string variables are ignored.

https://github.com/laravel/framework/blob/93e104d96924019986b24350f5090409725f8792/src/Illuminate/Collections/Arr.php#L652

https://3v4l.org/FEaaS

See https://github.com/laravel/framework/issues/37846


taylorotwell:
I don't see a big need to document this to be honest. I feel like very rare edge cases like this just clutter the documentation a bit for something that may only come up once every few years from one person.

PR #7162, opened 2021-06-29
[8.x] Remove production warning notice from Dusk docs

dakira:
In the current version of DuskServiceProvider, the login routes are only registered when not in production.

https://github.com/laravel/dusk/blob/6.x/src/DuskServiceProvider.php#L17

So this warning should not be needed anymore.


taylorotwell:
I think we can still leave it in. People may not call their "production" environment "production".

PR #7170, opened 2021-07-05
[8.x] Clarify UUID's and one-of-many with PostgreSQL

driesvints:
It's at the moment not possible to use UUID's in combination with one-of-many in PostgreSQL because of PostgreSQL's lack of an implementation of `MAX` for UUID's. Other engine's like MySQL do support this. I am not aware of an alternative for PostgreSQL to find the max UUID. Therefor I think it's important to document this.

See the related issue here: https://github.com/laravel/framework/issues/37854

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -318,7 +318,7 @@ public function largestOrder()
 }
 ```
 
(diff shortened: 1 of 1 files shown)


pushpak1300:
Thanks, @driesvints

PR #7171, opened 2021-07-06
adding warning about type hinting of form request objects

thyseus:
when using `laravel/octane`.


taylorotwell:
This is not entirely correct. Can you be more specific? You can definitely type hint a form request on any controller method.

thyseus:
Thanks - i was very confused about the sentence "In general, you should avoid injecting the application service container or HTTP request instance into the constructors of other objects.".

Actually there already is a hint below that is is _acceptable_ to inject Request Objects into Controllers, i have missed it.

PR #7173, opened 2021-07-06
[Cashier] Missing mandatory parameters to downloadInvoice

matthieumota:
I think there is a little typo about `downloadInvoice()`, method needs `data` array parameter with at least `vendor` and `product` keys.

### DIFF
--- a/billing.md
+++ b/billing.md
@@ -1534,7 +1534,10 @@ From within a route or controller, you may use the `downloadInvoice` method to g
     use Illuminate\Http\Request;
 
     Route::get('/user/invoice/{invoice}', function (Request $request, $invoiceId) {
-        return $request->user()->downloadInvoice($invoiceId);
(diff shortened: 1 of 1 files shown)


driesvints:
This isn't correct. In the latest v13 these aren't required anymore.

matthieumota:
I suspected because I had seen the modification on the documentation. But I haven't found this anywhere in code base on [13.x](https://github.com/laravel/cashier-stripe/blob/13.x/src/Concerns/ManagesInvoices.php#L225) branch. Maybe I missed it.

driesvints:
@matthieumota I'm sorry but you're correct. This is still required. I forgot to assign new defaults while working on v13 🤦‍♂️

matthieumota:
I'm ok to open PR but I don't know what default values to product and vendor, maybe `APP_NAME` or empty value ?

driesvints:
I'll take care of it in the next release 👍

PR #7178, opened 2021-07-08
Add missing parameter doc for @error directive.

Slipy90:


taylorotwell:
Instead of a "tip", can you provide an actual example? Please mark as ready when updated.

Slipy90:
Sure and I pushed my actual example.

PR #7182, opened 2021-07-11
Add hint about performance when using whereDate etc.

oliworx:
Unfortunately I see these methods a lot in Laravel tutorials and in real project. People dont know what they are doing, but think this is just a fancy way to query the database.

But if you depend on indexes for database performance, because there is a signifiat amount of data and a significant load, these methods can make the query really slow and even break the application.

### DIFF
--- a/queries.md
+++ b/queries.md
@@ -571,6 +571,8 @@ The `whereTime` method may be used to compare a column's value against a specifi
                     ->whereTime('created_at', '=', '11:20:45')
                     ->get();
 
+> {note} These methods can result in significant performance degration, because the database can not make use of indexes for the `where` contions!
+
 **whereColumn / orWhereColumn**
 
 The `whereColumn` method may be used to verify that two columns are equal:
--- a/queries.md
+++ b/queries.md
@@ -571,7 +571,7 @@ The `whereTime` method may be used to compare a column's value against a specifi
                     ->whereTime('created_at', '=', '11:20:45')
                     ->get();
 
-> {note} These methods can result in significant performance degration, because the database can not make use of indexes for the `where` contions!
+> {note} These methods can result in significant performance degration, because the database can not make use of indexes for the `where` conditions!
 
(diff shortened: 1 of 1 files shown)


queries.md
```diff
@@ -571,6 +571,8 @@ The `whereTime` method may be used to compare a column's value against a specifi
                     ->whereTime('created_at', '=', '11:20:45')
                     ->get();
 
+> {note} These methods can result in significant performance degration, because the database can not make use of indexes for the `where` contions!
```
oliworx, line 574:
```suggestion
> {note} These methods can result in significant performance degration, because the database can not make use of indexes for the `where` conditions!
```

taylorotwell:
I'm not sure it's our job to note various database indexing tips, etc.

SilvioDoMine:
@taylorotwell I was making a PR to add this information into the documentation and I found this.

> I'm not sure it's our job to note various database indexing tips, etc.

Well, this isn't really a indexing tip since this problem only occours with Laravel date/time where functions. This is the query we are trying to run in Laravel:

**Query builder:**
image

**Raw query:**
image

**The query explain, show no indexes:**
image

**Now we perform the exact same raw query directly into the database:**
image

@taylorotwell as you can see, without QueryBuilder/Eloquent we are able to successfully retrieve the data using indexes. Since this is a Laravel problem, we should update our documentation until we find a fix for the issue.

You can merge this PR, or I can make a PR to add this to the docs in a more explicit way:
image

oliworx:
@SilvioDoMine It is not the exact same query! The Eloquent way results in a **prepared statement**, where the date string is the parameter. The other example is not a prepared statement, but the date string is hard coded in the query. Maybe in this case the database is smart enough to use the index on wallet_id and created_at, but still reading 5 Million rows.

Can you please post the results of `show create table transactions;` and `select version();`

You should try to build a more simple example of this behaviour.

PR #7183, opened 2021-07-12
Fix namespace in Mail example

LucaRed:
The new routing is namespaced, so a back slash is required to import classes.


taylorotwell:
The new routing isn't namespaced though?

PR #7192, opened 2021-07-15
[8.x] Document returning Htmlable from Blade component's render method

nikazooz:
This documents the hidden feature of returning dynamically generated HTML from Blade component's render method by using object that implements `Illuminate\Contracts\Support\Htmlable` to avoid treating it as an inline template.


taylorotwell:
Has typos.

nikazooz:
Will this be reconsidered once the typos are fixed?

PR #7196, opened 2021-07-15
Reminder to confirm the referer header is present

joaojoyce:
Hi! 

Some new frontend frameworks like nextjs allow users to make client and server side requests.
After following the docs to configure Sanctum I realized my application could do client side requests but no server side requests.

After a lot of trial and error I've decided to dig deeper into Sanctum to debug the issue and found out that the reason was that when nextjs was making server side requests the request ended up with no "referer" header.

I believe other people had similar issues, for instance here https://github.com/laravel/sanctum/issues/229

I propose that we add some note to the docs about this issue.

### DIFF
--- a/sanctum.md
+++ b/sanctum.md
@@ -239,6 +239,8 @@ In addition, you should enable the `withCredentials` option on your application'
 
     axios.defaults.withCredentials = true;
 
+If your application makes server side requests in order to server side render content make sure the `referer` header is correctly filled with your application domain.
+
 Finally, you should ensure your application's session cookie domain configuration supports any subdomain of your root domain. You may accomplish this by prefixing the domain with a leading `.` within your application's `config/session.php` configuration file:
 
     'domain' => '.domain.com',


taylorotwell:
We would need more context in the documentation about what this is referring to. It seems disconnected and random atm.

ericwang401:
I am the author of issue https://github.com/laravel/sanctum/issues/229
I recommend merging this info in the documentation.
Whenever developers using Sanctum want to test their API routes in Postman, they might forget to include the Referer header in the request in Postman.

As a result, without the Referer header, API routes with the Sanctum middleware would return `{ "message" : "unauthenticated." }` with no hint or suggestion to add the Referer header.

I remember debugging this issue a while ago, and I felt like pulling out my hair because I didn't know what was causing the API route to return `{ "message" : "unauthenticated." }`. I probably spent 1-2 days fiddling with the RouteServiceProviders, middlewares, and the vendor directory. 

Sending back `{ "message" : "unauthenticated." }` can confuse or mislead developers because it doesn't tell them that something is wrong with their request's headers and they might spend a load of time debugging their backend when it is just a misconfiguration with their headers.

Merging this info in the docs can help potential developers who are facing this issue when they are developing and testing their SPA routes.

PR #7198, opened 2021-07-17
routing: Add regex constraint example when customizing the model key

julienbourdeau:
When using a regex constraint and customizing the key of a route parameter, you must only refer the first part (before `:`) as the name.

I'm not sure a paragraph is needed but I think an snippet example can help.


taylorotwell:
I'm not really sure what this is demonstrating tbh.

PR #7201, opened 2021-07-20
Update broken link to Symfony upgrade guide

JCarlosR:
Current `master` branch in the Symfony project doesn't have an `UPGRADE-4.0.md` file so it gives a 404 error.

Link updated.


driade:
Hi @taylorotwell, good afternoon. I'm not sure if this is the right place for this, but it seems this change is not online yet (https://laravel.com/docs/5.6/upgrade), so I may assume it's something broken somewhere as it's more than one year since it was merged.

PR #7206, opened 2021-07-21
Add link to Available Methods from each method

akulmehta:
When browsing/hunting for available methods it is probably helpful to have a link which takes you back to the index above of Available Methods so that you can view all the methods in one glance. Since the methods are alphabetical but often searching requires to view all methods as language does not necessarily club all similar words alphabetically e.g. `sole()` vs `filter`. This adds a small `^` symbol next to each methods title to take back to the top. We can probably add similar links in other sections of the documentation.

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -237,7 +237,7 @@ For the majority of the remaining collection documentation, we'll discuss each m
 </style>
 
 <a name="method-all"></a>
-#### `all()` {#collection-method .first-collection-method}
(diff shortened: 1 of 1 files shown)


akulmehta:
Just a follow-up there is an issue which I am not sure how to resolve. In the above case for example, the same methods can be linked from the `Available Methods` section and also from the `The Enumerable Contract` section. So should we just link to the `Available Methods`? Would like thoughts on this.

PR #7207, opened 2021-07-21
add baseUrl section

khalyomede:
This pull request adds a section to explain how to use the method [`Http::baseUrl()`](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Http/Client/PendingRequest.php#L174) on the documentation of the [HTTP client](https://laravel.com/docs/8.x/http-client) in version 8.X.


taylorotwell:
I don't really suggest doing this.

PR #7214, opened 2021-07-28
Update scope example to use `qualifyColumn` method

trovster:
I came across an issue regarding ambiguous where clause and solved this using the built-in `$model->qualifyColumn()` method. I thought the example would be better of using it to mitigate this issue in the future.


taylorotwell:
It may not be necessary for all situations.

PR #7221, opened 2021-08-03
Details about signal handling for workers

emil-nasso:
Add to the imformation about `queue:listen` to make the difference between `queue:listen` and `queue:work` clearer.

See issue: <https://github.com/laravel/framework/issues/38116>

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -1410,6 +1410,8 @@ Remember, queue workers, are long-lived processes and store the booted applicati
 Alternatively, you may run the `queue:listen` command. When using the `queue:listen` command, you don't have to manually restart the worker when you want to reload your updated code or reset the application state; however, this command is significantly less efficient than the `queue:work` command:
 
     php artisan queue:listen
+    
+Because the `queue:work` command needs to be restarted when your code is changed, it will listen to termination signals and exit gracefully by allowing running jobs to finish before shutting down the worker. For this to work, the php module `pcntl` needs to be installed. Please note that `queue:listen` doesn't handle termination signals and shouldn't be used if you plan on restarting the queue worker frequently as you risk having jobs terminated during execution. An example of when this can become an issue is when using `queue:listen` for an application deployed using docker.
 
 <a name="running-multiple-queue-workers"></a>
 #### Running Multiple Queue Workers


emil-nasso:
Please see the linked issue for background.

I am VERY open to feedback on how this extra information is written and the approach that was taken when trying to explain the issue. :)

taylorotwell:
Hmm, what part exactly do you feel is missing from the documentation currently? What was confusing to you?

emil-nasso:
@taylorotwell I should have written more of a summary about the issue here and not just linked the issue.

After reading the documentation, I thought that the only difference between `queue:listen` and `queue:work` was that `queue:work` reused the same application instance for each processed job. `queue:listen` re-creates the application for each job, for the convenience of not having to restart it if code is changed and having a "fresh application" each run, with the tradeoff of worse performance.

This is the only information available about the command in the documentation:

>Alternatively, you may run the queue:listen command. When using the queue:listen command, you don't have to manually restart the worker when you want to reload your updated code or reset the application state; however, this command is significantly less efficient than the queue:work command:

However, we became aware, in production, that `queue:listen` doesn't handle signals (like sigterm) the same way as `queue:work`.

When `queue:work` receives a signal, it is handled and the process is terminated once it is no longer processing a job, according to this: https://github.com/laravel/framework/blob/8.x/src/Illuminate/Queue/Worker.php#L676-L688

When `queue:listen`, no similar signal handling exits, and the signal is not handled.

In docker, when we deploy new versions of our apps, a termination signal is sent to the container and forwarded to the PHP process. When we use `queue:work` this signal is handled and it shuts down gracefully and quickly. When we use `queue:listen` the signal is not handled and docker waits a few minutes until it times out and forcefully kills the container. The result of that is that `queue:listen` can be killed in the middle of the execution of a job if we are unlucky.

A short section about how signals are handled and the difference between `queue:work` and `queue:listen` in that regard would be great and would have saved us a lot of headache and time in troubleshooting. :)

I am very open to rephrasing or completely rewriting my suggestion though.

taylorotwell:
Personally I would just suggest we add SIGTERM support to `php artisan queue:listen`. It is likely pretty easy to accomplish if you would like to make that PR instead.

PR #7225, opened 2021-08-08
Collections: mapWithKeys example that changes keys

levu42:
Provides an example for the Collection::mapWithKeys() method that has a callback changing the keys, since this possibility of changing the key was not mentioned in the documentation before.


taylorotwell:
I added a key example.

PR #7229, opened 2021-08-12
Create an Invoice with multiple line items

itsromack:


taylorotwell:
I don't think this makes sense as the very first thing you read when you click Invoices in the documentation.

PR #7233, opened 2021-08-13
[8.x] Adds the Rate Limiter section

DarkGhostHunter:
Since the `RateLimiter` now has its own Facade, I think it would be fine to document it even if it's usage is simpler than any other service.


taylorotwell:
Marking as draft until Tuesday.

PR #7235, opened 2021-08-14
[8.x] Facade singleton behaviour explanation

JayBizzle:
Resubmitting #6041 which was as a result of https://github.com/laravel/ideas/issues/1088


taylorotwell:
I don't really want to encourage facades that don't act like the usual facades.

PR #7230, opened 2021-08-12
Added tip to make using selenium-chrome-debug easier in sail

Patabugen:
I thought adding a tip to easily enable visualising Browser tests would be helpful (it'd have saved me a bunch of time today!). I've tried to get the wording/phrasing right this time - happy for any feedback!

PR #7238, opened 2021-08-17
Added docs to make using selenium-chrome-debug easier in sail

Patabugen:
Hey,

This is an updated version of https://github.com/laravel/docs/pull/7230 based on the feedback there.

I struggled to get enough information into a `tip` which was easy to read and also clear, so I've expanded it into a section (alongside the Laravel Dusk one underneath the "Running Tests" heading).

Is this useful enough to enough people to warrant a paragraph or shall I try again with just a tip?

### DIFF
--- a/sail.md
+++ b/sail.md
@@ -252,6 +252,32 @@ Finally, you may run your Dusk test suite by starting Sail and running the `dusk
 
     sail dusk
 
+
+<a name="laravel-dusk-debug"></a>
+### Visualizing Browser Tests
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
I dunno - this feels like more than I want to maintain 😅 , especially telling people to use VNC, etc.

Maybe this would work better as a blog post or something?

Patabugen:
haha, yeah I went back and forward on including the link to TightVNC a few times.

I'll revert to a Tip mostly pointing at the selenium-chrome-debug docs and mentioning --dusk

I really think it's helpful to include, but the Laravel docs are magic because of their balance of simplicity and links.

PR #7239, opened 2021-08-17
Update queues.md

mikevwoods:
artisan queue:work option max-time does not exist (did it ever?). I think it should be timeout.


taylorotwell:
It does seem to exist though? 😄

mikevwoods:
That's embarrassing, sorry. I've done the exercise (now!) and drilled down into the code and it's new in 8 from 7. Serves me right for being too lazy to upgrade from 7 to 8!

PR #7243, opened 2021-08-20
[8.x] Add composer post-update-cmd for local only Telescope installation

wsamoht:
Add an example on how to update Telescope when a Local Only Installation is being used.

If composer is not running in dev mode, then skip running the `telescope:publish` command.

[Stack Overflow source](https://stackoverflow.com/a/53022623/1129035)

### DIFF
--- a/telescope.md
+++ b/telescope.md
@@ -155,6 +155,15 @@ To keep the assets up-to-date and avoid issues in future updates, you may add th
         }
     }
 
+    // For Local Only Installation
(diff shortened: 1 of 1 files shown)


wsamoht:
I wonder if we need to add a Windows variant too using `%COMPOSER_DEV_MODE%`. I unfortunately cannot test that as I am on Linux.

taylorotwell:

PR #7244, opened 2021-08-21
[8.x] Update document withCount place snake case format

hongquan95:
- I spent almost half an hour to debug reason method withCount not run correctly. Because I used `{relation_count}` with relation in camelCase format. It seems that when calling alias it will be converted to snake case by the library. I find this [issue](https://github.com/laravel/framework/issues/18381) like my problem.
- [Reference](https://github.com/illuminate/database/blob/7ad5b2964080b5e683082ca6532affcb2e7b913a/Eloquent/Concerns/QueriesRelationships.php#L424)


taylorotwell:
Fixing it up on my end.

PR #7248, opened 2021-08-22
Describe container's `call` method

tontonsb:
This is a really useful tool whenever working with your own job-like things, e.g. actions.

### DIFF
--- a/container.md
+++ b/container.md
@@ -429,6 +429,28 @@ For example, you may type-hint a repository defined by your application in a con
         }
     }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Your example will actually throw an error. I have documented this now. Thanks.

tontonsb:
You are right, my bad. I use the atSign notation myself, but noticed that the method handles arrays and decided to document the more "modern" notation without trying out how it actually works.

Thanks for the great description!

PR #7250, opened 2021-08-23
Add return false on exception render method documentation

Juhlinus:
I found this very helpful when you have a specific condition that you need met, and otherwise a having a default Http response returned.


Juhlinus:
Discovered that this happens due to a response not being returned. So no return at all will yield the same result.

PR #7251, opened 2021-08-23
[8.x] Casting to array is not necessary

richardkeep:
(-_-)


taylorotwell:
But it shows you how multiple items are supported.

richardkeep:
Very true. my bad.

PR #7271, opened 2021-08-28
[8.x] Add the Query\Builder::implode() method

kevinb1989:


taylorotwell:
Hmm, didn't know this existed. I think I will hold on documenting this for now. I don't love documenting multiple ways to do basic things like implode an array.

PR #7274, opened 2021-08-31
Added: how to use unique ignore validation in Form Requests

PovilasKorop:
Not sure if it's the right way to phrase it, but I got this question quite a lot - how to use unique+ignore with Form Requests.

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -1420,6 +1420,10 @@ Instead of passing the model key's value to the `ignore` method, you may also pa
 
     Rule::unique('users')->ignore($user)
 
+If you use this inside of [Form Request class](#form-request-validation), you may access the Controller's variables with adding `$this->`:
+
+    Rule::unique('users')->ignore($this->user)
+
 If your table uses a primary key column name other than `id`, you may specify the name of the column when calling the `ignore` method:
 
     Rule::unique('users')->ignore($user->id, 'user_id')


taylorotwell:
Are you imagining that the `user` is present as a route parameter on the defined route?

PovilasKorop:
@taylorotwell hmm good point, it's not a full example, what I meant is the method in controller is

```
public function update(UserRequest $request, User $user) {
  // ...
```

But then again, not sure if it's a 100% typical scenario, not everyone is using it this way.

So maybe my PR is not something that should be in the official docs, not sure.

taylorotwell:
Hmm - I actually would not expect that to work 😅 ... looking at the code I do not see how it would. 

I tried to recreate similar behavior and `$this->userService` is `null`:

image

image

taylorotwell:
In your case it probably works because `{user}` is in your route definition. In which case Request `__get` will fallback to route parameters if there is not an input field on the request with the matching name.

So, I think I just need to adjust the wording in your PR a bit.

taylorotwell:
Will look into how to demonstrate this better.

PovilasKorop:
@taylorotwell thanks, you're right, it's `Route::resource()` in my case, with route model binding, so it may be a typical case but totally not the only one. 

You can also maybe ask Mohamed to shoot a Youtube video on such validation scenarios, he's doing an awesome job on Youtube, so cool to watch.

PR #7280, opened 2021-09-01
Mention the use of `@else`

u01jmg3:


u01jmg3:
It's not obvious that `@else` can be used. Would it be better to adjust the code examples?

PR #7284, opened 2021-09-02
Recommend changing composer.json name after installation

LukeTowers:
See https://github.com/Roave/SecurityAdvisories/issues/81 for what sort of problems can arise if this change isn't made.

### DIFF
--- a/installation.md
+++ b/installation.md
@@ -162,6 +162,8 @@ If your computer already has PHP and Composer installed, you may create a new La
     cd example-app
 
     php artisan serve
+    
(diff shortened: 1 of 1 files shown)


taylorotwell:
This is not unique to just installing via Composer.

LukeTowers:
@driesvints where would this be better located? Perhaps https://laravel.com/docs/8.x#initial-configuration?

driesvints:
@LukeTowers I don't think Taylor wants to merge in anything at this time. Now that the issue with the advisory will be solved soon I think it's fine to leave things as is.

LukeTowers:
Fair enough. I'd personally still recommending doing it regardless of installation method but it's up to you guys on that front. For @wintercms we'll probably introduce a post-install hook or command to do some cleaning tasks like that after install, or perhaps just make it a part of our install process through our CLI tool. @bennothommo see above FYI

PR #7286, opened 2021-09-03
[8.x] Add Meilisearch upgrade note to Scout docs

driesvints:
Adding this note will inform users to review Meilisearch breaking changes when they update their Meilisearch binary.

This also contains a note about compatibility between meilisearch-php and the meilisearch binary.

See https://github.com/laravel/scout/issues/516

### DIFF
--- a/scout.md
+++ b/scout.md
@@ -4,6 +4,7 @@
 - [Installation](#installation)
     - [Driver Prerequisites](#driver-prerequisites)
     - [Queueing](#queueing)
+- [Upgrading Scout](#upgrading-scout)
 - [Configuration](#configuration)
     - [Configuring Model Indexes](#configuring-model-indexes)
     - [Configuring Searchable Data](#configuring-searchable-data)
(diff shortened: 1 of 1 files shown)


georgeboot:
Perhaps we should also include something about compatibility between the `meilisearch-php` and the server? For example, composer installs latest `meilisearch-php` but `brew install meilisearch` on MacOS will install 0.20.0, which will now break.

driesvints:
Isn't scout now compatible with all Meilisearch versions? I thought your PR with the version check did just that.

georgeboot:
The whole thing is that if you have meilisearch server 0.21 and meilisearch-php 0.19, you are fine. Scout handles that. This is also the correct setup.

However, a (new) user might simple install the latest versions of everything. This will mostly go correct, but for example on Homebrew, 0.21 isn't available, and now you are getting a super strange error.

Users need to install the correct sdk version for the server version they have.

Probably super niche, so not worth a note in the docs, but felt I should at least share so you can make a decision on it.

driesvints:
@georgeboot that's a good remark, thanks for noting that. Can you maybe send in a PR to my PR for a proposal? Or just comment here how you'd note it?

georgeboot:
I would add the following in the docs under the driver prerequisites section, directly under the composer install for meili:

> Make sure you install a version of `meilisearch/meilisearch-php` that is compatible with your server version. You can check compatibility [here](https://github.com/meilisearch/meilisearch-php#-compatibility-with-meilisearch).

driesvints:
Thanks @georgeboot. I added that albeit a bit lower.

PR #7287, opened 2021-09-03
Change condition query example code

bradycharron:
The conditional query example code includes a second param for the callback, but this param does not appear to be used.  Instead it uses the `$value` used for the original condition.

I found this confusing when I tried to implement a conditional query with an `isset` condition, and then used the variable from the `isset` check as the second param.  It was not performing as expected because the callback is using the `$value` not the provided variable.

The proposed change shifts the callback's second param to a `use` statement for the callback, making sure that the variable is unchanged when used in the callback.

### DIFF
--- a/queries.md
+++ b/queries.md
@@ -776,7 +776,7 @@ Sometimes you may want certain query clauses to apply to a query based on anothe
     $role = $request->input('role');
 
     $users = DB::table('users')
-                    ->when($role, function ($query, $role) {
+                    ->when($role, function ($query) use ($role) {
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think the current documentation is fine. We are just showing the what you pass as the first argument to the `when` method is passed as the second value to the callback.

bradycharron:
@taylorotwell Thanks for clarifying.  The way you explained it makes sense.  I just expected the variable to be passed as the second param on the callback because they same, but that does not appear to be the case.

PR #7288, opened 2021-09-03
Fix method doc for return value

Khazl:
The subscribe method in the example returns an array and is not void.


ovillafuerte94 (approved):
is correct

PR #7290, opened 2021-09-04
Missing Form Request redirect options

RobGordijn:
I miss the docs on the options (`$redirect`, `$redirectRoute`, `$redirectAction`) for redirection in case the validation failed.
I think these few lines will add value, although I'm not 100% about the wording/order.

Feedback is appreciated.


taylorotwell:
I have documented this for you.

PR #7292, opened 2021-09-06
Horizon Notifications " long wait " clarity

jaiwalker:
I found that add the extra line in the documentation of notification would be more clear on how "long wait" is determined.

found this  issue referring to this https://github.com/laravel/horizon/issues/373

Feedback is appreciated.

### DIFF
--- a/horizon.md
+++ b/horizon.md
@@ -328,7 +328,9 @@ If you would like to be notified when one of your queues has a long wait time, y
 <a name="configuring-notification-wait-time-thresholds"></a>
 #### Configuring Notification Wait Time Thresholds
 
-You may configure how many seconds are considered a "long wait" within your application's `config/horizon.php` configuration file. The `waits` configuration option within this file allows you to control the long wait threshold for each connection / queue combination:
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hmm - not sure this brings any additional clarity.

GavG:
I think this clarity would actually be useful.

It took me finding the related issue (https://github.com/laravel/horizon/issues/373) to figure it out.

The current docs don't make it clear what the "wait" is. Initially, I thought it was the time it would take for the next job to be processed, so the wait time of the next job in the queue only.

As it happens, it's, apparently, an estimate of "_the time it takes to process all the jobs on the queue_", which is a bit less intuitive as not all jobs may take the same amount of time to process.

I also asked a couple of colleagues and neither knew what the LongWaitDetected event actually meant, they both assumed it had to do with cumulative backoffs or retries, which I suppose it does in an indirect sense.

siarheipashkevich:
@jaiwalker @GavG agree with you!

@taylorotwell please think about it

PR #7305, opened 2021-09-12
Change minimum php version to php7.4

mwakaambrose:
8.x depends on league/commonmark 2.0.2 which requires php7.4+. However the upgrade documentation still shows php7.3  as the required version.

Screenshot 2021-09-12 at 12 19 20

Ofcause you can still ignore this failure using `--ignore-platform-reqs` flag of composer.


Jubeki:
Laravel supports both CommonMark ^1.3 and ^2.0
https://github.com/laravel/framework/blob/240bf68c252a2cc21a68658f92a3356c3b073559/composer.json#L25

Also Laravel 8 supports PHP 7.3:
https://github.com/laravel/framework/blob/240bf68c252a2cc21a68658f92a3356c3b073559/composer.json#L18

You probably did run `composer update` on another PHP Version than your production one.
See also https://github.com/laravel/framework/issues/38747#issuecomment-916992990

PR #7306, opened 2021-09-12
[9.x] Update from Swift Mailer to Symfony Mailer

Jubeki:
Laravel 9.x will use the Symfony Mailer instead of the discontinued SwiftMailer.
Waiting for https://github.com/laravel/framework/pull/38481 to be merged.

Changes:
- Guzzle is no longer needed because symfony-http will be used by the API Transports instead.
- Each Transport now has now a symfony transport install (except array and log)

Unchanged:
- Config entries

Some things which should be discussed:
- Laravel 9.x will use Symfony 6.0 therefor the Links are absolut to https://symfony.com/doc/6.0/mailer.html there also exists a link to the current documentation if that is wished for: https://symfony.com/doc/current/mailer.html (Currently v5.3)
- Symfony Mailer is based on Symfony Mime which could be mentioned somewhere if necessary.

### DIFF
--- a/mail.md
+++ b/mail.md
@@ -427,6 +427,8 @@ If you already have a raw image data string you wish to embed into an email temp
 
 The `withSymfonyMessage` method of the `Mailable` base class allows you to register a closure which will be invoked with the Symfony Message instance before sending the message. This gives you an opportunity to deeply customize the message before it is delivered:
 
+    use Illuminate\Mail\Message;
+    
     /**
      * Build the message.
      *
@@ -436,7 +438,7 @@ The `withSymfonyMessage` method of the `Mailable` base class allows you to regis
     {
         $this->view('emails.orders.shipped');
 
-        $this->withSymfonyMessage(function ($message) {
+        $this->withSymfonyMessage(function (Message $message) {
             $message->getHeaders()->addTextHeader(
                 'Custom-Header', 'Header Value'
             );
--- a/mail.md
+++ b/mail.md
@@ -27,7 +27,7 @@
 <a name="introduction"></a>
 ## Introduction
 
-Sending email doesn't have to be complicated. Laravel provides a clean, simple email API powered by using the popular [Symfony Mailer](https://symfony.com/doc/6.0/mailer.html) component. Laravel and Symfony Mailer provide drivers for sending email via SMTP, Mailgun, Postmark, Amazon SES, and `sendmail`, allowing you to quickly get started sending mail through a local or cloud based service of your choice.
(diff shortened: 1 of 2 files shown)


driesvints:
Thank you @Jubeki

mail.md
```diff
@@ -434,7 +436,7 @@ The `withSwiftMessage` method of the `Mailable` base class allows you to registe
     {
         $this->view('emails.orders.shipped');
 
-        $this->withSwiftMessage(function ($message) {
+        $this->withSymfonyMessage(function ($message) {
```
driesvints, line 439: It might be prudent to show the type of the message here so people know which object it's referring to.

Jubeki: Do you mean something like this?
```suggestion
        $this->withSymfonyMessage(function (Email $message) {
```

driesvints: Yeah and import the namespace through a `use` statement.

Jubeki: I checked in Symfony-mailer branch and what is passed to the Callback is `Illuminate\Mail\Message` which wraps around Symfony Email.

Jubeki: I know find the name `withSymfonyMessage` a little bit confusing.

driesvints: @jubeki are you sure? I just checked as well and it's the email instance. The whole idea about this method is to directly interact with the email instance.

Jubeki: Oh it seems I kind of looked at the wrong callback. https://github.com/laravel/framework/blob/c8d16545df39de602d61998bdeb9326521e0bd9c/src/Illuminate/Mail/Mailer.php#L261-L268

mail.md
```diff
@@ -434,7 +438,7 @@ The `withSwiftMessage` method of the `Mailable` base class allows you to registe
     {
         $this->view('emails.orders.shipped');
 
-        $this->withSwiftMessage(function ($message) {
+        $this->withSymfonyMessage(function (Email $message) {
             $message->getHeaders()->addTextHeader(
```
taylorotwell, line 442: Does this syntax still work for adding headers to a Symfony Email message? We will want to update it accordingly if not.

Jubeki: addTextHeader still exists in Headers: https://github.com/symfony/mime/blob/dc5de94574db31fbb51d6b7d22e9d34daa2e305d/Header/Headers.php#L119-L125

and Headers is return on getHeaders() in Message (Email is extended from that):
https://github.com/symfony/mime/blob/dc5de94574db31fbb51d6b7d22e9d34daa2e305d/Message.php#L67-L70

PR #7325, opened 2021-09-29
[8.x] Adds documentation for `LazilyRefreshDatabase`

lukeraymonddowning:
This PR adds documentation for https://github.com/laravel/framework/pull/38861


taylorotwell:
We are debating simply adding this as a default in the base test case in which case we would not need to document it in this way.

PR #7327, opened 2021-09-30
join two commands

tal7aouy:
join the two npm commands  ```npm install ```  and  ```npm run dev```  in a command ```npm install && npm run dev```


taylorotwell:
Lots of unrelated change here.

PR #7328, opened 2021-10-01
Add a warning about direct uploads to S3

hdimitrov1:
S3 uploading of large files without specifying a mime type can cause you to run out of memory.

This has proven very hard to diagnose so I figured it's worth adding to the docs 😅


taylorotwell:
Holding off on this for now.

PR #7329, opened 2021-10-01
add docs regarding paginate parameters for multiple paginators one page

arborrow:


arborrow:
There may be a more Laravel way of expressing this but my use case scenario was for a customer relationship management (CRM) system. The contacts had multiple touchpoints, registrations, and donations and I wanted to avoid the scroll of death by paginating the results and I wanted multiple paginators on the contact.show blade. The optional parameters that were not mentioned in the docs are covered nicely in this video https://www.youtube.com/watch?v=aYYNmVnYfcs which revealed the method with the various parameters:

    public function paginate($perPage = null, $columns = ['*'], $pageName = 'page', $page = null)

My thinking is that it would be nice for the documentation to at least reference these optional parameters.

PR #7335, opened 2021-10-05
Add docs for `@aware` directive

calebporzio:
These are docs for the following PR: https://github.com/laravel/framework/pull/39100

### DIFF
--- a/blade.md
+++ b/blade.md
@@ -1037,6 +1037,38 @@ You may specify which attributes should be considered data variables using the `
 Given the component definition above, we may render the component like so:
 
     <x-alert type="error" :message="$message" class="mb-4"/>
+    
+<a name="accessing-parent-data"></a>
+#### Accessing Parent Data
+
+Sometimes you may want to access data from a parent component inside a child component. In these cases, you can use the `@aware` directive to make outside data available inside the component.
+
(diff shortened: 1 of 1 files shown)


calebporzio:
I struggled with the right code example for this. I landed on the simples example that wasn't overly contrived. The only tradeoff being that in a production app using purgeCSS, because we are dynamically generating the `text-` and `bg-` classes, they wouldn't get picked up.

I think the tradeoff here is worth it for demonstrating the technique.

blade.md
```diff
@@ -1037,6 +1037,38 @@ You may specify which attributes should be considered data variables using the `
...
+
+    <ul {{ $attributes->merge(['class' => 'bg-'.$color.'-200']) }}>
+        {{ $slot }}
+    </ul>
+
+Because the `$color` prop was only passed into the parent `<x-menu>`, it won't be available inside `<x-menu.item>`. However, if we use the `@aware` directive, we can make it available inside `<x-menu.item>` as well:
+
+    <!-- /resources/views/components/menu/item.blade.php -->
+
+    @aware(['color'])
```
inxilpro, line 1067: You need to set the default in the parent and the child because `@props` doesn't get evaluated before the child has finished rendering. This may be something we want to consider before merging…

```suggestion
    @aware(['color' => 'gray'])
```

PR #7345, opened 2021-10-08
Minor fix of the unique() index

jadamec:


taylorotwell:
I'm not sure there is an error here? Regardless no description on this PR so not sure what you're going for.

PR #7346, opened 2021-10-10
Fix documentation for collections `duplicate` methods

kohenkatz:
The documentation for Eloquent Collections implies that all base Collections methods work on Eloquent Collections too. As noted in https://github.com/laravel/framework/issues/34416, this is not actually true for the `duplicates()` and `duplicatesStrict()` methods.

This PR adds notes to those two methods to indicate that they perform differently for Eloquent Collections.

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -625,10 +625,16 @@ If the collection contains arrays or objects, you can pass the key of the attrib
 
     // [2 => 'Developer']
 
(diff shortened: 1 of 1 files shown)


Putr:
I just got caught by this problem. That would not happen if the base collection duplicate() definition would also have a warning that it works differently than the Eloquent version. Never contributed to the docs before, but I'm willing to do it, but would ask for some guidance.

PR #7356, opened 2021-10-14
whenFilled and whenHas do not accept a second closure

dougmiller:
The docs mention that a second closure can be passed to the two mentioned functions, but the signatures do not bear this out: `public function whenFilled($key, callable $callback)` and `public function whenHas($key, callable $callback)`

Indeed even testing this, the second closure is never executed:
`
$request->whenFilled('archived', function() use (&$filteredStudies) {
    $filteredStudies->where('date_archived', '!=',null);
}, function () use (&$filteredStudies) {
    $filteredStudies->where('date_archived', '=',null);     <---- never executed
});
`

So either to docs need to be updated (this request) or the functions need to be updated.

### DIFF
--- a/requests.md
+++ b/requests.md
@@ -331,14 +331,6 @@ The `whenHas` method will execute the given closure if a value is present on the
         //
     });
 
-A second closure may be passed to the `whenHas` method that will be executed if the specified value is not present on the request:
-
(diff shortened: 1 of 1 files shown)


dougmiller:
Yep, I dint realise I was behind. Cheers.

PR #7373, opened 2021-10-19
[8.x] Mention `touches` limits

creekorful:
The wording may not be perfect, I'm eager for feedback!

See: https://github.com/laravel/framework/issues/39259 for more information about the issue.


taylorotwell:
I don't ever do back to back notes. Also this is just a general database performance issue.

creekorful:
> I don't ever do back to back notes. Also this is just a general database performance issue.

Hi @taylorotwell,

Can we discuss this issue a bit more please? It's not only matter of performance issue. The whole parent `save()` call will fails if there are too many children in the relationship (in our case >127,268). See the linked issue for more details.

PR #7378, opened 2021-10-20
add details about bus closures

macbookandrew:


taylorotwell:
Your example will throw an error because no argument was passed to the CallQueuedClosure class. Did you try it?

PR #7384, opened 2021-10-21
fix/add missing "return"

Enaah:
- a `scope`-function should return an `\Illuminate\Database\Eloquent\Builder`-instance
- adds a missing `return` on the `scopeActive()`-function


taylorotwell:
A scope doesn't have to return a builder instance.

PR #7404, opened 2021-10-31
Add information on cache busting chunked assets

Fijvect:
Please consider adding some language to help those needing to cache-bust chunked assets.


taylorotwell:
Maybe it would be better to submit this to the Mix docs?

https://github.com/laravel-mix/laravel-mix/tree/master/docs

PR #7405, opened 2021-10-31
fix minor typo regarding foreignIdFor

miclaus:
`BIGINT` instead of `BIG INT`


miclaus:
@taylorotwell did i miss something? 😅

PR #7406, opened 2021-11-01
Correct note upsert

l3aro:
When digging deeper to `upsert` func, I find that `MySqlGrammar` has no use with params `uniqueBy`, but `SqlServerGrammar` do, as image below. After some investigate around google, I think this is a mistake in the documentation, not the source code.

image
image


taylorotwell:
Deleting the whole comment because I'm tired of PRs from people changing it back and forth. 😅

PR #7407, opened 2021-11-01
[8.x] Fix error with Stripe Checkout

ryanito:
I was getting the following error when copying from the example in the docs:

```php
Stripe\Exception\AuthenticationException
No API key provided. (HINT: set your API key using "Stripe::setApiKey(<API-KEY>)". You can generate API keys from the Stripe web interface. See https://stripe.com/api for details, or email support@stripe.com if you have any questions.
```

Adding Stripe::setApiKey(<API-KEY>) fixes this error.

### DIFF
--- a/billing.md
+++ b/billing.md
@@ -1617,7 +1617,6 @@ When defining your `success_url` checkout option, you may instruct Stripe to add
     use Illuminate\Http\Request;
     use Stripe\Checkout\Session;
     use Stripe\Customer;
-    use Stripe\Stripe;
 
     Route::get('/product-checkout', function (Request $request) {
         return $request->user()->checkout(['price_tshirt' => 1], [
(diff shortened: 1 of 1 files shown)


driesvints:
The PR originally sent in (https://github.com/laravel/docs/pull/7240) was incorrect in that you'd need the checkout session to retrieve the user. The example shows the stripe checkout session uses an authed user to instantiate the stripe checkout session so the user is still authed when you're redirected back with the session token. 

I've updated the example to only retrieve the session object with the passed along token. You always need to either use `Cashier::stripe()` or `$user->stripe()` with the SDK if you want to call the Stripe API so the Stripe secret that's being set by Cashier is passed along. `Stripe::setApiKey` is the old deprecated way of doing things and should be avoided.

Also see https://laravel.com/docs/8.x/billing#stripe-sdk

PR #7408, opened 2021-11-02
Update APP_SERVICE on image name change

iwasherefirst2:
This is important, because if APP_SERVICE is not changed accordingly, 
then any sail command will throw `ERROR: No such service: laravel.test'`.

This has been mentioned also in forums like https://stackoverflow.com/questions/65982603/laravel-cli-sail-artisan-returns-no-such-service-after-updating-service-nam or https://laracasts.com/discuss/channels/laravel/sail-error-no-such-service-laraveldottest

### DIFF
--- a/sail.md
+++ b/sail.md
@@ -396,7 +396,13 @@ Since Sail is just Docker, you are free to customize nearly everything about it.
 sail artisan sail:publish
 ```
 
-After running this command, the Dockerfiles and other configuration files used by Laravel Sail will be placed within a `docker` directory in your application's root directory. After customizing your Sail installation, you may wish to change the image name for the `laravel.test` service in your application's `docker-compose.yml` file. After doing so, rebuild your application's containers using the `build` command. Assigning a unique name to the `laravel.test` service image is particularly important if you are using Sail to develop multiple Laravel applications on a single machine:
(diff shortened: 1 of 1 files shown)


taylorotwell:
Why is the new image name appended to `laravel.test`?

iwasherefirst2:
I thought it may help the user to understand that "laravel.test.your-image-name" is the new image name that used to be  "laravel.test". But every other name works well too I guess. Shall I rename it from `APP_SERVICE=laravel.test.your-image-name` to `APP_SERVICE=your-image-name` ?

taylorotwell:
I don't know? Does it need to be a domain that ends in `.test`?

iwasherefirst2:
You can choose whatever image name you like. It does not need to be a domain nor does it need to end on `.test`.

taylorotwell:
I find this entire PR kinda confusing :/ ... 

I can change the image name of a service in my `docker-compose.yml` without needing to change the "service" name under the `services` array. Changing the image name will keep my image for the given application from colliding with other images on my machine. I don't understand why you state I also need to change `APP_SERVICE`?

Can you go into more detail? I just tried a test project and was able to change the image name and leave the rest of the things untouched and was able to use Sail perfectly fine and issue commands to it without problems.

iwasherefirst2:
Your right. I mixed it up, and I guess the people in the forum aswell. I actually over read that one should change the image name and not the service name.. One can change the service name aswell, but then one should adjust `APP_SERVICE..`. Sorry for the disturbance.

PR #7410, opened 2021-11-03
[8.x] Document custom factory and model discovery

phanan:
Just realized that the current documentation for model factories doesn't mention the ability to implement custom model and factory discovery logic with `Factory::guessModelNamesUsing()` and `Factory::guessFactoryNamesUsing()`. Unless there's a specific reason I'm not aware of, this ability is pretty powerful IMHO, especially if we're having a non-traditional app structure (for example, using [Laravel Modules](https://github.com/nWidart/laravel-modules)).

Let me know if there's anything I should fix, or if I should simply close this PR off 😄.

### DIFF
--- a/database-testing.md
+++ b/database-testing.md
@@ -138,6 +138,33 @@ Next, define a `model` property on the corresponding factory:
         protected $model = Flight::class;
     }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I kind of thing that is an overly complex feature for most users. I won't be removing it from the framework though in case power users or special packages need access to it.

PR #7411, opened 2021-11-03
Having a better and easier example for the "hasOneThrough" relationship.

kachelle:
The current example with the car & its owner in combination with a mecanic is very confusing.

So a better real life example could be: an _employee_, which belongs to a _team_, which belongs to a _company_.

Make use of these tables:
- **employees**
id - integer
name - string
team_id - integer
- **teams**
id - integer
name - string
company_id - integer
-  **companies**
id - integer
name - string

To access the `$company` of an `$employee`, you can now do `$employee->company`


taylorotwell:
This is basically the same example using different words, so I'm not sure it's that much simpler to understand. I find the previous example equally easy to understand.

kachelle:
Yes, but this is a working example.

In the existing example the "words" say: 

> each **Mechanic** model may be associated with one Car model, and each Car model may be associated with one **Owner** model

But the database says:
```
mechanics
    id - integer
    name - string

cars
    id - integer
    model - string
    mechanic_id - integer

owners
    id - integer
    name - string
    car_id - integer
```

Which is in words:
> each **Owner** model may be associated with one Car model, and each Car model may be associated with one **Mechanic** model

So the example used:
```php
namespace App\Models;

use Illuminate\Database\Eloquent\Model;

class Mechanic extends Model
{
    /**
     * Get the car's owner.
     */
    public function carOwner()
    {
        return $this->hasOneThrough(Owner::class, Car::class);
    }
}
```
Is not working with the given database.

That's why I suggested an alternative. With a clearer none dubious example.


Btw: love the framework.

PR #7425, opened 2021-11-11
[8.x] Added declined and declined_if validation rule documentation

Frozire:
Adds declined and declined_at validation rules to the documentation

Should only be merged if feature PR (https://github.com/laravel/framework/pull/39579) is merged.

Laravel PR: https://github.com/laravel/laravel/pull/5723
Framework PR: https://github.com/laravel/framework/pull/39579

> We are running a financial application where users answer yes, and no selectors, such as are you registered as a bad payer or if do you have existing debt. We could solve our issue by reversing the question's phrasing, e.g. "not_registered_as_a_bad_payer", and using accepted but this reduces intent.
> 
> At the same time, we allow financial instances to select their configurations of what they allow or disallow. This rule could help us align validation closer to the original intent and limit developer or interpretation errors when used to compare each input to each business's configuration; simply by avoiding a "semi-double negative".

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -769,6 +769,8 @@ Below is a list of all available validation rules and their function:
 [Date](#rule-date)
 [Date Equals](#rule-date-equals)
 [Date Format](#rule-date-format)
+[Declined](#rule-declined)
+[Declined If](#rule-declined-if)
 [Different](#rule-different)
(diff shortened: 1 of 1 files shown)


taylorotwell:
I will revisit this on Tuesday (release day).

PR #7426, opened 2021-11-11
Use arrow function in Basic Routing Hello World example

beaubus:
Since minimal PHP requirement is "php": "^8.0", would be cool to jettison Basic Routing Hello World greeting into the future with the arrow function. 🚀


taylorotwell:
I would prefer to show a full closure here. Thanks!

PR #7427, opened 2021-11-11
Update Homestead default port listing

georanma:
With  v12.4.0, Homestead no longer exposes ports for the various services that are offered. The Laravel docs do not currently reflect the change. Updated the ports and moved the previously mapped ports to a list of services you may want to forward.


svpernova09:
Thanks!

PR #7428, opened 2021-11-11
Prenoun correction

dali-rajab:
The prenoun refers to "user"


taylorotwell:
`them` is a gender neutral pronoun here.

PR #7440, opened 2021-11-18
Update sanctum.md

iotronlab:
https://laravel.com/docs/8.x/sanctum#how-it-works-spa-authentication
Need doc update for using guard other than web 
include in sanctum config
'guard' => 'custom'


taylorotwell:
I will have to work on rewording this.

PR #7441, opened 2021-11-18
Add more details on how to register RateLimiter

khlilturki97:
Hi,

I have recently tried to use the RateLimiter and followed the documentation. Somehow, i could not pass my RateLimiter to the throttle middleware. After few trials, i found out that my RateLimiter is not recognized by the application since it is not registered

I am suggesting these changes to prevent others from facing the same issue

Regards


taylorotwell:
Already included in a default Laravel application so this would be irrelevant to most I think:

image

khlilturki97:
From where did you take the screenshot ?

PR #7444, opened 2021-11-18
Update eloquent relationships.md

mostafa-rouzbahani:
In the belongsTo method, the parent term is misleading. It is not clear that it is the relation parent, not the PHP class parent.


taylorotwell:
I don't think it's very confusing.

PR #7446, opened 2021-11-18
Replace request() with $request

lupanvi:
Hi, when the view for resetting the password is returned in the boot method, it is sending the
$request variable as parameter, so for consistency should be better use
it instead of the global helper request()


taylorotwell:
Different endpoint.

PR #7449, opened 2021-11-20
[8.x] Change to openswoole

mohprilaksono:
https://www.swoole.co.uk/

### DIFF
--- a/octane.md
+++ b/octane.md
@@ -99,7 +99,7 @@ chmod +x ./rr
 If you plan to use the Swoole application server to serve your Laravel Octane application, you must install the Swoole PHP extension. Typically, this can be done via PECL:
 
 ```bash
-pecl install swoole
+pecl install openswoole
 ```
 
 <a name="swoole-via-laravel-sail"></a>


taylorotwell:
I have no idea what all this open Swoole vs Swoole stuff is about tbh. Why should I choose one or the other?

mohprilaksono:
@taylorotwell actually openswoole is a fork of the previous swoole extension since version v4.7.1, you can read about it here: 
https://www.getrevue.co/profile/openswoole/issues/we-are-now-open-swoole-812526

taylorotwell:
Why should I use this instead of swoole?

mohprilaksono:
The official documentation said that only support openswoole and only suitable for swoole extension version 4.7.1. also, in the installation process is using openswoole instead of swoole.
And i see Laravel octane recentely support both swoole and openswoole extension, 

https://github.com/laravel/octane/pull/407

But, maybe i should change this PR to support both? Like 'pecl install swoole or pecl install openswoole'?

PR #7457, opened 2021-11-23
[8.x] Trusted Proxies From Config File

rrpadilla:


taylorotwell:
I would need to revisit this and format it better. Is there a reason why you want to do this?

rrpadilla:
@taylorotwell This is very helpful when your local development is sail or laravel homestead but your production site is hosted on AWS behind a load balancer. The best use case for this is to get the values form a config file.

The current **TrustProxies** middleware is base on the [TrustedProxies package by Chris Fidao](https://github.com/fideloper/TrustedProxy) and that package is using a config file.

This is a nice guide for any developer that still have to use the config file.

You can read more about this [here](https://github.com/laravel/framework/pull/38295#issuecomment-975881124).

Thanks.

mtb3:
For what it's worth, I agree that this could be a useful addition to the documentation. In particular, for developers coming from older / existing Laravel applications (that are / were using the previous external TrustedProxies package), it can be initially confusing why a config file is not used.

PR #7465, opened 2021-11-25
Update Octane's sail installation instructions

aj-norman:
This is an edit for Swoole's installation under Sail.

A step is missed under "Swoole Via Laravel Sail" which requires the user to change the port configuration from `80` to `8000` for the `laravel.test` service. Without this step, accessing http://localhost returns an "Unable to connect" screen.


taylorotwell:
Not the correct fix - I will fix it.

PR #7468, opened 2021-11-26
[8.x] Sail Multiple Services Selections

ahmadelmalah:
Multiple selections is allowed already
This instruction is to tell users how to add multiple services on docker compose file


taylorotwell:
It prompts you.

ahmadelmalah:
@taylorotwell It just says `Which services would you like to install?` doesn't tell you how to multiselect

PR #7472, opened 2021-11-30
[8.x] Correcting the Billing canceled method name

rogerarruda:
The method name to check if the subscription is canceled is incorrect, the correct one must be `cancelled()` according to the file `Laravel\Cashier\Subscription` line 296 and 307.

->notCanceled() to ->notCancelled() `Laravel\Cashier\Subscription` line 318.


taylorotwell:
Both are supported.

PR #7473, opened 2021-11-30
[8.x] Remove duplicated part

MrMohamedAbdallah:
That part is almost identical to the tip box that's before that section.


taylorotwell:
I don't treat docs like code - it is OK to duplicate things to drive a point home. Thanks!

PR #7483, opened 2021-12-03
[8.x] Add note about different DB connections with whereHasMorph

driesvints:
`whereHasMorph` doesn't seems compatible with multiple database connections on models. I'm not entirely sure if this is a bug or a limitation. I tried to dig into the (Eloquent) query builder but couldn't figure out how to fix this.

It's around here where the constraints get applied: https://github.com/laravel/framework/blob/8eb1827d7290dba9c55e54fa296275eb6f4b491a/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php#L229-L230

So I opted to put it in the documentation for now.

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -1299,6 +1299,8 @@ You may occasionally need to add query constraints based on the "type" of the re
             $query->where($column, 'like', 'code%');
         }
     )->get();
+    
+> {note} Unfortunately, different database connections for models aren't compatible with the `whereHasMorph` call.
 
 <a name="querying-all-morph-to-related-models"></a>
 #### Querying All Related Models


michaeldyrynda:
Do any of the relationship methods work on models between database connections? I’ve always had issues historically with any query that wants to joint across connections.

driesvints:
@michaeldyrynda hmm good point. That could indeed also be it.

modernben:
@michaeldyrynda It works correctly on belongsTo relationships. I think since the whereHasMorph was added at a later point that might not have been built quite the same.

taylorotwell:
No relationships work like this I don't believe.

modernben:
Isn't that what `src/Illuminate/Database/Query/Builder->prependDatabaseNameIfCrossDatabaseQuery($query)` does?

michaeldyrynda:
That’s handling different databases but doesn’t cater to databases on different connections. 

Across connections, the configuration can have different users with different credentials and may exist on different database servers entirely. 

I’m not sure you could run queries like that in MySQL at all. Cross database on the same server, providing you have a user with access to both, should be no problem.

modernben:
@michaeldyrynda Is there a way to specify the same connection but a different database? Like a `protected $database` var or something. I'm just specifying a different connection to use a different database. All of the connection details are the same except for the database name.

michaeldyrynda:
That _should_ just work already if I’m not mistaken. The issue I had was that those databases were actually different servers. I don’t think this is a problem for the framework to solve (nor one that it can, I don’t think)

PR #7487, opened 2021-12-03
Added important environment variables for Sail Dusk testing

Andrewearls:
The DUSK_DRIVER_URL must target the selenium docker image which starts on port 4444. The chromium browser will access the local webapp on port 80 for testing. These ports can be changed, however these are the default ports that they are started on. Without these environment variables Dusk testing on Sail returns errors. 

https://stackoverflow.com/questions/65569147/laravel-sail-dusk-selenium-connection-refused

### DIFF
--- a/sail.md
+++ b/sail.md
@@ -257,6 +257,15 @@ depends_on:
     - redis
     - selenium
 ```
+Then, the following lines should be added to the '.env' file.
+
+```yaml
+APP_SERVICE="laravel.test"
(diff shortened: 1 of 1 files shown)


taylorotwell:
Following Sail docs on a fresh application works fine for me with default Sail:

CleanShot 2021-12-04 at 16 36 21@2x

I did have to use `seleniarm/standalone-chromium` image though on M1 MacBook.

PR #7489, opened 2021-12-05
Prevent lazy loading version requirement 

keizah7:
Tried to use this functionality, but it didin't work. Later, we realised that this was added in v8.44.0.
It would be nice to have version requirement information in docs

https://github.com/laravel/framework/pull/37480


taylorotwell:
Docs should always be considered to reflect latest Laravel version. I'm not cluttering up the docs with these types of things.

PR #7495, opened 2021-12-08
duplicate from ### Updating Tables

dmitriilopotovskii:


taylorotwell:
Docs aren't code. They can repeat things. They don't have to be DRY.

PR #7511, opened 2021-12-13
Reference updating/adding APP_SERVICE key in .env file

DougThwaites:
If a user customises their docker-compose service name without an APP_SERVICE key set sail will fail to run commands on the correct service.


taylorotwell:
The service name is not the same concept as the image name.

taylorotwell:
The docs do not say to change the service name - they say to change the **image** name.

PR #7514, opened 2021-12-14
Reference updating/adding APP_SERVICE key in .env file

DougThwaites:
Reopens #7511
References: laravel/sail#256

Opening a new PR as my wording/understanding was off in #7511.

Add reference to adding/updating an APP_SERVICE key in a users .env file after changing the laravel.test service image name.

The scenario behind this addition to the docs is a developer wanting a custom domain for there Sail project so they change the service image name in the docker-compose.yml file but then commands like `sail artisan` etc aren't proxied through to the correct application container and fail silently.

Thanks for your time team, appreciate the work everyone does.

### DIFF
--- a/sail.md
+++ b/sail.md
@@ -412,7 +412,7 @@ Since Sail is just Docker, you are free to customize nearly everything about it.
 sail artisan sail:publish
 ```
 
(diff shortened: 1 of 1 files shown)


DougThwaites:
Sorry for wasting your time here Taylor, I obviously need to brush up on my Docker terminology/skills.

Thoughts on..

> remember to update your applications .env file to include an APP_SERVICE key with a value of your customized name to proxy commands through to the correct application container then:

taylorotwell:
I don't know what to change about that entire section. It is only talking about changing an **image name** - it never instructs you to change the name of the `laravel.test` service in any way.

taylorotwell:
If you are trying to access multiple applications at once - you should proxy them on different ports in your `docker-compose.yml`.

This entire section is only about having a unique **image** for each application.

taylorotwell:
I removed any reference to `laravel.test` from that entire section. It isn't relevant.

DougThwaites:
Cheers man, new wording looks thanks for spending time with me on this.

Merry Xmas from NZ

PR #7524, opened 2021-12-17
Adjustment: Improve Instructions in Notifications

ajmeireles:
This PR increments the instructions necessary for understanding custom channel namespaces for notifications. This is because `make:channel` by default creates the channel with the namespace `Broadcasting`, when the recommended for custom channels is to define the namespace as `Channels` for better organization of the application.


taylorotwell:
I just adjusted the example namespace.

PR #7530, opened 2021-12-22
Change canceled() to cancelled();

brainlet-ali:
As canceled() is not defined in base library.


taylorotwell:
Both are defined.

CleanShot 2021-12-22 at 10 46 38@2x

PR #7531, opened 2021-12-23
[8.x] Mention package for custom claims

driesvints:
A long outstanding issue on the Passport repo [is the ability to add custom claims to issued tokens](https://github.com/laravel/passport/issues/94). However, atm league/oauth2-server currently does not provide an easy way to do this. 

Until an easy way has been provided so we can add this natively to Passport, I recommend mentioning the l[aravel-passport-claims](https://github.com/corbosman/laravel-passport-claims) package. This will allow people to add their custom claims while we don't need to perform any changes right now in Passport.

The reason why I don't want to go down the path of adding these changes to Passport is because I don't want to diverge too far from the way league/oauth2-server works. It's best to wait until they provide the functionality for us to implement this easily.

### DIFF
--- a/passport.md
+++ b/passport.md
@@ -16,6 +16,7 @@
     - [Refreshing Tokens](#refreshing-tokens)
     - [Revoking Tokens](#revoking-tokens)
     - [Purging Tokens](#purging-tokens)
+    - [Custom Token Claims](#custom-token-claims)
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't usually recommend or endorse third party packages in the docs unless they are very well known and maintained for a while.

fredsal:
https://github.com/thephpleague/oauth2-server/pull/1122

PR #7535, opened 2021-12-27
Correct meaning of sentence reversed

HiroKws:
The contents section is specified to remain used classes in html/js files. So maybe meaning reversed.

However, I'm not an English native, please correct/purge if I'm wrong.

### DIFF
--- a/mix.md
+++ b/mix.md
@@ -94,7 +94,7 @@ Your application's `webpack.mix.js` file is your entry point for all asset compi
 
     npx tailwindcss init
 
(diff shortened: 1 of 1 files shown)


HiroKws:
Thx correct it.

I'll point out one more.

I checked on the naked Tailwind CSS and laravel mix version, tailwind will purge non-used class definitions every time. Not only the production environment, but also development. If no purge, the generated CSS file will be huge.

So, I think need more adjust.

PR #7537, opened 2021-12-27
Fix doctype for $response in Middleware::terminate()

larsnystrom:
Since $response can be either \Illuminate\Http\Response or \Illuminate\Http\JsonResponse, the doctype should be the base class they both inherit from.


taylorotwell:
It's fine.

PR #7538, opened 2021-12-28
Change to make full theme tailwind config file.

HiroKws:
Mix used tailwind v3.
And `tailwind init` will generate an almost empty setting file. It will work well normally.
However, when `@apply` directive is used own CSS definition, it makes an "undefined" error for tailwind default classes. (Maybe it is a bug of tailwind.)
So, the full-defined-theme tailwind config file is safe to use. It is easy to generate.


taylorotwell:
Discussed this with @adamwathan - he said not to merge for now.

PR #7539, opened 2021-12-28
Fix `middlewarePriority` docblock

julesjanssen:
Make sure the PHPDoc type matches the current Foundation value.

### DIFF
--- a/middleware.md
+++ b/middleware.md
@@ -226,7 +226,7 @@ Rarely, you may need your middleware to execute in a specific order but not have
      *
      * This forces non-global middleware to always be in the given order.
      *
-     * @var array
+     * @var string[]
      */
     protected $middlewarePriority = [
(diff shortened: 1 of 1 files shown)


middleware.md
```diff
@@ -226,7 +226,7 @@ Rarely, you may need your middleware to execute in a specific order but not have
      *
      * This forces non-global middleware to always be in the given order.
      *
-     * @var array
+     * @var string[]
```
michaeldyrynda, line 229:
```suggestion
     * @var class-string[]
```

If you want to be explicit.

julesjanssen: I agree, but I just wanted to match it with [the line in the kernel](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Http/Kernel.php#L71).

PR #7543, opened 2022-01-01
Update npm command for laravel breeze

devinnived:
TailwindCSS [recommends](https://tailwindcss.com/docs/guides/laravel) using ```npm run watch``` to generate styles on demand with the JIT engine.

### DIFF
--- a/starter-kits.md
+++ b/starter-kits.md
@@ -1,11 +1,13 @@
 # Starter Kits
 
-- [Introduction](#introduction)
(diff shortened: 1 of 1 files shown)


devinnived:
Apologies for the unintended formatting of the Table of Contents. It was done automatically and I did not catch it. I've fixed it to be inline with the ```8.x``` branch.

taylorotwell:
Still unrelated changes here. You can use `dev` or `watch`. It just depends on what you are wanting to do.

PR #7545, opened 2022-01-02
Update eloquent-collections.md

james0r:
semantic correction. The collection is iterated upon.


taylorotwell:
I think we are referring to the PHP `Iterator` contract.

james0r:
Wouldn't it be capitalized as lowercase implies a general concept in programming and could be confused.

james0r:
or not

PR #7546, opened 2022-01-02
[8.x] chore(http-tests): added missing response assertion APIs

dammy001:


dammy001:
@taylorotwell should i add a section that list the available authentication assertions just like response assertions?

taylorotwell:
Removed the credential examples as we would need an example of what the actual array structure should be. Also removed similarJson example because it doesn't explain what it actually means or does.

dammy001:
@taylorotwell Alright.

PR #7547, opened 2022-01-03
Document `clickAndWaitForReload()`

u01jmg3:
Document [`clickAndWaitForReload()`](https://github.com/laravel/dusk/pull/953)

I chose not to keep the `waitForReload()` example as I couldn't think of a scenario where it wouldn't be combined with a preceding click event.


SjorsO:
I just tried it, the current example doesn't work very well:

```php
// This seems to only work if your application is slow
$browser->click('.some-action')
    ->waitForReload()

// This works every time
$browser->waitForReload(function (Browser $browser) {
    $browser->click('.some-action');
}); 

```

PR #7554, opened 2022-01-06
added missing return value of update method

thatal:
added missing return value of update method into master branch.


taylorotwell:
The docs will be merged forward by us.

PR #7556, opened 2022-01-06
clarify how to define a logging channel for PHP deprecation warnings

smknstd:
I created a channel named `deprecations` so my config looked like:

```
return [

    'deprecations' => env('LOG_DEPRECATIONS_CHANNEL', 'null'),

    'channels' => [
        ...
        'deprecations' => [
            'driver' => 'single',
            'path' => storage_path('logs/php-deprecation-warnings.log'),
        ],
    ],
]
```

But it was logging even if I didn't specify a LOG_DEPRECATIONS_CHANNEL in .env. I finally understood that the name I chose for my channel did overwrite and it made it log warnings anyway.

Not easy to write but I hope this PR make it a bit clearer...

### DIFF
--- a/logging.md
+++ b/logging.md
@@ -94,6 +94,19 @@ PHP, Laravel, and other libraries often notify their users that some of their fe
 
     'deprecations' => env('LOG_DEPRECATIONS_CHANNEL', 'null'),
 
+    'channels' => [
+        ...
+    ]
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
Thanks for contributing to Laravel! ❤️

PR #7559, opened 2022-01-07
[8.x] Document removePaymentMethod for Cashier Stripe

driesvints:


vlakoff:
The place where you insert this paragraph brings confusion. See the paragraph below the one you inserted: the paragraph for `$user->deletePaymentMethods('sepa_debit')` mentions the paragraph for `$user->deletePaymentMethods()`.

Therefore, you should insert the new paragraph _below_ the paragraph for `$user->deletePaymentMethods('sepa_debit')`.

driesvints:
Thanks. Adjusted it so it's placed above it right below the other way to delete a specific payment method.

taylorotwell:
So what is the difference between `deleteMethodMethods('sepa_debit')` and `removePaymentMethod('pm_visa')`? Is there a reason one starts with `remove` and one starts with `delete`?

driesvints:
@taylorotwell hah, I contemplated to first send in a PR to cashier to consolidate the terminology before I sent in this one. I'll do that as well then.

`deleteMethodMethods('sepa_debit')` deletes all sepa debit payment methods.
`removePaymentMethod('pm_visa')` deletes a specific payment method by its ID.

In essence, a payment method is never truly deleted but rather detached from a customer. It continues to live on in Stripe if it's still needed for later payments in some edge cases.

driesvints:
Here you go: https://github.com/laravel/cashier-stripe/pull/1298

driesvints:
Updated this PR as well

taylorotwell:
Marking as draft until Cashier release that includes new syntax.

PR #7562, opened 2022-01-07
Explained that "public urls" are still accessible during maintenance …

titonova:
…mode

Not sure if that's a bug lol, or if there's any reason why that would be wanted to be avoided. But just wanted to add that

Btw @taylorotwell , if you are reading this, I just wanted to ask whether there's a specific reason why there are no "Next" and "Previous" pages in the laravel docs. It makes reading it a bit stressful, having to scroll all the way to the top to go to the next page

If you are ok with it, I'd gladly contribute into adding them to the docs. Thanks

### DIFF
--- a/configuration.md
+++ b/configuration.md
@@ -140,6 +140,7 @@ You may also provide a `retry` option to the `down` command, which will be set a
 <a name="bypassing-maintenance-mode"></a>
 #### Bypassing Maintenance Mode
 
+
 Even while in maintenance mode, you may use the `secret` option to specify a maintenance mode bypass token:
 
     php artisan down --secret="1630542a-246b-4b66-afa1-dd72a4c43515"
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not sure this is particularly relevant? Why is this worth documenting? 🤔

Scrolling to the top of a website is built into your keyboard / OS already.

titonova:
@taylorotwell alright, lemme give a practical example, that actually happed to me in real life. I wanted to create a high-performance live search functionality, but I wanted it to be outside Laravel's framework  because bootstrapping the entire framework for each search request is slow( but I  couldn't use an open source search engine like elasticsearch or meillisearch, or a concurrency framework like swoole, because they are not supported on shared hosting servers), so I decided to build it outside  the framework at `/public/search.php`

When I downed the application for maintenance, I didn't  expect any search request to be coming in from users, thus adding new records to my DB and logs. It would have been really helpful to have known that routes outside laravel framework's route could still be accessed when the app is down.

Concerning the "next" and "previous" buttons, please at least tell me why putting it there isn't a good idea? I mean all other docs of  frameworks in the TALL stack have it, [tailwind,](https://tailwindcss.com/) [alpine](https://alpinejs.dev/), [Livewire](https://laravel-livewire.com/), like I said, if you accept my suggestion I'd love to contribute it to the docs. 

Just my 2 cents
Keep up the good work, Taylor !

PR #7564, opened 2022-01-09
Upgrade info on new model factories is missing

killerog:
I've been searching for ages why my factories wouldn't work anymore after upgrading. Please test the upgrade docs in a real-life situation.

There might be a better way or place to document this, but I don't see an option to ask a question or raise an issue.


driesvints:
@killerog draft PR's aren't reviewed. Please mark this as ready once you're ready with this PR.

PR #7565, opened 2022-01-09
Writing mail to log doesn't work anymore

killerog:
Something about the config makes it try to send an actual mail instead of delivering it to the logs.


driesvints:
@killerog draft PR's aren't reviewed. Please mark this as ready once you're ready with this PR.

PR #7569, opened 2022-01-10
[8.x] Pagination docs are incomplete when using Tailwind CSS

yoeriboven:
I couldn't get the Tailwind pagination to render properly.

Tailwind (when not using a CDN) only checks the `resources` folder for classes to generate. That's why you have to add the pagination views directory to `tailwind.config.js`. Another fix is publishing the files to your `resources` folder.

I created [a repo](https://github.com/yoeriboven/paginate) to test the issue. It also has some more info.


taylorotwell:
It probably makes more sense to add this to the Laravel Breeze / Jetstream tailwind.config.js files that are scaffolded in.

taylorotwell:
Ah - I see they are already there.

yoeriboven:
What do you think about adding ‘tailwind.config.js’ to the Laravel boilerplate? That way you’d have it also when you’re not using Breeze or Jetstream.

PR #7572, opened 2022-01-12
Avoid confusion over RefreshDatabase

spebon:
Clarify that the `Illuminate\Foundation\Testing\RefreshDatabase` trait does not restore the database prior to any tests but instead only prevents the current test from making permanent changes. Additionally, add information on how to achieve a complete database reset for a test.

I am not the only one confused by this, see [this](https://laracasts.com/discuss/channels/testing/refreshdatabase-trait-doesnt-refresh-database) thread.

### DIFF
--- a/database-testing.md
+++ b/database-testing.md
@@ -1,7 +1,7 @@
 # Database Testing
 
 - [Introduction](#introduction)
-    - [Resetting The Database After Each Test](#resetting-the-database-after-each-test)
+    - [Resetting The Database After Each Test](#resetting-the-database-after-a-test)
 - [Defining Model Factories](#defining-model-factories)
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not quite following - provide an example of something you do in a previous test that affects later tests? Like altering the schema?

spebon:
I have a test with RefreshDatabase querying an index method to see if the entries are sorted correctly by expecting a static JSON response. Inserting in the same table during a test run earlier without the RefreshDatabase trait will make the index test return an additional row (because the previous test's changes are not rolled back), thus breaking it.

taylorotwell:
Added this on master branch - thanks.

PR #7575, opened 2022-01-13
Fix misnamed validation rule.

neyaoz:
It is actually "mac_address", not "mac".


taylorotwell:
I'll fix it - the table of contents isn't updated here.

PR #7579, opened 2022-01-13
Update sanctum.md

Null78:
I spent three hours debugging this issue, that just required me to decode `%3D` to `=` in the token :)

and i think it's common because i got the same error with postman, and JavaScript fetch (copy paste manually)
so a small clarification will help a lot


taylorotwell:
Will need better formatting, capitalization, etc.

PR #7583, opened 2022-01-14
Documents whenContains method for Stringable

yoeriboven:
New `when...` string helpers were documented in [this commit](https://github.com/laravel/docs/commit/9406cb60ddca3e322118a47442f48c4e89d7f95f).

`whenContains` was added to the index but not documented.

Also fixes indentation (other `when...` examples are all on one line).


taylorotwell:
CleanShot 2022-01-14 at 13 40 37@2x

PR #7588, opened 2022-01-16
Improve documentation of phpredis usage

TheLevti:
Added documentation about how to use phpredis's compression and serialization options. Also improved the documentation of other phpredis connection parameters.

### DIFF
--- a/redis.md
+++ b/redis.md
@@ -162,19 +162,50 @@ By default, Laravel will use the phpredis extension to communicate with Redis. T
         // Rest of Redis configuration...
     ],
 
-In addition to the default `host`, `port`, `database`, and `password` server configuration options, phpredis supports the following additional connection parameters: `name`, `persistent`, `prefix`, `read_timeout`, `retry_interval`, `timeout`, and `context`. You may add any of these options to your Redis server configuration in the `config/database.php` configuration file:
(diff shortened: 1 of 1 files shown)


taylorotwell:
@TheLevti why does the serialization and compression stuff have to be nested within an `options` key? I am not seeing that in the code or am I missing something?

TheLevti:
> @TheLevti why does the serialization and compression stuff have to be nested within an `options` key? I am not seeing that in the code or am I missing something?

```php
    /**
     * Create a new clustered PhpRedis connection.
     *
     * @param  array  $config
     * @param  array  $options
     * @return \Illuminate\Redis\Connections\PhpRedisConnection
     */
    public function connect(array $config, array $options)
    {
        $connector = function () use ($config, $options) {
            return $this->createClient(array_merge(
                $config, $options, Arr::pull($config, 'options', [])
            ));
        };

        return new PhpRedisConnection($connector(), $connector, $config);
    }
```

The whole config set up for phpredis in single and cluster mode is very confusing and messy. I would like to improve that in a future MR. Currently specific params can be placed in so many different places and its hard to understand where to put them. For example on the cluster config the failover behaviour is set globally although this should be per cluster connection configurable.

taylorotwell:
I've documented this now. Thanks.

PR #7596, opened 2022-01-19
Added early return example for after validation hook.

Sammyjo20:
The "after" validation hook in the FormRequest class is really useful, however, we thought that it would run only after the rest of the validation rules have passed. I have added a helpful if statement which should clear up any confusion in the future.


taylorotwell:
Hmm - I'm not sure if I want to stick this in this section. It feels a bit out of place. Will think about it.

PR #7602, opened 2022-01-19
Add hint to call order in database-testing.md

patriziotomato:


taylorotwell:
Unclear what this is trying to inform the user of.

patriziotomato:
I was having a situation where the callback itself did a SELECT statement to see how many models already have been created to increment a counter. Thats why realized that if you create 3 models using the factories, the process order is:

Callback for model 1 > callback for model 2 > callback for model 3 > creation of model 1 > creation of model 2 > creation of model 3

instead of what I've expected:

Callback for model 1 > creation of model 1> callback for model 2 > creation of model 2 > callback for model 3  > creation of model 3

PR #7605, opened 2022-01-20
[9.x] Adds note on `Implicit Route Bindings With Enums`

nunomaduro:
This pull request as the note that Implicit Route Bindings With Enums is only available for PHP 8.1+.


taylorotwell:
We mention the version just below:

CleanShot 2022-01-20 at 09 51 52@2x

PR #7609, opened 2022-01-21
[master] Minor html fixes

driade:
Fix minor html issues which leads to fatal errors when validating the html


driade:
I'd suggest to move the images on releases.md to laravel.com/images

Also, it'd be somehow easy to make a github action/validation process which converts the files to html and validates them looking for fatal errors.

PR #7616, opened 2022-01-22
[8.x] Added note to upsert method about unique index

aneeskhan47:
Hi,

I was using the `upsert` method provided by eloquent and it was still inserting rows. I checked through google etc and found out that you must have a `primary or unique` index on uniqueBy columns but this isn't documented in the docs hence I spent hours thinking why it is inserting new rows instead of updating.

this pr adds it and I checked laravel database grammar's classes and `compileUpsert()` method have `uniqueBy` argument in all of them.


taylorotwell:
I'm pretty sure this used to be in the documentation but was removed?

aneeskhan47:
> I'm pretty sure this used to be in the documentation but was removed?

Yes it was there...but idk why same previous pr's were closed as well as this

PR #7622, opened 2022-01-23
[8.x] Added `Rule::when` (recent 8.x addition)

matiaslauriti:
I just saw a post on [StackOverflow](https://stackoverflow.com/questions/70822355/different-validation-rules-for-input-based-on-another-input-value-from-formreque) and someone answered it using `Rule::when` (I was totally unaware of it) and I think adding it to the documentation would be an awesome addition as it is a really simple way of adding or not rules based on a complex conditional (either on a `Validator` or inside a `FormRequest` but I just used the `Validator` example).

I am not 100% happy with the example I added to the documentation, so please, feel free to correct me and add a better example if you have one 👍

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -1600,6 +1600,8 @@ If you need to use a validation rule that already exists but you need to do add 
         'value' => Rule::when($request->type === 'numeric', ['required|numeric'], ['required|string']),
     ]);
 
+The `Rule::when` method can be used inside a Form Request's `rule` method.
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
This rule is not necessary in PHP 8+. You can accomplish the equivalent of the rule using pure PHP syntax.

comes:
How, @taylorotwell ?

PR #7633, opened 2022-01-27
[8.x] Add upsert unique index note

mnabialek:
Upsert at leat in MySQL add code like:

```
on duplicate key update `value` = values(`value`), `updated_at` = values(`updated_at`)
```

 to the query so in case no unique index is created, then new records are created instead of updating existing ones and it's not mentioned in documentation that unique index should be created on those columns


taylorotwell:
If we are adding this note back please add it back in its original form.

PR #7634, opened 2022-01-28
Add tip how to disable failed jobs database logging

CodeTechNL:


taylorotwell:
Not entirely accurate - I will fix it.

taylorotwell:
Had to make a slight change to the framework to make this all work a little better, actually.

CodeTechNL:
@taylorotwell so setting that config value to `null` doesn't stop registering failed jobs?

taylorotwell:
No I don't think so - it may just be triggering an error that is silent or in your logs.

CodeTechNL:
What is the good way to solve this? I don't want to log the failed jobs to the database, any idea for me?

PR #7642, opened 2022-01-31
[8.x] Document assertSimilarJson for http tests

WesleyNZ:
Added missing `assertSimilarJson` description


taylorotwell:
Everyone keeps trying to document this but the documentation never explains what "similar" actually means 😅

PR #7644, opened 2022-02-01
[9.x] Custom mail transports

driesvints:
Adds a section on how to write custom transports either through a package or in your own app. It seems `Mail::extend` was undocumented before this. Is the `register` the correct method for this? Or should it be `boot`? 🤔 

I tried my best with providing a realistic example. But haven't tried this one out manually. The code should be good (based on https://mailchimp.com/developer/transactional/guides/send-first-email/#send-your-first-email).

See https://github.com/laravel/framework/issues/40330

### DIFF
--- a/mail.md
+++ b/mail.md
@@ -24,6 +24,7 @@
 - [Testing Mailables](#testing-mailables)
 - [Mail & Local Development](#mail-and-local-development)
 - [Events](#events)
+- [Custom Transports](#custom-transports)
 
 <a name="introduction"></a>
 ## Introduction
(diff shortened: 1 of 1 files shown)


taylorotwell:
Should be `boot` method - I'll update it.

RoodFruit:
@taylorotwell First of all, thanks for the docs. However, this section focuses on a fully new custom mail transport. There are also some already defined and usable by requiring the necessary symfony package, for example `symfony/sendinblue-mailer`.

In case of Sendinblue, your docs is to extensive, since you only need to do the following in the `boot()` method, since the implementation is now already there:

```
use Illuminate\Mail\MailManager;
use Symfony\Component\Mailer\Bridge\Sendinblue\Transport\SendinblueTransportFactory;
use Symfony\Component\Mailer\Transport\Dsn;
...

$this->app[MailManager::class]->extend('sendinblue', function () {
    return (new SendinblueTransportFactory())->create(
        new Dsn('sendinblue+api', 'default', $this->app['config']->get('services.sendinblue.api'))
    );
});
```

My suggestion would be to expand the docs with an example of an existing symfony mailer, so people know.

driesvints:
@RoodFruit can you send in a PR?

RoodFruit:
@driesvints @taylorotwell PR: https://github.com/laravel/docs/pull/7647

PR #7645, opened 2022-02-01
[8.x] Document Ngrok account signup and auth token

driesvints:
Ngrok requires you to register an account first these days and set up an authentication token. By noting this in the docs we can warn users in advance about this.

Resolves https://github.com/laravel/valet/issues/1172


Aishwow:
من شأن ماذا التحذير

PR #7669, opened 2022-02-08
[9.x] Update releases.md related to event:list command

imanghafoori1:


taylorotwell:
I don't think the visual aspect of this command changed? Just some minor fixes?

imanghafoori1:
@taylorotwell Can not be considered as a minor fix, since it solves the good old problem of the "pub/sub" pattern which says that you can not easily find listeners of an event since they can lurk anywhere.

PR #7671, opened 2022-02-08
Fix typo

allayGerald:
- Fixed typo while mentioning _rendering inline blade components_ contributors


taylorotwell:
I don't see the problem. It's two separate features.

PR #7677, opened 2022-02-08
[9.x] Removed Forced Scoping Of Route Bindings from release notes

gpressutto5:
This is not new in Laravel 9, it exists since Laravel 8.


taylorotwell:
We are highlighting some very recent L8 features in these release notes as they may have slipped under people's radar.

PR #7709, opened 2022-02-13
Update docs for `@once` directive

apih:
Based on the source code, [src/Illuminate/View/Compilers/Concerns/CompilesConditionals.php#L291](https://github.com/laravel/framework/blob/7bcf724546cdd4cb929a3daa9a61b2e8574f549a/src/Illuminate/View/Compilers/Concerns/CompilesConditionals.php#L291), it seems that `@once` directive supports custom ID. Update the documentations to include this information.

### DIFF
--- a/blade.md
+++ b/blade.md
@@ -536,10 +536,29 @@ The `@once` directive allows you to define a portion of the template that will o
 @endonce
 ```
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
No need to include this I don't think. Shouldn't really be used by end users.

PR #7712, opened 2022-02-14
[9.x] Use Backed Enums For Enum Casting

lioneaglesolutions:
The documentation does not state that backed enums must be used for enum casting.

When using a standard enum for casting (not a backed enum), the error message is `Undefined property: App\Enums\ServerStatus::$value` which may be confusing for those who are just starting to use enums and are just following the docs. 

This PR adds a backed enum example for enum casting (on both the releases page and the mutators page). The enum naming convention used matches the naming used in the [Implicit Enum Binding] section (https://laravel.com/docs/9.x/routing#implicit-enum-binding) in the docs.


taylorotwell:
Fixed.

PR #7719, opened 2022-02-15
Document `json` as a supported castable value.

alexbowers:
`isJsonCastable@HasAttributes` has `json` in the list, but this isn't documented.


taylorotwell:
I think the idea is that you would actually just use `array` or `collection`.

PR #7722, opened 2022-02-18
[9.x] Document the extend() method in global scopes

lucasdotdev:
This PRs adds a paragraph about the use of the `extend()` method in global scopes to register query builder macros.

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -1049,7 +1049,7 @@ The `Scope` interface requires you to implement one method: `apply`. The `apply`
     use Illuminate\Database\Eloquent\Model;
     use Illuminate\Database\Eloquent\Scope;
 
-    class AncientScope implements Scope
+    class ExpiringScope implements Scope
     {
         /**
          * Apply the scope to a given Eloquent query builder.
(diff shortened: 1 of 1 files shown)


lucasdotdev:
Hey, @taylorotwell! I thought a lot about what you said (and even updated some of my own codes) and used the `SoftDeletingScope` as an inspiration for this example below. What do you think? 😅
```php
use Illuminate\Database\Eloquent\Scope;

class ExpiringScope implements Scope
{
    public function apply(Builder $builder, Model $model)
    {
        $builder->whereNull('expired_at');
    }

    public function extend(Builder $builder)
    {
        // Here we are supressing the scope, so the query will include expired models.
        $builder->macro('withExpired', function (Builder $builder) {
            return $builder->withoutGlobalScope($this);
        });

        // Here we are both supressing the scope and adding a whereNotNull clause,
        // so the query will return only expired models.
        $builder->macro('onlyExpired', function (Builder $builder) {
            return $builder->withoutGlobalScope($this)
                ->whereNotNull('expired_at');
        });
    }
}
```

PR #7723, opened 2022-02-18
[9.x] Remove redundant `return` key for `when` methods

siarheipashkevich:
https://github.com/laravel/scout/pull/592


ahmadmayahi:
It would be nice to use the arrow functions.

PR #7724, opened 2022-02-18
Always use ->query()

mbukovy:
Hey,

is there any particular reason, why an official documentation still suggests accessing Eloquent static methods through magic?
I think that instead of this
image

an official documentation should always recommend `->query()` version
image

This is a big advantage, since it doesn't use magic, IDE autocomplete always works without any plugins or helpers and it is a correct, explicit way of using these methods.

---

Or shouldn't a `->query()` variant be at least mentioned somewhere visible in the docs?

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -90,7 +90,7 @@ A one-to-one relationship is a very basic type of database relationship. For exa
 
 The first argument passed to the `hasOne` method is the name of the related model class. Once the relationship is defined, we may retrieve the related record using Eloquent's dynamic properties. Dynamic properties allow you to access relationship methods as if they were properties defined on the model:
 
-    $phone = User::find(1)->phone;
+    $phone = User::query()->find(1)->phone;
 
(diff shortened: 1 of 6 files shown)


eloquent-relationships.md
```diff
@@ -172,15 +172,15 @@ Once the relationship method has been defined, we can access the [collection](/d
...
+    $comments = Post::query()->find(1)->comments;
 
     foreach ($comments as $comment) {
         //
     }
 
 Since all relationships also serve as query builders, you may add further constraints to the relationship query by calling the `comments` method and continuing to chain conditions onto the query:
 
-    $comment = Post::find(1)->comments()
+    $comment = Post::query()->find(1)->comments()
```
siarheipashkevich, line 183: will be better to place each call of method on one line

siarheipashkevich:
Another possible solution for better DX it's add these method to phpdoc for `Model` class.

dammy001:
@siarheipashkevich i go with you.

siarheipashkevich:
@taylorotwell could you please explain why?

mbukovy:
> Or shouldn't a `->query()` variant be at least mentioned somewhere visible in the docs?

Any thoughts about this at least?

PR #7731, opened 2022-02-19
[9.x] Fix Rule::foreEach typo

pieter-irsan:
Changed `Rule::foreEach` to `Rule::forEach` in the [Validation](https://laravel.com/docs/9.x/validation#accessing-nested-array-data) docs.


cesarkohl:
All good here.

PR #7736, opened 2022-02-21
json and jsonb methods equivalent columns

mounir-tk:
json and jsonb methods creates longtext equivalent columns instead of JSON


taylorotwell:
On MySQL (and maybe others) they create JSON columns.

PR #7744, opened 2022-02-24
[9.x] Clarify with behavior on query builder

driesvints:
I keep seeing people trip over this so probably best to document this.


taylorotwell:
I'm going to move this to a different section.

PR #7746, opened 2022-02-24
Adding note about upsert to Eloquent page

dbpolito:
Moving the same note we have on Query Builder Upsert to Eloquent Upsert:

https://laravel.com/docs/9.x/queries#upserts

This took me a while to track down what was the issue 😅


acidjazz:
Nice find!

acidjazz:
it might be cool if the upsert method detected if the 2nd parameter was specified, and then threw an exception linking to this warning. save people some time.

PR #7747, opened 2022-02-25
Update scout.md

mmachatschek:
This adds a note to the queue section for laravel/scout, that updates to records is handled by the third party service and is asynchronously by itself.


driesvints (approved):

taylorotwell:
A bit confusing - I assume you are meaning that index updates are not immediately reflected if you were to perform a search directly after indexing?

mmachatschek:
@taylorotwell Exactly

PR #7750, opened 2022-02-25
Removed duplicated link entry

drradao:


taylorotwell:
This is not break existing links.

drradao:
When clicking the title it seems to add the wrong hash https://laravel.com/docs/8.x/routing#customizing-the-key

And the correct link is nested within the link.

I think one of them has to go away.

taylorotwell:
Thanks

PR #7751, opened 2022-02-26
fix(eloquent-mutators): use new object instead of facade

Carlwirkus:
Docs seem to be out of date with the new mutator syntax. see [release notes](https://laravel.com/docs/9.x/releases#eloquent-accessors-and-mutators)


taylorotwell:
The Attribute::make method is new since the release.

PR #7767, opened 2022-03-04
Update upgrade.md: add information about Symfony Request changes

IlyaVorozhbit:


taylorotwell:
This is not related to Laravel, and is not a method we document that people should use.

PR #7769, opened 2022-03-05
Grammar mistakes, trailing ending line fix

rizimore:
There are a few grammar mistakes and on some files there were two ends of lines, so to just keep consistency and follow standards, removed extra space from there.


taylorotwell:
Some of these fixes don't make sense.

rizimore:
Thanks for the reply, I just want to contribute, can you mention which did not make sense so I don't consider that next time.

PR #7773, opened 2022-03-06
[9.x] minor grammar/usage change: 'actions which' --> 'actions that'

rizimore:


enderandpeter:
I think I agree, based on what I'm understanding about defining vs. non-defining clauses, [thanks to Grammarly](https://www.grammarly.com/blog/which-vs-that/).

rizimore:
@enderandpeter thanks for confirmation

PR #7785, opened 2022-03-07
For compatibility reasons all table rows should have borders (pipe symbols) at the start and at the end

rizimore:


taylorotwell:
These tables aren't broken. I think it's enough PRs on this for now.

rizimore:
> These tables aren't broken. I think it's enough PRs on this for now.

Sure @taylorotwell, I just find so think about to fix all the places.

PR #7788, opened 2022-03-08
missing subtitle from overview

rizimore:
image

image


taylorotwell:
`####` level headings are never sublinked

PR #7790, opened 2022-03-09
document callAfterResolving method

tabacitu:
I was surprised to learn about `callAfterResolving()` from https://github.com/laravel/framework/pull/41397. And it looks like a lot of other popular package maintainers don't know about it either:
- https://github.com/spatie/laravel-permission
- https://github.com/tighten/ziggy
- https://github.com/404labfr/laravel-impersonate
- https://github.com/albertcht/invisible-recaptcha

Because some 3rd-party packages use `Blade::directive()` and others use `app()->afterResolving()`, a race condition is created. If we recommend all packages use `callAfterResolving()`, no more race condition.

So I think `callAfterResolving()` should be added to the docs. Not sure this is the proper form - let me know if it needs rephrasing or moving.

### DIFF
--- a/providers.md
+++ b/providers.md
@@ -146,6 +146,28 @@ You may type-hint dependencies for your service provider's `boot` method. The [s
             //
         });
     }
+    
+
+<a name="call-after-resolving"></a>
+#### Call After Resolving
+
+You can use the `callAfterResolving()` method, when working with other packages that provide Blade directives or Blade components - especially if those packages use `app()->afterResolving()`. This will ensure that your callback gets run after the Blade compiler has been registered:
+
+    use Illuminate\View\Compilers\BladeCompiler;
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't understand why `callAfterResolving` is necessary in your example. You're in the `boot` method. All of those packages were having trouble because of calling Blade::directive in `register` which they should not have been doing in the first place?

taylorotwell:
I'm fine with documenting `callAfterResolving`, but we will need to use an example that makes sense... it's not needed here because you're in `boot`. Note the `laravel-impersonate` package is registering Blade directives from the `register` method of a service provider. This is **not recommended**. They are hacking around it using `afterResolving`, but they should just move the registration to the `boot` method of the service provider as intended.

PR #7797, opened 2022-03-10
replace eloquent attribute from make to constructor

chrisk-7777:


chrisk-7777:
I believe `Attribute::make` is no longer available in v9

taylorotwell:
CleanShot 2022-03-10 at 09 14 14@2x

PR #7802, opened 2022-03-12
Document `withSymfonyMessage()` for notifications

u01jmg3:
Documentation for `withSymfonyMessage()` already exists in the [mail docs](https://laravel.com/docs/9.x/mail#customizing-the-symfony-message), however, it was missing for notifications.

### DIFF
--- a/notifications.md
+++ b/notifications.md
@@ -17,6 +17,7 @@
     - [Customizing The Templates](#customizing-the-templates)
     - [Attachments](#mail-attachments)
     - [Adding Tags & Metadata](#adding-tags-metadata)
+    - [Customizing The Symfony Message](#customizing-the-symfony-message)
     - [Using Mailables](#using-mailables)
     - [Previewing Mail Notifications](#previewing-mail-notifications)
 - [Markdown Mail Notifications](#markdown-mail-notifications)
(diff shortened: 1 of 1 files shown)


notifications.md
```diff
@@ -557,7 +581,7 @@ If needed, you may return a full [mailable object](/docs/{{version}}/mail) from
      * Get the mail representation of the notification.
      *
      * @param  mixed  $notifiable
-     * @return Mailable
+     * @return \Illuminate\Notifications\Messages\MailMessage
```
Jubeki, line 584: This should still be a Mailable.
Because in this Section it is about returning a Mailable instead of a MailMessage.
```suggestion
     * @return Mailable
```

notifications.md
```diff
@@ -577,7 +601,7 @@ If you are sending an [on-demand notification](#on-demand-notifications), the `$
...
```
Jubeki, line 604: This should still be a Mailable.
Because in this Section it is about returning a Mailable instead of a MailMessage.
```suggestion
     * @return Mailable
```

notifications.md
```diff
@@ -498,7 +499,7 @@ Unlike attaching files in mailable objects, you may not attach a file directly f
...
```
Jubeki, line 502: This should still be a Mailable.
Because in this Section it is about returning a Mailable instead of a MailMessage.
```suggestion
     * @return Mailable
```

u01jmg3:
Thanks @Jubeki

PR #7806, opened 2022-03-14
Specify Laravel's lock-system on jobs.

Husky110:
Recently I stumbled over a problem which has been issued here: [https://github.com/laravel/framework/issues/41464](https://github.com/laravel/framework/issues/41464)  
Since I don't have the time right now to issue a PR to the framework itself, maybe we can warn others to not make the same assumptions I did.

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -249,7 +249,7 @@ Furthermore, when a job is deserialized and model relationships are re-retrieved
 <a name="unique-jobs"></a>
 ### Unique Jobs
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I have added this to the docs with slightly different wording. Thanks!

Husky110:
Thanks for implementing the PR. That issue was driving me insane on sunday... 😄 I was wondering why you actually let the locks rest inside the Cache-system instead of the queue-system itself. It feels kind of weird to me since most of Laravel I've discovered so far seems to be intuitive by having all it's mayor components be kind of standalone... Like instruments in an orchestra - they can play together but do not have to. Having to use the cache-system in a specific way to get the queues to work correctly fells kinda weird, since they look like complete different subsystems to me, which should not depend on each other like that... Thanks for Laravel. :)

PR #7811, opened 2022-03-16
changing 'nexmo' => 'vonage', 'toNexmo' => 'toVonage', and 'NexmoMess…

petethewizard:
…age' to 'VonageMessage'

as the 'nexmo' variant does not work.


petethewizard:
I tested this in Laravel 8 and it only works when you change it to 'vonage' as in the pull request.

PR #7819, opened 2022-03-21
[9.x] Document message ID and `MessageSent` changes

driesvints:
I think it's important to inform people about the changes made to SES message ID's and the `MessageSent` event.

See https://github.com/laravel/framework/issues/41585


upgrade.md
```diff
@@ -550,6 +550,16 @@ Again, many applications may not be interacting with these methods, as they are
 
 SwiftMailer offered the ability to define a custom domain to include in generated Message IDs via the `mime.idgenerator.idright` configuration option. This is not supported by Symfony Mailer. Instead, Symfony Mailer will automatically generate a Message ID based on the sender.
 
+##### SES Message IDs
+
+In Laravel v8 and before, the SES Transport generated separate `X-Message-ID` and `X-SES-Message-ID` headers. In Laravel v9, these headers are removed. You should be able to safely rely on the generated header by Symfony Mailer.
```
taylorotwell, line 555: How do you access the generated header?

driesvints: Updated

massimodellarovere:
Now the problem is that MessageId is not same value the X-SES-Message-Id.. Message Id is create from Symphony with domain used.. But the Amazon call API SES return an value different that is important for check the notifications above status message as BOUNCED, DELIVERED etc.. Amazon not use MessageId generated Symphony.. In Laravel 8 MessageId e SesMessageid has same value. But Now en Laravel 9 MessageId has a value and SES Messageid tha is different now is not present..

**Before in Laravel 8 MessageID is same SES-MessageID**
```php
$message->getOriginalMessage()->getHeaders()->addTextHeader('X-Message-ID', $messageId);
$message->getOriginalMessage()->getHeaders()->addTextHeader('X-SES-Message-ID', $messageId);
```
Now in Laravel 9 Messageid is created by Mailer not by Amazon SES 
```php
$event->sent->getMessageId() // Is not X-SES-Message-ID returned from API SES
```
I would add specific header as in the previous version, while the value X-Message-ID is ok new Symphony/Mailer
/vendor/laravel/framework/src/Illuminate/Mail/Transport/SesTransport.php add this
```php
$message->getOriginalMessage()->getHeaders()->addTextHeader('X-SES-Message-ID', $messageId);
```

ankurk91:
https://github.com/laravel/framework/pull/41615

PR #7837, opened 2022-03-28
fix: missing "return" in code example

igor-kamil:
I guess this is missing (also for older versions)


nshiro:
@igor-kamil
It is intentional.
Document says 
```
Scopes should always return the same query builder instance or void:
```
This is the pull request for it. #7372

igor-kamil:
Now I see. Thank you for pointing this out 👍

PR #7843, opened 2022-03-29
Update database.md SQL Server Configuration

tomfortmuller:
Include additional parameters needed for encrypted SQL Server connections per [https://stackoverflow.com/questions/71493855/laravel-the-certificate-chain-was-issued-by-an-authority-that-is-not-trusted](https://stackoverflow.com/questions/71493855/laravel-the-certificate-chain-was-issued-by-an-authority-that-is-not-trusted)


taylorotwell:
I'll just stub these out in the skeleton.

PR #7850, opened 2022-04-03
[9.x] Document new isolate command

joelbutcher:
Resolves this [issue](https://github.com/laravel/valet/issues/1227) on `laravel/valet`


valet.md
```diff
@@ -208,6 +209,17 @@ Sometimes, you may wish to configure Valet to serve a "default" site instead of
...
+
+### Isolating Sites
+
+If you have multiple projects with varying PHP versions, you may wish to isolate each site to a specific PHP version. You can do this with the `isolate` command:
+
+```shell
+valet isolate php@8.0
+```
+
+> {tip} You can proxy calls to your projects underlying PHP binary using `valet php`. This also works with `valet composer` and `valet which-php`
```
joelbutcher, line 221:
```suggestion
> {tip} You can proxy calls to your projects underlying PHP binary using `valet php`. This also works with `valet composer` and `valet which-php`.
```

NasirNobin:
@joelbutcher I think PR is can include the details of `valet unisolate` command. I suppose we can show that in similar way `valet secure/unsecure` does. 

CleanShot 2022-04-04 at 11 38 22

NasirNobin:
Also, perhaps the `valet isolated` command (that lists all isolated sites with the PHP versions) and `valet which-php` can be listed in the [Other Valet Commands](https://laravel.com/docs/9.x/valet#other-valet-commands) section.

joelbutcher:
@NasirNobin Updated

PR #7853, opened 2022-04-04
[9.x] Documentation for eloquent resources macros and whenNotNull method

smokills:


taylorotwell:
Even though macros are supported - that isn't how I would handle this situation. I would use `stub:publish` to customize the resource stub to extend your own defined resource base class, and within that base class I would define any common helpers you need.

PR #7854, opened 2022-04-04
[9.x] Add note to restart computer after upgrading Valet

driesvints:
This might be necessary sometimes. See https://github.com/laravel/valet/issues/1228

### DIFF
--- a/valet.md
+++ b/valet.md
@@ -130,7 +130,7 @@ If you are having trouble getting your Valet installation to run properly, execu
 <a name="upgrading-valet"></a>
 ### Upgrading Valet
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
This feels kinda janky? Not sure I want to recommend this without an understanding of why it's recommended so we can explain it.

driesvints:
@taylorotwell we haven't figured out the exact reason for this since it's only happening on some machines (including mine) after upgrading to Valet v3. Rebooting fixes the certification issues. I feel like this isn't much to ask from people to upgrade.

Pinging @mattstauffer to help explain

mattstauffer:
I agree it's janky. We don't have a better solution yet but I will try to dig into it tonight; hopefully I can downgrade my valet then upgrade again and repro the issue.

driesvints:
Sent in https://github.com/laravel/valet/pull/1230 as an alternative.

PR #7865, opened 2022-04-08
add tabPrice for multiple items

Safemood:
add the `tabPrice` method to the documentation for multiple items per invoice.


taylorotwell:
If you call `tabPrice` multiple times, when is the customer actually charged?

Safemood:
when we call 

$user->invoice();

taylorotwell:
I have updated the docs.

PR #7870, opened 2022-04-10
revert anon class in first migration example

jordancjones:
the text after the first example migration states that the example should show the automatically assigned class name ie from running php artisan make:migration create_flights_table


jordancjones:
alternatively the text below this example could remove "As you may have noticed in the example above" so and the anonymous class could be left in

PR #7878, opened 2022-04-14
[9.x] Add note about custom tld's

driesvints:
After discussing with @mattstauffer we decided to add an explicit note about custom tld usage. Even though Valet ships with a `tld` command to set a custom one we don't document this. This is because it can potentially lead to unexpected results (see https://github.com/laravel/valet/issues/1231). We believe Valet was intended to be used with the shipped `.test` tld and using a custom tld should only be done as a last resort. However, if you do so we can't offer support because there's too much unknowns that can lead to broken usage.

### DIFF
--- a/valet.md
+++ b/valet.md
@@ -97,8 +97,6 @@ Once Valet is installed, try pinging any `*.test` domain on your terminal using 
 
 Valet will automatically start its required services each time your machine boots.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
@driesvints @mattstauffer instead of adding that note here where it probably won't be seen... I suggest adding the warning / confirmation text to the `tld` command itself in Valet so it is seen right when the command is executed. 👍

PR #7880, opened 2022-04-17
[9.x] Improve `groupBy` preserveKeys argument syntax

Aberdeener:
I *believe* this was the intended syntax here, using PHP 8.0 named arguments. Apologies if I am incorrect :p
Couldn't find any other places where this occurred.


taylorotwell:
This was likely written before named arguments existed and we were just showing you what the argument is.

PR #7883, opened 2022-04-19
[9.x] Fix description of Mailer return types

driesvints:
It seems this was mistakenly referencing "number of recipients" which wasn't returned in the first place. That happened on transport level, not mail level. It's of more interest to document the Mailer return types of the most common methods.

Related: https://github.com/laravel/framework/pull/42035


driesvints:
To be very honest, we can remove this entire section as it doesn't have any impact.

PR #7884, opened 2022-04-19
More details on mail's plain-text part

smknstd:


taylorotwell:
I'm not sure why we would need to include this?

PR #7893, opened 2022-04-21
[9.x] add Eloquent relationship cheat sheet

browner12:
the current docs are a spectacular resource for people newer to Eloquent relationships and how they work in depth.

however, for more experienced users, it can become a little difficult/annoying to dig through all this text to try and remember exactly what the relationship methods are, or what the methods are to attach or detach models in certain relationships.

having this at-a-glance table would be helpful to grok exactly what we're looking for.

I left out how you would call the attaching and detaching from the inverse method not because I wanted to, but more so because I was a little worried how wide this would display, and if it'd be easy to read.

If someone could confirm I actually wrote the correct methods down that would be appreciated, because honestly I went through this a little quick.

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -38,6 +38,7 @@
     - [Belongs To Relationships](#updating-belongs-to-relationships)
     - [Many To Many Relationships](#updating-many-to-many-relationships)
 - [Touching Parent Timestamps](#touching-parent-timestamps)
+- [Cheat Sheet](#cheat-sheet)
 
 <a name="introduction"></a>
 ## Introduction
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hmm - while I appreciate the sentiment behind this, I'm not sure I want to maintain it in the docs. It feels like a great project for like a "laravel-eloquent-cheatsheet.com" or something that would be a fun community project.

Orrr.... even a "cheatsheet.laravel.com" that we could do officially as a succinct, quick guide to Laravel's methods?

taylorotwell:
Hmm - while I appreciate the sentiment behind this, I'm not sure I want to maintain it in the docs. It feels like a great project for like a "laravel-eloquent-cheatsheet.com" or something that would be a fun community project.

Orrr.... even a "cheatsheet.laravel.com" that we could do officially as a succinct, quick guide to Laravel's methods?

browner12:
Yah, I'm open to other options. If there are other sections that would benefit from a more succinct explanation, the cheatsheet.laravel.com might be a nice solution.

PR #7900, opened 2022-04-26
Heading update

gauravmak:
Not a native but `Architectural Concepts` seems better than `Architecture Concepts`


taylorotwell:
I think it's OK.

PR #7902, opened 2022-04-26
Add a hint to also add an amount of $tries when using the backoff array in queues

Bassadin:
I tripped over this when trying to set an array of 5 backoff values in my project. I think when you set an array with 5 backoff values, Laravel should just default to a `$tries` amount of 5. But even if not, it should at least be mentioned in the docs, imho.

Keep on the great work, I love the framework even though I much disliked PHP before :)

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -1684,6 +1684,8 @@ You may easily configure "exponential" backoffs by returning an array of backoff
     {
         return [1, 5, 10];
     }
+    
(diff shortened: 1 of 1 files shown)


taylorotwell:
What happens if you don't do this?

taylorotwell:
I'm not sure this is really necessary here.

PR #7904, opened 2022-04-27
SFTP driver configuration remove duplicate key

jagrutimetaliya:
PHP array automatically removes the duplicate key, we have two 'password' => env('SFTP_PASSWORD') in SFTP driver file system configuration Example. So there is no need to define two similar keys.


taylorotwell:
it's fine. it's showing you the private key password and username password.

PR #7907, opened 2022-04-28
Clarify using toSearchableArray with the Scout database engine

keithbrink:
Using `toArray` when using the Scout database engine will return an empty array if it's not specifically defined on the model, and this empty array will mean that no columns are filtered.

This should be clarified in the documentation.


taylorotwell:
I don't feel like this message is super clear.

PR #7908, opened 2022-04-28
Fix upgrade.md "Deleting Missing Files"

lucaguindani:
Attempting to `delete` a file that does not exist returns `false`

https://github.com/laravel/framework/blob/1dd741f76d255fbc16edb0f350d6654a648ef102/src/Illuminate/Filesystem/FilesystemAdapter.php#L485


driesvints:
This isn't correct. Deleting both an existing and non-existing file will return true. The line you link to is only invoked on a file operation error.

PR #7910, opened 2022-04-29
Explain that serializeDate doesn't apply to timestamps

anthonymf:
+ As explained https://github.com/laravel/framework/issues/32335


taylorotwell:
Not sure this is very clear based on original issue.

PR #7913, opened 2022-04-29
Add Note About Relationship make() caveat

erichowey:
Per discussion on issue https://github.com/laravel/framework/issues/41313
This note explains the persist behavior of a factory on a relation.


taylorotwell:
Not sure this red note really belongs here.

erichowey:
> Not sure this red note really belongs here.

What would you suggest so folks are aware of this behavior?

VladyslavSikailo:
I also think we still need some warning regarding that implicit behaviour

PR #7918, opened 2022-05-04
Update logging.md to reference action_level

Dachande663:
Added information around the `action_level` logging parameter.


taylorotwell:
Doesn't show any example of how to set it up.

PR #7922, opened 2022-05-05
Add note to database-testing to mention reproducible tests

bert-w:
The code examples of database factories in testing are sprinkled with Faker functions, but it is worth mentioning that you should not test attributes that are different on each run of the test.


taylorotwell:
I don't think this is necessarily true.

PR #7923, opened 2022-05-05
success_url

nicolasbeauvais:
When creating a checkout session, `success_url` and `cancel_url` should be passed as the first parameter to the `checkout()` method, `$sessionOptions` instead of `$customerOptions`.

Reference: https://github.com/laravel/cashier-stripe/blob/13.x/src/Checkout.php#L63


driesvints:
This isn't correct. It's referencing this method: https://github.com/laravel/cashier-stripe/blob/13.x/src/Concerns/PerformsCharges.php#L140

PR #7931, opened 2022-05-12
Add APP_SERVICE env changes

mariowhowrites:
Add context for changing service name, and add `APP_SERVICE` env required change.

I changed some of the wording around for some of the preceding text to add more context around how to change the service name. if those changes aren't acceptable I can trim to just the `APP_SERVICE` required env change below the command.


taylorotwell:
As has been explained before, this section is ONLY talking about images. NOTHING about services.

https://github.com/laravel/docs/pull/7514

PR #7932, opened 2022-05-12
Clarify that csrf endpoint is only required for 3rd party and mobile apps

bernardwiesner:
This section is causing confusion to many so it would be good to clarify this is only needed for 3rd party SPA's and mobile apps.

### DIFF
--- a/csrf.md
+++ b/csrf.md
@@ -65,7 +65,7 @@ The `App\Http\Middleware\VerifyCsrfToken` [middleware](/docs/{{version}}/middlew
 <a name="csrf-tokens-and-spas"></a>
 ### CSRF Tokens & SPAs
 
-If you are building a SPA that is utilizing Laravel as an API backend, you should consult the [Laravel Sanctum documentation](/docs/{{version}}/sanctum) for information on authenticating with your API and protecting against CSRF vulnerabilities.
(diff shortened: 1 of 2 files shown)


taylorotwell:
Wording is unclear. What is third-party? What is first-party? A first-party SPA (in my mind) DOES need to call that endpoint.

bernardwiesner:
@taylorotwell 
Thanks for checking, I apoligize if my wordings were unclear. Based on the docs for setting up your SPA you need to enable the `EnsureFrontendRequestsAreStateful` middleware: https://laravel.com/docs/9.x/sanctum#sanctum-middleware.

If you set that middleware it means `XSRF-TOKEN` cookie the will be set automatically through the middleware: https://github.com/laravel/sanctum/blob/2.x/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php#L30-L31

So basically there is no need to call that endpoint for the SPA because the `XSRF-TOKEN` cookie is already set. There are quite a few question regarding this over the web such as:

https://stackoverflow.com/questions/61975282/laravel-sanctum-csrf-cookie-request-optional/64695186#64695186
https://stackoverflow.com/questions/72210505/should-we-make-a-request-to-sanctum-csrf-cookie-first-before-registration-that?noredirect=1#comment127597867_72210505

PR #7934, opened 2022-05-12
Document the usage of DB::raw() within the default value as to not be…

eduance:
As discussed with @driesvints within issue [laravel/framework#42336](https://github.com/laravel/framework/issues/42336) a documentation change to recommend against using raw query statements within the default() method.

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -927,7 +927,7 @@ The `default` modifier accepts a value or an `Illuminate\Database\Query\Expressi
         }
     };
 
-> {note} Support for default expressions depends on your database driver, database version, and the field type. Please refer to your database's documentation.
+> {note} Support for default expressions depends on your database driver, database version, and the field type. Please refer to your database's documentation. Be aware that is strongly recommended to not use `DB::raw()` for specifying the default value as Laravel utilizes doctrine/dbal to do migration changes and there is no guarantee for raw statements to be formatted correctly.
 
 <a name="column-order"></a>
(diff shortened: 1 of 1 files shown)


eduance:
@taylorotwell This is referring to the usage of raw statements using the `DB::raw()` within the default() within the migration scheme. When chaining the `change()` to the default, e.g:

```
Schema::table('example', function (Blueprint $table) {
    $table->uuid('uuid')->default(DB::raw('(replace(uuid(),\'-\',\'\'))'))->change();
});
```

The ChangeColumn class utilizes the database platform which essentually calls `alterTableSql()` which then tries to populate this raw statement without validating what's given in Laravel.

This addition to the documentation would explain that it is recommended to not use raw statements within the default method.

The issue is linked under where @driesvints has recommended going for a documentation notation.

@taylorotwell Otherwise it is possible to actually fix this issue, but this would mean to follow up on doctrines way which means to actually customize the platform. If this is preferred, I would be happy too.

https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/platforms.html

Essentually the only thing we'd have to do would be to check if the value is an instance of a query expression.

Otherwise this documentation change would be the easiest way of warning against using raw statements. Let me know what is preferred and I will make sure to get it done properly.

https://github.com/laravel/framework/issues/42336

driesvints:
@taylorotwell it's best that you resend this with that explanation.

PR #7937, opened 2022-05-14
[9.x] Adds refresh section.

DarkGhostHunter:
Pending: https://github.com/laravel/framework/pull/42342

### DIFF
--- a/cache.md
+++ b/cache.md
@@ -187,6 +187,44 @@ If you need to retrieve an item from the cache and then delete the item, you may
 
     $value = Cache::pull('key');
 
+<a name="retrieve-update"></a>
+#### Retrieve & Update
+
+With a cache driver supporting [Atomic Locks](#atomic-locks), you will be able to retrieve and update an item from the cache using the `refresh` method.
+
+This method accepts the cache key, a callback, and an optional lifetime to refresh. The callback receives the cached item, or `null` if it doesn't exist, and it should return the updated value.
+
+    $item = Cache::refresh('message', function ($message) {
+        $message ??= 'Hello';
+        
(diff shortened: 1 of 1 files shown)


cache.md
```diff
@@ -187,6 +187,48 @@ If you need to retrieve an item from the cache and then delete the item, you may
 
     $value = Cache::pull('key');
 
+<a name="retrieve-update"></a>
+#### Retrieve & Update
+
+If your driver supports [atomic locks](#atomic-locks), you will be able to retrieve and update an item without race conditions, avoiding any change on the value by other process while it's being manipulated.
```
dammy001, line 193:
```suggestion
If your driver supports [atomic locks](#atomic-locks), you will be able to retrieve and update an item without race conditions, avoiding any change in the value by other processes while it's being manipulated.
```

cache.md
```diff
@@ -187,6 +187,48 @@ If you need to retrieve an item from the cache and then delete the item, you may
...
+    	if ($mission->isAccomplished()) {
+		    $expire->now();
+	    }
+    	
+	    return $mission;
+    });
+
+> {tip} If you don't want to refresh the item and its lifetime, you can return `null`.
+
+When the lock cannot be acquired immediately, it waits for a maximum of 10 seconds. You may further configure the atomic lock by omiting the callback, which will return an object to alter the lock properties and waiting time.
```
dammy001, line 223:
```suggestion
When the lock cannot be acquired immediately, it waits for a maximum of 10 seconds. You may further configure the atomic lock by omitting the callback, which will return an object to alter the lock properties and waiting time.
```

PR #7941, opened 2022-05-16
Update starter-kits.md to mention Fortify

RicLeP:
Fortify can sometimes get lost in the documentation. I’ve seen many articles and stories mentioning the starter-kits and forum posts with people asking how to customise or remove X from the frontend. I myself have used a starter-kit only to remove the frontend in the past. Like how the Fortify docs reference the starter-kits, to starter-kits should reference Fortify as an alternative.


taylorotwell:
Eh, Fortify is pretty different from starter kits and requires a bit more advanced knowledge of Laravel. Not sure I want to recommend it here.

From your experience, what do people typically want to remove from the frontend?

PR #7944, opened 2022-05-18
Update octane.md

msonowal:
updated port to use `80` as by default as default sail `docker-compose.yml` internal port configured to `80` 
so for the sail users who is simply following the docs will be able to run as is.

You need to choose `no` as an answer otherwise it registers some grpc plugins and it will try to find the proto files thus resulting in error on startup.


msonowal:
Hi @taylorotwell  any reason behind this so that I understand detailed

PR #7951, opened 2022-05-24
replace first class callables with comments

timacdonald:
With [PHP first class callables](https://www.php.net/manual/en/functions.first_class_callable_syntax.php) now using this syntax, we should probably change the way we express "and you put your things in here" for function calls.

Suggesting we utilise `/* ... */`, which is a thing I've seen pretty commonly I feel.

### DIFF
--- a/broadcasting.md
+++ b/broadcasting.md
@@ -748,9 +748,9 @@ If you would like to listen for events on a private channel, use the `private` m
 
 ```js
 Echo.private(`orders.${this.order.id}`)
-    .listen(...)
-    .listen(...)
-    .listen(...);
+    .listen(/* ... */)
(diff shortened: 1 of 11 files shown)


broadcasting.md
```diff
@@ -748,9 +748,9 @@ If you would like to listen for events on a private channel, use the `private` m
 
 ```js
 Echo.private(`orders.${this.order.id}`)
-    .listen(...)
-    .listen(...)
-    .listen(...);
+    .listen(/* ... */)
+    .listen(/* ... */)
+    .listen(/* ... */);
```
timacdonald, line 753: These aren't first class callables, being JS, but I adjusted these to match the rest of the docs.

PR #7955, opened 2022-05-27
extended example for release lock by owner

agoalofalife:
I suppose the example seems clear if we define owner when attempting to lock.


taylorotwell:
We would likely need more explanation than this.

PR #7956, opened 2022-05-29
Use to_route() helper instead of redirect()->route()

PerryvanderMeer:
Since we can use to_route() to redirect to a route, it would be nice to use it in the docs.


taylorotwell:
We use `redirect()->route()` in other places. We would want to update them all at once if we are going to update them.

PR #7962, opened 2022-05-31
[9.x] Correct docs for the observer boot method

PixellUp:
This is a small PR that fixes the missing `static` in the `boot()` method according to https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Model.php#L250

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -1421,7 +1421,7 @@ To register an observer, you need to call the `observe` method on the model you 
      *
      * @return void
      */
-    public function boot()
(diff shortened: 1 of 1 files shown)


taylorotwell:
It's referring to the `boot` method of a **service provider**.

ArthurPatriot:
@taylorotwell sorry for tagging here) Because we haven't issues in this repository.

Can we add an up arrow to the documentation to scroll up the page?

PR #7963, opened 2022-06-02
Fix Log::withContext() middleware example

matical:
`$request->header()` does not set any values. It only retrieves.

https://github.com/laravel/framework/blob/ff5c52b7556a41901b961d9ceee4fee55da27b3a/src/Illuminate/Http/Concerns/InteractsWithInput.php#L45-L48

### DIFF
--- a/logging.md
+++ b/logging.md
@@ -233,7 +233,9 @@ Occasionally, you may wish to specify some contextual information that should be
                 'request-id' => $requestId
             ]);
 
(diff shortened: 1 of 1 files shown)


matical:
Oh, my bad. I assumed the original intention was to attach it to the `Request` so that it would be retrieved (internally) from a controller with `$request->header("Request-Id")`.

PR #7964, opened 2022-06-02
[9.x] Document new Cashier Paddle fake test methods

driesvints:
Documents https://github.com/laravel/cashier-paddle/pull/165. I'm not sure I'm 100% happy with the examples as they don't demonstrate real-life app situations. Although giving specific app examples might make things too complicated.


taylorotwell:
I personally think I will leave this a bit undocumented for now until we can firm it up a bit. Will just let us use it internally.

PR #7966, opened 2022-06-03
[9.x] Add validation response example

timacdonald:
This adds the validation response format with an example of a validation response containing

- A flat key
- A nested object key
- An array with nested objects

Which i feel gives a pretty good example of the response format in general.

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -8,6 +8,7 @@
     - [Displaying The Validation Errors](#quick-displaying-the-validation-errors)
     - [Repopulating Forms](#repopulating-forms)
     - [A Note On Optional Fields](#a-note-on-optional-fields)
+    - [Validation Error Response Format](#validation-error-response-format)
 - [Form Request Validation](#form-request-validation)
     - [Creating Form Requests](#creating-form-requests)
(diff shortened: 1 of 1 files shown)


validation.md
```diff
@@ -258,6 +259,30 @@ By default, Laravel includes the `TrimStrings` and `ConvertEmptyStringsToNull` m
 
 In this example, we are specifying that the `publish_at` field may be either `null` or a valid date representation. If the `nullable` modifier is not added to the rule definition, the validator would consider `null` an invalid date.
 
+<a name="validation-error-response-format"></a>
+### Validation Error Response Format
+
+When your application throws a `ValidationException` and the request is expecting a Json response, Laravel will automatically format the error messages for you and return a `422 Unprocessable Entity` status code. Below is an example of the response format. You will notice that the error keys are flattened into dot-notation format:
```
dammy001, line 265:
```suggestion
When your application throws a `ValidationException` and the request is expecting a JSON response, Laravel will automatically format the error messages for you and return a `422 Unprocessable Entity` status code. Below is an example of the response format. You will notice that the error keys are flattened into dot-notation format:
```

PR #7967, opened 2022-06-03
Order of arguments inconsistent

Vinze:
It seems like the example in the docs isn't inconsistent with the method signature. See: Illuminate/View/Compilers/BladeCompiler.php:633

I already noticed that Laravel swaps the arguments when they are in the wrong order, but it might be confusing because an IDE suggests `$class, $alias`. Whereas the docs say `$alias, $class`.


taylorotwell:
It works either way.

PR #7984, opened 2022-06-15
[9.x] Vite

jessarcher:
This PR includes documentation for Vite as the default asset compilation tool.


jessarcher:
> I don't think we should delete the Mix docs here. All back linking will be broken. Instead, just remove it from the navigation.

The page hasn't been deleted, just trimmed down a lot and moved in the navigation.

driesvints:
@jessarcher ah I see. No worries then!

vite.md
```diff
@@ -0,0 +1,371 @@
...
+npm run build
+```
+
+<a name="working-with-scripts"></a>
+## Working With JavaScript
+
+<a name="aliases"></a>
+### Aliases
+
+The Laravel plugin provides a common aliases to help you hit the ground running:
```
tontonsb, line 162: aliases -> alias

vite.md
```diff
@@ -0,0 +1,370 @@
...
+Within a fresh installation of Laravel, you will find a `package.json` file in the root of your application's directory structure. The default `package.json` file already includes everything you need to get started using Vite and the Laravel plugin. You may install your application's frontend dependencies by running:
+
+```sh
+npm install
+```
+
+<a name="configuring-vite"></a>
+### Configuring Vite
+
+Vite is configured in a `vite.config.js` file in the root of your project. You are free to customize this file based on your needs and you may also install any other plugins your application requires, such as `@vitejs/vue-plugin` or `@vitejs/react-plugin`.
```
mpociot, line 71: The plugins are `@vitejs/plugin-vue` and `@vitejs/plugin-react`

vite.md
timacdonald: Fixed. Thank you!

vite.md
timacdonald: Fixed. Thank you!

frontend.md
```diff
@@ -0,0 +1,190 @@
...
+         * @param  int  $id
+         * @return \Inertia\Response
+         */
+        public function show($id)
+        {
+            return Inertia::render('Users/Profile', [
+                'user' => User::findOrFail($id)
+            ]);
+        }
+    }
```
jessarcher, line 141:
```suggestion
```php
<?php

namespace App\Http\Controllers;

use App\Http\Controllers\Controller;
use App\Models\User;
use Inertia\Inertia;

class UserController extends Controller
{
    /**
     * Show the profile for a given user.
     *
     * @param  int  $id
     * @return \Inertia\Response
     */
    public function show($id)
    {
        return Inertia::render('Users/Profile', [
            'user' => User::findOrFail($id)
        ]);
    }
}
```
```

File: frontend.md [hunk 4]
```diff
@@ -0,0 +1,192 @@
...
+
+<a name="livewire"></a>
+### Livewire
+
+[Laravel Livewire](https://laravel-livewire.com) is a framework for building Laravel powered frontends that feel dynamic, modern, and alive just like frontends built with modern JavaScript frameworks like Vue and React.
+
+When using Livewire, you will create Livewire "components" that render a discrete portion of your UI and exposes methods and data that can be invoked and interacted with from your application's frontend. For example, a simple "Counter" component might look like the following:
+
+```php
+use Livewire\Component;
```
jessarcher, line 65: For consistency with the Inertia example below, do you think we should add this?

```suggestion
<?php

namespace App\Http\Livewire;

use Livewire\Component;
```

PR #7986, opened 2022-06-15
Clarified laravel_token cookie creation

ritenn:
This small docs fix will help people to understand how it really works.

There is a lot of confusion about laravel_token cookie creation by CreateFreshApiToken  middleware, because people usually try to implement this as POST "login" route.

Inside CreateFreshApiToken [https://github.com/laravel/passport/blob/9.x/src/Http/Middleware/CreateFreshApiToken.php#L82](https://github.com/laravel/passport/blob/9.x/src/Http/Middleware/CreateFreshApiToken.php#L82) there is condition which checks if request method is equal to "GET", so in case of POST login route - it won't create cookie.

### DIFF
--- a/passport.md
+++ b/passport.md
@@ -1161,7 +1161,7 @@ Typically, if you want to consume your API from your JavaScript application, you
 
 > {note} You should ensure that the `CreateFreshApiToken` middleware is the last middleware listed in your middleware stack.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
It may not be clear what "web" routes means in this context.

PR #7990, opened 2022-06-15
Mention the Rate Limiter is IP-based

jozydapozy:
I think it makes sense to mention this. Currently a reader could misinterpret it as if a the limit is based only on the amount of times a certain route was hit (by whoever hits the route combined)


taylorotwell:
I don't think this is necessary atm.

PR #7993, opened 2022-06-16
Scout 9 does not yet support PostgreSQL

gregmartyn:
It looks like Scout 9 no longer has PostgreSQL support.


taylorotwell:
Need more information as to why it doesn't.

gregmartyn:
If you try using SCOUT_DRIVER=database, it "searches" using `ILIKE % search term %`, which isn't an indexed full text search. Postgres full text search uses the `@@` operator. More on that here: https://www.postgresql.org/docs/current/functions-textsearch.html and the "stored generated column" part of https://www.postgresql.org/docs/current/textsearch-tables.html

PR #7994, opened 2022-06-16
Added information on how do disable wrapping on resource class

ArneAnka:
Its possible to disable wrapping on single resource classes, not just the outermost resource.


taylorotwell:
Unclear how this ties into the documentation below it.

ArneAnka:
I tought it made perfect sense. If you think of it like a flow.

1. How to change wrap 
2. How to disable wrap on single resource 
3. How to disable wrap across resources

PR #7995, opened 2022-06-17
[9.x] update Laravel documentation to contain event fake except

CharlesBilbo:
Adds the documentation for excluding events from fake and scoped events from fake.


taylorotwell:
I have added to the documentation. Thanks.

PR #7996, opened 2022-06-17
Update cache.md to include some info on Cache tags

danmatthews:
After a lengthy discussion in the Laravel UK slack group and some investigation by our members, we were all left a little confused about how cache tags worked specifically, and to avoid any potential misunderstandings, we've cobbled together a summary of our findings to use as a "buyer beware" at the top of the cache tags section.


taylorotwell:
Hey there - thanks for the PR. I've trimmed this down to adding an additional sentence to the first section of cache tags to try to make this more clear.

PR #7999, opened 2022-06-21
[9.x] Fix attachable object example missing contract implementation

tonysm:
### Fixed

- Fix attachable object example missing contract implementation


Aberdeener:
Nice, noticed this too 👍🏼

PR #8002, opened 2022-06-22
[9.x] Document global `fake()` helper

timacdonald:


siarheipashkevich:
@timacdonald hello

But this reduces performance, since we need to return an instance of the class every time we call `fake()` for each property, which would not happen when calling `$this->faker` inside the `definition` method of the factory class.

And factory initializes the faker by default when we are creating a new factory every time.

It's nice to have a `fake()` helper, but I think inside the factory we shoul use already created instance of the factory

PR #8003, opened 2022-06-23
[9.x] Add note in upgrade guide about directory permissions

driesvints:
In Flysystem v3, directory permissions are a bit stricter than in v1. Because of this, if you use different users than the user that created the directory to manipulate files in said directory, you'll run into permission issues. This can be circumvented by making these directories publicly. 

It's still recommended however, to not do this, and use the same user to manipulate these files as the directory creator.

See https://github.com/laravel/framework/issues/42586


taylorotwell:
Think I'll leave this out for now.

PR #8004, opened 2022-06-23
Added a short note for the usage of name parameter

hameesakhan:
Just added a small note to mention that the Storage::download() method automatically adds the concerned file's extension at the end of the file name and that the developer do not have to pass parameter including that file's extension.


taylorotwell:
Hmm - not sure we need a big red note for this.

hameesakhan:
Then how about adding a simple line in above paragraph for this knowledge?
It will save someone's time in future.

PR #8011, opened 2022-06-28
Add tip about Octane development experience

tontonsb:
I'm not sure if this optional is intentionally left out because it lacks a bit of elegance, but I had a great time with just

```php
php artisan octane:start --workers=1 --max-requests=1
```
taking care of code reloads while developing.

The only issue is that (at least on Swoole) the code gets loaded after the request and waits for the next incoming request as it is. So a code change requires one empty request to load the new code and then the real request to see the change. I just got used to double F5.


taylorotwell:
Hmm, think I will leave this out for now.

PR #8022, opened 2022-07-04
[9.x] Document Vite refresh option

timacdonald:
See: https://github.com/laravel/vite-plugin/pull/43, https://github.com/laravel/vite-plugin/pull/55

### DIFF
--- a/vite.md
+++ b/vite.md
@@ -328,7 +328,7 @@ module.exports = {
 <a name="working-with-blade-and-routes"></a>
 ## Working With Blade & Routes
 
-When your application is built using traditional server-side rendering with Blade, Vite can improve your development workflow by automatically refreshing the browser when you make changes to view files in your application. To watch the "default" files for changes, you can simply specify the `refresh` option as `true`.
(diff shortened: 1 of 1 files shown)


vite.md
```diff
@@ -14,6 +14,7 @@
   - [Inertia](#inertia)
   - [URL Processing](#url-processing)
 - [Working With Stylesheets](#working-with-stylesheets)
+- [Working With Views](#working-with-views)
```
jessarcher, line 17: What do you think about "Working With Blade"?

(I'd have added a suggestion, but it requires the anchor to be updated as well)

timacdonald: I put Views as I felt that "PHP" was too generic. The project might be 99% PHP. Views felt more specific and Blade felt too specific.

So I landed in the middle. Happy to change it if you still feel that "PHP" would be better?

jessarcher: My only concern with "Views" is the term can apply to JavaScript views, which this configuration is not intended for.

timacdonald: Maybe "Vues" just to really lean into it?

Yea, I get that. What do you think about leaning to the Blade direction then?

"Working with Blade"

jessarcher: That's what I first thought, but it doesn't really cover the routes.

Is "Working with Blade and Routes" too much?

timacdonald: I can get behind this TBH

PR #8023, opened 2022-07-04
Fix return type of scopeActive example

pacoorozco:
Reviewing the documentation I've seen that a code block is nor right:
```
    /**
     * Scope a query to only include active users.
     *
     * @param  \Illuminate\Database\Eloquent\Builder  $query
     * @return void
     */
    public function scopeActive($query)
    {
        $query->where('active', 1);
    }
```
The method `scopeActive($query)` should return an instance of ` \Illuminate\Database\Eloquent\Builder` instead of a `void`.

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -1185,7 +1185,7 @@ Scopes should always return the same query builder instance or `void`:
          * Scope a query to only include active users.
          *
          * @param  \Illuminate\Database\Eloquent\Builder  $query
-         * @return void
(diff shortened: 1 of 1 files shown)


pacoorozco:
My previous comment is probably wrong, I've seen a message before the code block:

`Scopes should always return the same query builder instance or void:`

So maybe, it was showing both options... I'm gonna close the PR.

PR #8024, opened 2022-07-05
Update vite.md

shealan:
Added additional information on customising the SSR file output directory. This should help Envoyer users from getting stung by this potential gotcha.

### DIFF
--- a/vite.md
+++ b/vite.md
@@ -392,4 +392,24 @@ npm run build
 node storage/ssr/ssr.js
 ```
 
+### Customising Your SSR File Output Directory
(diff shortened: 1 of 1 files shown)


jessarcher:
Thanks, @shealan. I'm trying to think if there's a different location we can place it by default rather than needing extra documentation.

I'm worried that placing it in the `public` directory opens a potential security concern as this file is a server and not something for browsers to download.

I'm thinking `bootstrap/ssr/ssr.js` is a contender as Laravel already uses this directory for storing files generated at build-time (as opposed to files generated at run time like in `storage`)

shealan:
@jessarcher yes I think`/bootstrap/ssr/ssr.js` makes a lot of sense.

taylorotwell:
@jessarcher yeah, I would prefer to just move this file now so it works well with our first-party tools.

PR #8027, opened 2022-07-06
Remove reference to --view option

tobya:
`--view` option does not seem to exist on the ComponentMakeCommand so reference to this option should be remove from the docs.

Calling

> php artisan make:component Alert --view

raises an error 

>   The "--view" option does not exist.


taylorotwell:
CleanShot 2022-07-06 at 10 48 38@2x

PR #8029, opened 2022-07-06
Update eloquent.md

morpheus7CS:
Hey,

I was just going through some of the Eloquent scopes documentation and realized that there might be a return value missing in the scopeActive example.

Kind regards,
g


taylorotwell:
It is not necessary to return here.

PR #8035, opened 2022-07-09
[9.x] Updated Routing Defining Rate Limiters to Include Proper Boot Method Configuration

vsahinid:
The `configureRateLimiting` method in the `boot` method of the application needs to be set initially before the rate limiter is set.


taylorotwell:
It's already there in the default skeleton.

PR #8049, opened 2022-07-17
Vite 3 support

timacdonald:


jessarcher (approved):
This is good to go once the skeleton is shipping with the latest vite plugin

PR #8070, opened 2022-07-24
Just fixing the typo

mahmoudmohamedramadan:
This PR fixes the typo

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -343,7 +343,7 @@ If validation fails, a redirect response will be generated to send the user back
 <a name="adding-after-hooks-to-form-requests"></a>
 #### Adding After Hooks To Form Requests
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
The original wording is actually correct. You are adding "after" rules **before** the validator executes.

mahmoudmohamedramadan:
How! but when I tried to follow the validation cycle, I found that the `rules` method executes before `withValidator` method

mahmoudmohamedramadan:
Here is the link, where `after` hooks triggered after passes through the `rules`

https://github.com/laravel/framework/blob/abf06591936a897ff703516a913945a597e789f5/src/Illuminate/Validation/Validator.php#L432

PR #8071, opened 2022-07-25
Fix Eloquent event name trashed to softDeleted

medilies:
When you check the `SoftDeletes` trait, you find that the name of the static method that registers a closure for the `trashed` event is named `softDeleted`.

I don't know about the method name for the **observers** since it isn't mentioned explicitly in the docs and the stubs do not include it.

But if itsn't accurate to change the name of the event as I did in this PR. I guess the docs need a note in the [Using Closures](https://laravel.com/docs/9.x/eloquent#events-using-closures) section.


taylorotwell:
CleanShot 2022-07-25 at 08 23 20@2x

PR #8073, opened 2022-07-25
Fix the issue of not defined route

mahmoudmohamedramadan:
This PR fixes the issue of undefined `verification.resend` route


taylorotwell:
Not sure I understand - you can name this whatever you want? How do you get an undefined route?

mahmoudmohamedramadan:
Oops, I catch the error, the issue was from `laravel/ui`.

PR #8074, opened 2022-07-25
[9.x] Avoid to make an implicit use of union types with `void` functions

phansys:
Mixing any type and `void` is allowed by PHP if the functions are not using type declarations. In fact, internally it converts those return values to `null`. But from a stricter point of view, this behavior is conceptually wrong. That's why union types don't allow to use `void` alongside other types.
This PR tries to cover that, by using an explicit `return null` statement instead of just do nothing.
The return type declaration was added just to clarify the situation, but IMO, it can be omitted if it goes against the current recommendations.

See:
* https://wiki.php.net/rfc/void_return_type
* https://www.php.net/manual/es/migration71.new-features.php#migration71.new-features.void-functions
* https://php.watch/versions/8.0/union-types.

### DIFF
--- a/authorization.md
+++ b/authorization.md
@@ -233,6 +233,8 @@ Sometimes, you may wish to grant all abilities to a specific user. You may use t
         if ($user->isAdministrator()) {
             return true;
         }
+
+        return null;
     });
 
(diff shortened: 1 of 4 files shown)


driesvints:
This isn't necessary

PR #8082, opened 2022-07-29
Add documentation about "name" option at scheduling

danilopinotti:


scheduling.md
```diff
@@ -278,6 +278,15 @@ To indicate that the task should run on only one server, use the `onOneServer` m
                     ->at('17:00')
                     ->onOneServer();
 
+Sometimes you will need to schedule the dispatch of same Job Class multiple times, but using different parameters. If you are using `onOneServer` option, only the first call will be dispatched. To avoid this behaviour, you need to define a unique name in each task:
```
dammy001, line 281:
```suggestion
Sometimes you will need to schedule the dispatch of the same Job Class multiple times, but using different parameters. If you are using `onOneServer` option, only the first call will be dispatched. To avoid this behaviour, you need to define a unique name for each task:
```

danilopinotti: Thanks @dammy001 for reviewing

taylorotwell:
I don't feel like this example is very clear.

PR #8083, opened 2022-07-30
Add watch option for roadrunner server

crazy252:
This will add the watch option for the roadrunner server. If you follow the docs you will maybe be stuck with a laravel projects where you can't see changes because of the missing watch option. This will fix it


taylorotwell:
This would need to include language about chokidar, etc. which is required to be installed before using watch.

PR #8094, opened 2022-08-02
Mention try count and exception scenario

garrettboone:
This is from a recent issue I describe here https://github.com/laravel/framework/discussions/43502


taylorotwell:
I'm not sure I totally follow this note. Will look into the issue.

garrettboone:
@taylorotwell  Regarding job overlapping, I interpret "prevent" as "keep job from starting" but it appears in practice it is behaving as "keep job from continuing".

PR #8096, opened 2022-08-03
Update return type of example `before` method for policy filters

eli-s-r:
The [current return type](https://github.com/laravel/docs/blame/c3ee189/authorization.md#L517) of the example, `void|bool`, contradicts the [description below the example](https://github.com/laravel/docs/blame/c3ee189/authorization.md#L526) that explains the meaning of an explicit `null` return value.


taylorotwell:
I think it's fine.

PR #8102, opened 2022-08-05
Fix typo in `Eloquent local scopes`

jordanhavard:
The local scope is missing the `return` element. This change fixes that


jordanhavard:
Well after reviewing this, the return type is void...

PR #8103, opened 2022-08-05
Exists Rule class with addional condition query

Nir-An:
1. Simplified the example, by removing complex where query.
2. Change to a bit more common example with a users table and real column ('email_verified_at')

3. Being more clear about how extended the query is going to be,
it doesn't replace existence checking, "out of the box" it is just adding
extra conditions to the same query.


taylorotwell:
Current docs seem fine.

PR #8104, opened 2022-08-08
[9.x] Document combining global scopes with Scout

driesvints:
Documents https://github.com/laravel/scout/issues/639 since we're not pursuing a fix right now (see https://github.com/laravel/scout/pull/640).


taylorotwell:
@driesvints how does this tie in with global scopes? I don't see global scopes mentioned in the issue unless I'm just overlooking it?

driesvints:
@taylorotwell hmm did you look somewhere else? It's mentioned everywhere in the issue: https://github.com/laravel/scout/issues/639

PR #8118, opened 2022-08-12
[9.x] Separate Eloquent factory docs

jessarcher:
The model factory documentation is currently located on the database testing page. Model factories are very commonly used in database seeders as well, so I don't find the current location particularly intuitive, and I often fumble around trying to find it.

Additionally, the model factory documentation makes up the majority of the content on the database testing page, which makes it a little overwhelming when looking for other content specific to database testing (such as available assertions).

This PR aims to solve this by moving the model factory documentation to a new "Factories" page under the "Eloquent ORM" heading.

I've left a section on the database testing page that briefly introduces model factories, but refers the user to the new page for the full documentation. This is very similar to what is already on the database seeding documentation.

I haven't really written any new documentation - I've just leveraged existing documentation for the model factories section on the database testing page, and I've made some minor tweaks so that everything makes sense in this new standalone location. I've separated the commits so that the first commit only moves the documentation and updates the links, and any changes to the wording are in subsequent commits, making it easier to review the actual content changes.

Of course, I'll understand if the current location is preferred :slightly_smiling_face:


taylorotwell:
This looks good - but for some reason the "Notes" aren't rendering locally on this page. Strangely, they render fine on other pages within the documentation.

@driesvints @jessarcher any ideas?

image

jessarcher:
My bad, it looks like the implementation on laravel.com requires two trailing whitespaces after the `**Note**` to generate a `<br>` and I had removed them because the GitHub implementation doesn't require them.

@driesvints is it worth removing this requirement so that it's compatible with GitHub?

driesvints:
@jessarcher yes the two whitespace characters were required otherwise it would inline the note. But I agree it's probably better to not make it a requirement.

PR #8120, opened 2022-08-12
Add documentation for new Ably broadcaster and connector

owenpearson:
**Note: this PR is one of four parallel pull requests to add first class support for Ably to Laravel. Please find the other PRs here:**
- laravel/laravel#5969
- laravel/framework#43685
- laravel/echo#351

Background: the current Ably broadcaster uses the Pusher JavaScript SDK and the documentation instructs users to use Ably with the Pusher compatibility mode enabled. We’ve received a lot of feedback that various features of the Ably Laravel integration don’t work as expected, often caused by the fact that there’s a dependency on the Pusher JavaScript SDK which we have no control over. This collection of PRs adds a new broadcaster which uses the ably-js SDK which means that we can ensure that the Ably broadcaster will remain stable. We are more than happy to take feedback and answer any questions about the work presented here :)

Adds usage instructions for the new Ably broadcaster. See the PRs linked above for implementation.


taylorotwell:
Resubmit this if the others get accepted.

PR #8127, opened 2022-08-16
Collection `every()` method with comparison operator

schulerj89:
In collections the `every()` method also takes the comparison operator.


taylorotwell:
Honestly think it's easier to just not document this. We would have to list all valid operators, etc.

PR #8128, opened 2022-08-16
Added Vite Documentation For Build Path

webdevhayes:
* Vite.md - Added documentation for build path

### DIFF
--- a/vite.md
+++ b/vite.md
@@ -16,7 +16,7 @@
 - [Working With Stylesheets](#working-with-stylesheets)
 - [Working With Blade & Routes](#working-with-blade-and-routes)
 - [Custom Base URLs](#custom-base-urls)
-- [Override Build Path](#override-build-path)
+- [Override BuildDirectory](#override-build-directory)
 - [Environment Variables](#environment-variables)
 - [Disabling Vite In Tests](#disabling-vite-in-tests)
 - [Server-Side Rendering (SSR)](#ssr)
@@ -411,8 +411,8 @@ https://cdn.example.com/build/assets/app.9dce8d17.js
 
 Remember that [absolute URLs are not re-written by Vite](#url-processing), so they will not be prefixed.
 
-<a name="override-build-path"></a>
-## Override Build Path
+<a name="override-build-directory"></a>
+## Override Build Directory
 
 You can customise the folder name which vite uses when compiling. The example will change the folder name from 'build' to just 'folderName'
 
@@ -424,7 +424,7 @@ export default defineConfig({
     plugins: [
         laravel({
             // Default is 'build' 
-            buildPath: 'folderName'
+            buildDirectory: 'folderName'
         }),
     ],
 });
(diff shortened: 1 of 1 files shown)


vite.md
```diff
@@ -410,6 +411,25 @@ https://cdn.example.com/build/assets/app.9dce8d17.js
...
+
+```js
+import { defineConfig } from 'vite';
+import laravel from 'laravel-vite-plugin';
+
+export default defineConfig({
+    plugins: [
+        laravel({
+            // Default is 'build' 
+            buildPath: 'folderName'
```
jessarcher, line 427: This should be `buildDirectory`

vite.md
```diff
@@ -410,6 +411,25 @@ https://cdn.example.com/build/assets/app.9dce8d17.js
 
 Remember that [absolute URLs are not re-written by Vite](#url-processing), so they will not be prefixed.
 
+<a name="override-build-path"></a>
+## Override Build Path
```
jessarcher, line 415: Rather than add another top-level heading, I think this would be better in the "Configuring Vite" section.

jessarcher (commented):
The Laravel Vite plugin has quite a few configuration options, but many of them, like this one, are only intended for exceptional use cases.

Is there a common use case for changing this from the default? If not, I think I'd prefer not to encourage changing it.

If we were to add this, I wonder whether we should just have a more general section under "Configuring Vite" that lists all of the configuration options.

vite.md
webdevhayes: Yup you are right.. no idea where/why I did build path .

webdevhayes:
> The Laravel Vite plugin has quite a few configuration options, but many of them, like this one, are only intended for exceptional use cases.
> 
> Is there a common use case for changing this from the default? If not, I think I'd prefer not to encourage changing it.
> 
> If we were to add this, I wonder whether we should just have a more general section under "Configuring Vite" that lists all of the configuration options.

I am not really sure being able to change the build folder name is an "exceptional" circumstance. I think it's a configuration that should be known to people without source diving. 

I am not really sure why you would "discourage" someone being able to change the build folder name or any other configuration for that matter. Laravel mix you literally could customise everything how you wanted on personal preference base. By saying you want to discourage people from using an option is effectively pushing an opinionated thought process on to someone else. 

Also as the build tool has changed to vite over mix then people need to know what the tool can do in order to migrate say from mix to vite, by limiting knowledge/documentation it will just divide people in moving/changing tools because its not obvious how to customise something they need before migrating. 

I agree that maybe having in the documentation a section around configuration however this change in particular is not related to vite but the plugin for vite. In theory the read me should be updated in the plugin repo and have a link to in the vite documentation to the plugin read me.

vite.md
webdevhayes: Please read my reply below

taylorotwell:
Hmm - I'm still not sure a strong reason for changing this directory has been presented.

webdevhayes:
ok.. I am not really sure why there is push back. Like any package/module/plugin it should be documentated what it can do in terms of functionality 

Even vite itself has all the documentation for example around rollup.. seeing as laravel has a plugin for vite this should have the documentation it needs for configuration and save people time. I just can't see the logic behind rejecting trying improve the documentation to help others regardless how insignificant you think it will help people.

Maybe not how I have presented it in this pull request however the repo here - https://github.com/laravel/vite-plugin should have a better read me i think rather than pointing to the laravel site. The configuration options should be mentioned in the read me.

PR #8130, opened 2022-08-17
Fix value object casting example

choowx:
The properties of the `Address` should be wrapped in a `$key` key when returning in the custom cast `set` method.


taylorotwell:
Shouldn't be necessary.

PR #8133, opened 2022-08-17
Added description of how to view the app

EightArmsWebmaster:
When using Vite for a standalone project, the Vite URL shows the application. I wasn't clear that with Laravel Vite, this URL is only used internally. I hope this addition will make it clear to future users that, while they must still run Vite server, they should then return to the orginal local app url to continue with development.

### DIFF
--- a/vite.md
+++ b/vite.md
@@ -183,6 +183,9 @@ npm run dev
 npm run build
 ```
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
No plans to merge this atm.

PR #8135, opened 2022-08-18
[9.x] Cursorpagination: query expressions with parameters not supported

gdebrauwer:


taylorotwell:
What is a query expression with parameters?

gdebrauwer:
```php
User::selectRaw('(CONCAT(firstname, ?) as foo', ['bar'])->cursorPaginate();
```
when providing a cursor to cursorPaginate, the `(CONCAT(firstname, ?)` will be added as a where clause, but `?` will not be filled with the correct value

PR #8143, opened 2022-08-20
Update migrations.md

vauteer:
Missing array brackets when creating a unique index with more fields.


taylorotwell:
That's not what this is demonstrating though.

vauteer:
Sorry, I must have been a little unfocused.

PR #8144, opened 2022-08-21
[9.x] Correct shaky notations on collections.md.

daikisuyama-fs:
## How I found

When I grep with "inverse" on collections.md, I found that the past version of the documentation have a mixture of those that say "method" and those that don't.

## How to correct

By unifying those that say "method".


taylorotwell:
Current docs seem fine?

daikisuyama-fs:
@taylorotwell 
Sorry for the inadequate explanation.

Can you see the below link?
You can see "For the inverse of the reject method, see the filter method."
Only this line says "For the inverse of the xxx method" and the other lines say "For the inverse of the xxx".
So, I thought the former expression is better and fixed to it.

https://laravel.com/docs/9.x/collections#method-reject

PR #8145, opened 2022-08-21
Update foreign key name to comment_id

nikolaynikolaevn:
If I understand correctly, this should be `comment_id` instead.


nikolaynikolaevn:
Actually, it is correct the way it is. My bad.

PR #8153, opened 2022-08-22
[9.x] Add Meilisearch Filterable Attribute Section

Cav19:


taylorotwell:
Your PR doesn't give any context as to where I'm supposed to get that `$client` variable from?

PR #8155, opened 2022-08-23
[9.x] Document broadcastOn message

mpociot:
This PR documents the `broadcastOn` method that exists on a notification class.
That's right now the only way to use on-demand notifications in combination with broadcasting, so I thought it'd be nice to have it in the docs 👍

### DIFF
--- a/notifications.md
+++ b/notifications.md
@@ -986,6 +986,21 @@ If you would like to customize which channel that an entity's broadcast notifica
         }
     }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Would rather just make it actually possible - so I did 😅

https://github.com/laravel/framework/commit/d2b14466c27a3d62219256cea27088e6ecd9d32f

```php
Notification::route('broadcast', [new PrivateChannel('foo')])
        ->notify(new InvoicePaid);
```

PR #8166, opened 2022-08-26
Asset inline limit config when using views and static files

lukesnowden:
Production build will fail using `Vite::asset()` if file is less than 4kb.

### DIFF
--- a/vite.md
+++ b/vite.md
@@ -366,6 +366,16 @@ import.meta.glob([
 ]);
 ```
 
+By default, Vite will only add assets to the manifest file if they are greater than 4kb. If they are less, Vite will try to inject base64encoded URL strings which will fail using static files and blade views as Laravel will try to read the asset from the manifest file. To overcome this, simply add the following to your `vite.config.js` file.
+
+```js
+export default defineConfig({
+    build: {
+        assetsInlineLimit: 0
+    },
+});
+```
+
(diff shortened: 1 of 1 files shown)


jessarcher:
I'd be inclined to always enable this via https://github.com/laravel/vite-plugin

Alternatively, we could update the Vite class to render the base64.

timacdonald:
I think I'd be keen to set it to `0` in the plugin, respecting userland config of course.

I don't think we can generally assume the file will still be in it's pre-bundled location (e.g. `resources/images/pixel.png`) as apps that clear that stuff out for space reasons (e.g. Serverless). I know I've worked on apps that don't include this stuff in their deployed apps. Encoding at runtime kinda sucks as well and the inconsistency is a little weird.

At least if you specify it, you are opting into the behaviour and can handle it in your own app.

timacdonald:
Pull request: https://github.com/laravel/vite-plugin/pull/131

lukesnowden:
I’m happy either way I just wanted to shine a light on the fact it breaks in production which no one wants

lukesnowden:
Also thanks to Jaap for figuring it out for me https://stackoverflow.com/questions/73502963/larave-vite-vite-manifest-missing-odd-files

timacdonald:
Appreciate it @lukesnowden! I was unaware this was a Vite feature. I think we should head down the path of fixing in the plugin as Jess suggested, so I will close this one.

PR #8170, opened 2022-08-30
Add Scout take() limit information

benholmen:
This PR adds information about Scout's `take` method and how it differs from Eloquent's `take` method. I was pretty tripped up by `Model::search($term)->get()` only returning some results - limited by Meilisearch's `limit` default - and I think this documentation will help other avoid my mistake.

### DIFF
--- a/scout.md
+++ b/scout.md
@@ -520,6 +520,14 @@ You may use the `whereIn` method to constrain results against a given set of val
 
 Since a search index is not a relational database, more advanced "where" clauses are not currently supported.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
This feels very Meili specific and a bit confusing. Will revisit.

benholmen:
@taylorotwell Agreed that my doc suggestion was very Meili-specific. The things that would have been most helpful to me was knowing about `take` without source diving - any mention of "limit" or "take" on the Scout doc page would be helpful.

PR #8171, opened 2022-08-30
[9.x] Remove trailing whitespace

jessarcher:
This PR removes trailing whitespace from the docs.

In https://github.com/laravel/laravel.com/pull/256 I removed the need for two trailing whitespace following the callout type in GitHub-style note/warning quotes. The GitHub implementation doesn't require the trailing whitespace and their examples don't include it. I feel it makes sense to remove it from our existing docs as I think future contributions are more likely not to include them, so this will hopefully make things consistent.

While searching for these, I also found several other places where trailing whitespace exists, mostly in code samples. While they don't cause any harm, I think it's nicer without it.


driesvints:
Ow, can we please not do this? This will inline all headers of notes which look worse on GitHub than the non-inlined version 😢 

Screenshot 2022-08-30 at 09 25 02

driesvints:
I mean, it's nbd. I guess it's just my personal preference 😅

jessarcher:
> **Note**
> Test

jessarcher:
Seems like the newline is a bit inconsistent with GitHub at the moment - in PR comments and the diff view it shows on a separate line. But when browsing to a markdown file directly it's inline.

I'm happy to close this PR. Personally, I'd take the inline note over the trailing whitespace (never liked that Markdown syntax), but I don't have a strong opinion about it.

PR #8173, opened 2022-08-31
[9.x] Use constructor property promotion instead of old constructor syntax

omerbaflah:


driesvints:
No plans for this sorry

omerbaflah:
Hey @driesvints I think this will be fine if it merged and property promotion allows us to combine class fields, constructor definition and variable assignments all into one syntax instead of doing old syntax

what are your thoughts?

PR #8183, opened 2022-09-03
Fix method signature in queue docs for `->catch()` callback

ralphjsmit:
Today I got an error in my error reporting tool about a wrong parameter type. I checked with the docs and it looks like the docs are wrong, since the `$batch` parameter is given before the `$exception` parameter.

See [this part](https://github.com/laravel/framework/blob/7fa4beb386a139cabb730a7e3b9571ba8ea2a181/src/Illuminate/Bus/Batch.php#L334) of the source code.


taylorotwell:
Are you sure this is correct? Chains are not batches.

PR #8186, opened 2022-09-05
HTML5 selected attribute

Rattone:
Insted of selected="selected", that sounds very old and to prevent horizzontal scroll on desktop

before:
image

after:
image


Rattone:
@taylorotwell Can it be useful to make the same pull request for previous versions, for consistency?
From what I've seen it was introduced in version 7.x. For me it would be a pleasure.

PR #8190, opened 2022-09-07
[9.x] Add custom messsages for rule objects

nshiro:
related
https://github.com/laravel/framework/pull/41145
https://github.com/laravel/framework/pull/43925

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -2073,6 +2073,21 @@ Or, if your validation rule requires access to the validator instance performing
         }
     }
 
+#### Customizing The Error Messages
+
+If needed, you may provide custom error messages instead of the default ones provided by the rule objects.
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not sure why it would be necessary. This is an application rule you define so it seems like your rule would just have the error message you want?

nshiro259:
@taylorotwell
It's ok if you think so.
But I just saw you mentioned 'can you make a PR to the docs?' in https://github.com/laravel/framework/pull/41145
That's way I sent a PR.

nshiro:
@taylorotwell
It's ok if you think so.
But I just saw you mentioned 'can you make a PR to the docs?' in https://github.com/laravel/framework/pull/41145
That's why I sent a PR.

PR #8192, opened 2022-09-07
[9.x] Document `extend` method for scopes

saulens22:


taylorotwell:
I dunno - I'm not sure I totally love this feature 😬

saulens22:
It's already used in `SoftDeletingScope` in Laravel's core, just not documented.

https://github.com/illuminate/database/blob/8aa4ab575a56d1831eb25344871d46f05ad2cf09/Eloquent/SoftDeletingScope.php#L32

It's defined in `Builder` here:

https://github.com/illuminate/database/blob/8aa4ab575a56d1831eb25344871d46f05ad2cf09/Eloquent/Builder.php#L164-L166

The only reference I've found is this link: https://stackoverflow.com/questions/42228542/why-method-existsscope-extend-exists-in-method-withglobalscope-of-lara

Maybe I could've used better wording for documentation😅

PR #8198, opened 2022-09-10
Update upserts section

margarizaldi:
Just add a little explanation if the third argument of `upsert` is omitted.


taylorotwell:
Not sure this is necessary.

margarizaldi:
Ah sure this is not necessary... I forgot that too detailed docs will make tutorials unsold ...

PR #8199, opened 2022-09-11
Update Task Scheduling Events example

salkuadrat:
Update code example for Task Scheduling events, so it will be more align with the style of event listener mappings at the Laravel's Events documentation


taylorotwell:
Too many imports.

salkuadrat:
@taylorotwell 

Since (most) code editor can auto import php classes when developing project with Laravel, I think it's better to present a uniform style in writing this event listener mappings in all parts of Laravel documentation. 

Especially for beginners who just starting their journey with Laravel, it can give them more clarity and less confusion about the style (should I write a full class namespace, or can I just write the simple ::class like in the Events docs?)

Just for consideration...

PR #8201, opened 2022-09-11
Update Authentication Events example

salkuadrat:
Update code example for Authentication events, so it will be more align with the style of event listener mappings at the Laravel's Events documentation


taylorotwell:
Too many imports.

PR #8207, opened 2022-09-13
Document @props being converted to camelCase

q--:
It was mentioned nowhere in the docs that \@props were treated differently from other attributes and underwent the same case conversion as class arguments. It also wasn't mentioned that `snake_case` got converted and not only `kebab-case`.

### DIFF
--- a/blade.md
+++ b/blade.md
@@ -762,7 +762,7 @@ When your component is rendered, you may display the contents of your component'
 <a name="casing"></a>
 #### Casing
 
-Component constructor arguments should be specified using `camelCase`, while `kebab-case` should be used when referencing the argument names in your HTML attributes. For example, given the following component constructor:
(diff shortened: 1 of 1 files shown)


q--:
Okay, I'll try to explain. Suppose you have an anonymous blade component, which you include in the following way:

<x-test label_html="label_html"/>

Now, inside the component (test.blade.php) you're trying to get the contents of the attribute. Since this is an anonymous component without an accompanying PHP class, you've got two options:

1. Get it using `{{$attributes['label_html'}}`. This works as you'd expect, but I'm 
   a. not using it as an HTML attribute, and 
   b. the contents are [getting passed through `htmlentities`](https://github.com/laravel/framework/issues/34346), where I want the raw HTML.
2. so, the alternative is using @props. I defined them using `@props(['label_html'])` and they didn't work, because it turned out Laravel is converting them to camelCase like it does for constructor arguments. (So, it needs to be  `@props(['labelHtml'])`.)  The docs do not mention this, hence this pull request.

PR #8218, opened 2022-09-16
[10.x] Duration handlers

timacdonald:


driesvints:
@timacdonald should this PR be reconsidered?

PR #8219, opened 2022-09-16
Fixing new subscription via invoice example

axeloz:
The current documentation on the creation of new subscription with payment on invoice does not work until the `days_until_due` argument is supplied. So I modified the example code snippet to: 

```
    $user->newSubscription('default', 'price_monthly')->createAndSendInvoice([], [
        'days_until_due' => 30
    ]);
```


driesvints:
We added default support for this to Cashier Stripe. I've sent in a PR to reword this. Thanks

https://github.com/laravel/docs/pull/8233

PR #8225, opened 2022-09-17
[9.x] Make Laravel handle URLs ending .php

pistvan:
This little fix enables Laravel to handle URLs ending with `.php`.

Otherwise, requests like `/not-existing-file.php` return a `File not found.` plain text response instead of Laravel's error page, while nginx writes the following to it's error log:
```
[error] 16#16: *85881 FastCGI sent in stderr: "Primary script unknown" while reading response header from upstream
```

### DIFF
--- a/deployment.md
+++ b/deployment.md
@@ -73,7 +73,7 @@ server {
 
     error_page 404 /index.php;
 
-    location ~ \.php$ {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Change this to whatever you want for your own needs.

PR #8234, opened 2022-09-20
[11.x] Document implicit commit usage with batches

driesvints:
It's not possible to use database statements that cause implicit commits within batches since batches are wrapped in a transaction. See https://github.com/laravel/framework/issues/44152

I added this as a second warning because I thought this was the correct place to warn about it.


driesvints:
I took the same example as in the issue:

```php
Bus::batch([function () {
    DB::statement('CREATE TEMPORARY TABLE IF NOT EXISTS temporary_table (
        `id` int(11),
        `ColumnToDrop` varchar(200)
    ) ENGINE=INNODB');

    DB::statement("ALTER TABLE temporary_table DROP `ColumnToDrop`");
}])->dispatch();
```

taylorotwell:
Hmm, wow - pretty edge case ... let me think if I want to add a big warning to the docs for this.

kmizzi:
Just a note here on the practical implication of this.  If we are doing large amounts of data importing involving updates & inserts, we want to use batched jobs and the processing would involve the use of temporary tables for optimum import performance.

If we create a trait to handle bulk data processing abstractly, the use of temporary tables would rely on statements that would cause implicit commits.

So I don't feel it is as much of an edge case as a situation that occurs from handling large data imports in an optimal way.

driesvints:
@taylorotwell wondering if this PR can be merged?

taylorotwell:
We would need to work it into the docs and give an actual example of what we mean by implicit commits, etc.

driesvints:
@taylorotwell I've tried to write it out with an example. Hope this is clear enough.

PR #8236, opened 2022-09-20
[9.x] Adds working example to Closure Request Guards chapter

maxim-kravets:
In the last version of Laravel the method `viaRequest` was moved from `Illuminate\Support\Facades\Auth` to `Illuminate\Auth\AuthManager`, so the documentation shows a non-working example.


taylorotwell:
The Auth facade points to AuthManager.

PR #8241, opened 2022-09-21
Include a section on transforming the paginator's collection using the through method 

Tunsworth:
Currently, there is no information in the docs on how to use the 'through' method on the returned pagination collection. It could be helpful to include a section on this as it is a very useful method when working with the paginated data.


taylorotwell:
Can you make the capitalization and such match the rest of the docs and also provide a real example of a transformation you might perform instead of just returning the user unchanged?

PR #8242, opened 2022-09-22
Clearing some things about Service Providers

salkuadrat:
Clearing some things about Service Providers to make it easier to digest.

...and remove some unnecessary `<?php` from code examples to make it compact and more friendly to read.


taylorotwell:
I think the current docs are pretty much OK. Thanks.

PR #8247, opened 2022-09-22
[9.x] Add early introduction to queue:work command

salkuadrat:
Add early introduction to the default `php artisan queue:work` Artisan command (without any options).


taylorotwell:
Not sure this is necessary.

PR #8252, opened 2022-09-26
add guest checkout docs

alexgaal:
This PR adds a low level documentation of guest checkout which will be released in a few in [laravel/cashier-stripe](https://github.com/laravel/cashier-stripe).

### DIFF
--- a/billing.md
+++ b/billing.md
@@ -67,19 +67,28 @@
 - [Testing](#testing)
 
 <a name="introduction"></a>
+
 ## Introduction
 
-[Laravel Cashier Stripe](https://github.com/laravel/cashier-stripe) provides an expressive, fluent interface to [Stripe's](https://stripe.com) subscription billing services. It handles almost all of the boilerplate subscription billing code you are dreading writing. In addition to basic subscription management, Cashier can handle coupons, swapping subscription, subscription "quantities", cancellation grace periods, and even generate invoice PDFs.
+[Laravel Cashier Stripe](https://github.com/laravel/cashier-stripe) provides an expressive, fluent interface
+to [Stripe's](https://stripe.com) subscription billing services. It handles almost all of the boilerplate subscription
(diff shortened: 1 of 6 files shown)


billing.md
```diff
@@ -1922,6 +1923,13 @@ When this method is invoked, a new checkbox will be available to the customer th
 > **Warning**  
 > If you have already configured [automatic tax collection](#tax-configuration) in your application's service provider then this feature will be enabled automatically and there is no need to invoke the `collectTaxIds` method.
 
+<a name="guest-checkout"></a>
+## Guest Checkout
+
+You might want to use guest checkouts for users that are not registered. When this method is called Cashier automatically creates a guest checkout session for you and you can inject you items, session configuration as you are used to via `create`.
+
+    $checkout = Checkout::guest()->create($items, $sessionOptions);
```
driesvints, line 1931: Maybe also add an example of how they can apply a discount here:

`->withDiscount()`

alexgaal: Thank you, I added an example. I'll dive into the implementation this week and could provide some detailed PR for this docs.

alexgaal:
I‘ll provide some further details, thank you for your improving commit. 👏

alexgaal:
I improved the documentation a little more, still not perfect imho.

driesvints:
@alexgaal can you remove the .idea folder? 

See https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

alexgaal:
> @alexgaal can you remove the .idea folder?
> 
> See https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

Thank you for that tip, I just added `.idea` to my global gitignore. 🎉

driesvints:
I tried to improve this even further by referencing the `checkout.session.completed` at the end and how to fulfill orders. @taylorotwell I don't think we need to explain `$items` as we don't do that anywhere in the other checkout docs. It should be pretty obvious that you can pass prices.

PR #8267, opened 2022-09-29
[9.x] Removes section on local network sharing

thinkverse:
This PR removes the section on local sharing which was added to the documentation in https://github.com/laravel/docs/pull/5677. The reasoning for this removal is that the feature doesn't exist in Laravel Valet. If a user follows the steps outlined in the documentation they'll always get back the default Laravel Valet 404 not found page.

I'm not sure why it was added to the documentation, but it seems at least @driesvints knows this feature doesn't exist. Given issue (https://github.com/laravel/valet/issues/1093) was converted to a discussion for feature requests (https://github.com/laravel/valet/discussions/1143) instead of an issue.

There is also the following comment on another issue regarding local network access (https://github.com/laravel/valet/issues/440#issuecomment-579830736) where a user suggested a fix for getting local networking to work, but that requires changing internals for Laravel Valet.

As Laravel Valet is now there is no feature for local network sharing and the section can be removed.

### DIFF
--- a/valet.md
+++ b/valet.md
@@ -12,7 +12,6 @@
 - [Sharing Sites](#sharing-sites)
     - [Sharing Sites Via Ngrok](#sharing-sites-via-ngrok)
     - [Sharing Sites Via Expose](#sharing-sites-via-expose)
-    - [Sharing Sites On Your Local Network](#sharing-sites-on-your-local-network)
 - [Site Specific Environment Variables](#site-specific-environment-variables)
 - [Proxying Services](#proxying-services)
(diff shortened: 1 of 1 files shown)


thinkverse:
I made this PR a draft since https://github.com/laravel/valet/issues/440#issuecomment-579830736 suggests a fix, and maybe it's good enough to have added? Then this PR wouldn't be needed.

driesvints:
I actually didn't know this feature existed. If that's broken somehow we need to look into it. Can you maybe send in a PR with that fix?

PR #8274, opened 2022-10-05
Upserts and UUID & ULID Keys

plakhin:
It turns out that Upserts require auto-incrementing primary key as mentioned in https://github.com/laravel/framework/issues/44430


taylorotwell:
I'll look into this.

PR #8275, opened 2022-10-05
Replace route link to more accurate point

zobay:
## Description 
- The discussion is about running a queue worker `You will also need to configure and run a [queue worker](https://laravel.com/docs/9.x/queues).` But, it points to queues main page. It should be linked to `(https://laravel.com/docs/9.x/queues#running-the-queue-worker)`.
- Such specific routing is available is same page of documentation, such as `Typically, this value should be set via the ABLY_KEY [environment variable](https://laravel.com/docs/9.x/configuration#environment-configuration):`


taylorotwell:
I think this one is OK.

PR #8281, opened 2022-10-07
[9.x] Add information on how increment method behave

croustibat:
When i review, i too often see 

`DB::table('users')->increment('votes')->save()`

so now i can tell, check the doc, everything is written in the doc !


taylorotwell:
That's exactly how we already document it.

PR #8295, opened 2022-10-17
Replace code example using `LARAVEL_START` constant with `now()`

adamhopkinson:
It appears that `LARAVEL_START` no longer exists (in my v9.31 instance at least), so I propose replacing the carbon command utilising this with `now()`


taylorotwell:
https://github.com/laravel/laravel/blob/9.x/public/index.php#L6

PR #8308, opened 2022-10-24
Clarify middleware avoid chainable header methods

darynmitchell:
See PR #8305 for the rationale on why a note is needed; TLDR: developers are having their middleware break when following current Response docs.

On that PR @taylorotwell gave feedback to not put big code blocks in a warning,  
so this is a redo, with a more concise paragraph.


taylorotwell:
I dunno - I may find a better place to document this. Thanks.

PR #8317, opened 2022-10-27
Added $mail->build, otherwise subject remains null

JBtje:
For both queued mails and non-queued (sync) mails, the `subject` propertie of `$mail` is `null` unless the mail is `build()` first.

Solution provided by bobbybouwmann on: https://laracasts.com/discuss/channels/testing/php-unit-testing-mail


taylorotwell:
This is not necessary when using the currently documented / new version of mailables.

JBtje:
True

Nonetheless, the example as currently provided does not work.
I would suggest removing the non-working part from the example `$mail->hasSubject('...')` as it would only cause confusion.

PR #8330, opened 2022-11-07
Update testing.md

romanbll:
Updating with a solution to easily build end-to-end tests


taylorotwell:
This would be better suited for a blog post.

PR #8333, opened 2022-11-09
add function getOriginal() in eloquent event

DurandSacha:


taylorotwell:
We already document `getOriginal` in another location.

PR #8344, opened 2022-11-16
Update database.md

macdanson:
recommended choice of MySQL is InnoDB which is the default but for some reasons it is the default especcially on local servers


taylorotwell:
Not sure this belongs in our docs.

PR #8346, opened 2022-11-16
Update queues.md

slavicd:
Mention garbage collection issue in the queue section.


taylorotwell:
I will have to think about the best place to document this.

PR #8351, opened 2022-11-18
[9.x] Fix menthod signature

reppair:
This change fixes the `boot()` method signature for registering model observer class trough the `Model::boot()` method (it is a static one).


taylorotwell:
Service providers don't have static boot methods.

reppair:
Ah yeah, I did not read that properly it seems 😅, thought the observer should be registered in the `Eloquent\Model::boot()` method.

PR #8356, opened 2022-11-22
Changing routing password steps

AntonioCarlos850:
<h3>Summary 📰</h3>

Following the documentation to reset password, at step `Handling The Form Submission` on `Requesting The Password Reset Link`, I found a error, the `UrlGenerator` class call the route method passing `password.reset` as parameter, but this route is on next step of the documentation. To fix that i suggest move `Resetting The Password` to the top.

<h3>Relevant logs and/or screenshots 🖼️</h3>

Application log

image


taylorotwell:
I think it's fine.

PR #8360, opened 2022-11-26
[9.x] Adds an example for request missing parameter using an array

fernandokbs:
I have seen some projects using several times 
```php
$request->missing('name')
```
Maybe adding an example that shows you can use an array should help.

```php
$request->missing(['name', 'email'])
```


taylorotwell:
We would need to explain what this actually does. Do all the parameters have to be missing? Only some?

PR #8361, opened 2022-11-28
[9.x] Added information about the injected Swoole's `Server` class

rennokki:
I had some trouble finding out how to access the underlying Swoole server when running Octane, so I found out that the Swoole Server gets booted when the server starts.

Thought the documentation needed this.


taylorotwell:
I may need to find a better section for this. Will look into it.

PR #8370, opened 2022-12-01
Adds replyTo usage in main docs

fylzero:
This tripped me up a bit in a recent Laravel 9 upgrade. Would definitely appreciate it being added to the official docs. Thanks!

### DIFF
--- a/mail.md
+++ b/mail.md
@@ -210,6 +210,22 @@ In addition, you may define a global "reply_to" address within your `config/mail
 
     'reply_to' => ['address' => 'example@example.com', 'name' => 'App Name'],
 
+<a name="using-reply-to"></a>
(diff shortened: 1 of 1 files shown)


taylorotwell:
Added here: https://github.com/laravel/docs/commit/7cbf985bf595e2d4a70ae744554b92ee9b1ef6ac

Thanks

fylzero:
@taylorotwell Verified this in a fresh project.

If replyTo is not placed inside an array and a name is added to the Address, the following error is encountered.

```
Email "Name Of Recipient Here" does not comply with addr-spec of RFC 2822.
```

Not sure if the corrected documentation is wrong or if there is a bug in the implementation. I'll note here and submit a bug in laravel/framework.

PR #8371, opened 2022-12-02
[9.x] Add a note regarding "stream_reads" when using S3 driver

brezzhnev:
I've spent hours trying to figure out why streaming from S3 stopped working when I found out it was due to this flag "stream_reads" after updating our dependencies. 

This may be helpful for other people who want to stream big files to users.


taylorotwell:
We would likely need an actual example of what to set in the configuration and where, etc.

PR #8374, opened 2022-12-06
FIX: add warning to finally block docs

jonnylink:
Docs do not properly warn users of unexpected behavior of finally blocks leading to confusion and frustration.


taylorotwell:
I don't like adding multiple warning blocks. Any further notes should be incorporated into the text itself to make it clearer.

PR #8381, opened 2022-12-08
There is no "box"

ideabrian:
Can we let go of this phrase? "Out of the box, "


ideabrian:
Reading docs is confusing enough. There's no need to write words that don't add any value.

PR #8392, opened 2022-12-11
Add json directive example

MrMohamedAbdallah:
The `@json` directive isn't mentioned in the docs, so I added it.


taylorotwell:
I think Js::from is sufficient.

PR #8400, opened 2022-12-15
[9.x]: Another way to register commands

flavio-paqt:


taylorotwell:
Not sure we really need to show another way?

PR #8406, opened 2022-12-16
Add warning about finally not reliably firing when there's a chain in a batch

jonnylink:
Just trying to document a very common use case where finally will not work


taylorotwell:
I'll update the docs for this now.

jonnylink:
@taylorotwell seems like this won't be merged? It certainly doesn't have to be what I wrote, but there's still no mention of finally not firing if a chain has a failed job in it. 

Speaking of my team, this lead to a production bug because we assumed it would function the same way `finally` does everywhere else (i.e. it is _always_ executed).

PR #8407, opened 2022-12-16
Add commands index, delete-index and delete-all-indexes

kichetof:


taylorotwell:
Need more information regarding when this is even necessary. I know on some engines you do not need to create an index manually, so this could be confusing in those cases. Please resubmit when that is fixed. 👍

PR #8408, opened 2022-12-18
Add mention that date/time where functions won't use indexes

SilvioDoMine:
To avoid repeat myself, here is why this is needed: <https://github.com/laravel/docs/pull/7182#issuecomment-1356801051>

### DIFF
--- a/queries.md
+++ b/queries.md
@@ -609,6 +609,9 @@ The `whereTime` method may be used to compare a column's value against a specifi
                     ->whereTime('created_at', '=', '11:20:45')
                     ->get();
 
+> **Warning**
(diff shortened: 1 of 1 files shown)


taylorotwell:
You saw the SQL generated by Laravel is the exact same as what you pasted into your DB program that showed indexes being used. As to why they weren't using when the query was issued by Eloquent, I don't know - the query is the same and Laravel doesn't have some "disable indexes" feature. Need more proof of an issue and an actual explanation for WHY there is an issue before merging this.

oliworx:
It is not the exact same query, Eloquent creates a prepared statement, the other example is not.

PR #8411, opened 2022-12-19
Restoring Redis::funnel doc

erickneverson:
Redis::funnel was mentioned previously in v 5.8.  I'd like to restore it here as it's a useful feature to funnel jobs instead of throttling.


taylorotwell:
I think this is maybe solved more generically by cache locks? 

https://laravel.com/docs/9.x/cache#atomic-locks

PR #8414, opened 2022-12-23
Updating index to unique

Beefsupreme21:
The example provided did not work, but changing index to unique fixed the issue. 

I think the following line needs to be updated from:

    $table->index(['account_id', 'created_at']);

Proposed change: 

    $table->unique(['account_id', 'created_at']);


taylorotwell:
Works fine for me. No details provided in your PR.

PR #8415, opened 2022-12-23
Add zero $tries meaning infinite retries to queue docs

blackshadev:
I stumbled on this edge case (of feature) in laravel while working on a client code base and couldn't find any documentation for it. 

Apparently a job with `0` $tries will be retried indefinitely. What is more, I tried defining `0` on the job and `1` on the worker, because if `0` meant infinite then,  `1` attempt  would be more specific to `infinite` attempts right? But the laravel workers doesn't check for specificity, it is the default if the Job didn't specify anything. 

The word `specific` threw me off, because what does it mean to be more specific? In this case it meant: The job as a `$tries` set, which means it is more granular and thus more specific. While I read it as, a `$tries` value which is more specific, and thus the least amount of `attempts` .

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -996,7 +996,7 @@ Alternatively, you may specify the job's connection by calling the `onConnection
 
 If one of your queued jobs is encountering an error, you likely do not want it to keep retrying indefinitely. Therefore, Laravel provides various ways to specify how many times or for how long a job may be attempted.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not sure if this warrants a big red note. Can probably be integrated into the text better.

blackshadev:
@taylorotwell do you expect any changes or are you going to change it yourself. I can propose an alternative text, but your direct closing of this PR seems to indicate you have a different plan for it.

Do you at least agree this edge case of 0 being treated differently should be mentioned in the docs?

PR #8420, opened 2022-12-24
[9.x] Add note to observer after commit

mikecummingswf:
It may be helpful to note that the methods available for model state examination will not return the original values if `$afterCommit=true`.

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -1630,6 +1630,9 @@ When models are being created within a database transaction, you may want to ins
             //
         }
     }
+	
+> **Note**  
+> If you choose to handle events after transactions are committed, you will not be able to examine attribute changes, such as by using `getOriginal`, in the observer event methods.
 
 <a name="muting-events"></a>
 ### Muting Events


taylorotwell:
This feels a bit vague. Can you give examples?

mikecummingswf:
Sure thing. One example would be: 

```
$user = User::create(['name' => 'Mike']);

$user->name = 'Nick';
$user->save();
```

```
// UserObserver
$afterCommit = false;
public function updating(User $user)
{
  dd($user->getOriginal('name')); // Returns 'Mike'
}
```

```
// UserObserver
$afterCommit = true;
public function updating(User $user)
{
  dd($user->getOriginal('name')); // Returns 'Nick'
}
```

taylorotwell:
Yeah, I'm not sure afterCommit makes sense in combination with `updating` though? `updating` would normally be invoked before the update has actually been done. However, when combined when after commit it is basically the same as the `updated` event, which takes place after the saving has already been done.

mikecummingswf:
Perhaps the `updated` method would have been a better example. This scenario holds true for `updated`, `created`, and `saved` as well.

PR #8423, opened 2022-12-27
Set language to shell in code blocks for shell commands

lazypolymath:


taylorotwell:
Some of these look better as nothing.

PR #8424, opened 2022-12-28
[9.x] Add "Replace" instruction for session.md

mohammad-kasiri:
Add "Replace" instruction in document that is use for rewrite session.


taylorotwell:
I think it's fine to just document `put`. I don't like to document a lot of ways to do the same thing.

PR #8430, opened 2023-01-01
Fix method name booted to boot

azabolo:


nunomaduro:
Thank you for the suggestion. But for now, we don't have plans to change this method name.

PR #8432, opened 2023-01-02
[9.x] Add "Seeders I/O" section to "Database: Seeding" docs

renoirtech:
Inspired by [How to echo to console in Laravel and Artisan?](https://stackoverflow.com/questions/16733957/how-to-echo-to-console-in-laravel-and-artisan)


taylorotwell:
Personally don't think it's that great of an idea to have prompts in your seeds.

PR #8441, opened 2023-01-06
[9.x] Clarify redis tagged cache items limitations

clemblanco:
Trying to clarify what is happening with expiring tagged cache items using the `redis` driver as this seems to trip people when using tags on Redis.

See https://github.com/laravel/framework/issues/43850

### DIFF
--- a/cache.md
+++ b/cache.md
@@ -266,7 +266,7 @@ When the `cache` function is called without any arguments, it returns an instanc
 ## Cache Tags
 
 > **Warning**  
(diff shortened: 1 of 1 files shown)


taylorotwell:
Going to see if we can improve our Redis implementation first.

clemblanco:
Ah, nice even better. I think https://github.com/tenantcloud/laravel-better-cache is a good place to start.

Funny enough, I looked into switching to `memcached` as the docs seems to say it's more performant for tags but I could still see some keys being left behind when tagged items have expired.

PR #8443, opened 2023-01-07
fix table overflow

hichemfantar:
This should fix the problem where tables sometimes overflow and cause the entire page to become horizontally scrollable.


taylorotwell:
Please submit with before and after pictures and device information.

hichemfantar:
@taylorotwell done

PR #8449, opened 2023-01-10
Note about `$wrap` not applying to resource collections

kohenkatz:
The `$wrap` value from the individual resource does not apply when calling the `JsonResource::collection()` method (because that method returns an instance of `AnonymousResourceCollection`). This should be noted in the documentation.


taylorotwell:
I feel like this is maybe something better suited to a blog post. It's hard to follow as a warning block.

PR #8450, opened 2023-01-11
Clarify how the SerializesModels trait works

etlam:
Because of the paragraph in the section it is not 100% clear, that only the identifier for the model will be serialized onto the queue.

I talked to three people and all of them got it wrong:
They told me that we have to use the trait if we want to get all properties and relations serialized and if we do not use it, it would only serialize the identifier for the model and refresh it from database.

Therefore, I would propose this change.


taylorotwell:
I don't think this change is really necessary. Thanks.

etlam:
You are right. It is not necessary, but it would improve the docs, if three of three asked people do not understand it the right way.  The name of the Trait also does not show that only the identifier is going to be serialized.

PR #8451, opened 2023-01-11
[10.x] Adds introductions about Composer's Minimum Stability

nunomaduro:
This pull request adds introductions about Composer's Minimum Stability for Laravel 10.


u01jmg3:
I too am curious. Also, as the [default setting for `minimum-stability` is `stable`](https://getcomposer.org/doc/04-schema.md#minimum-stability), it might be better to have nothing.

PR #8455, opened 2023-01-16
[9.x] Drop an index when modifying a column

hafezdivandari:
* laravel/framework#45513


taylorotwell:
Honestly I don't like documenting too many ways of doing the same thing.

PR #8459, opened 2023-01-17
add "mysql" to list of available features

browner12:
MySQL is now a "feature", so we'll add it to the list here.

ping @svpernova09


svpernova09 (commented):
Looks good, Thanks!

PR #8482, opened 2023-01-27
[9.x] Adds doc for the `PendingCommand@execute`

sethsandaru:
Related to this https://github.com/laravel/framework/issues/45811, the method needs to be documented so it won't confuse people

### DIFF
--- a/console-tests.md
+++ b/console-tests.md
@@ -34,6 +34,19 @@ Of course, all terminal commands typically exit with a status code of `0` when t
 
     $this->artisan('inspire')->assertFailed();
 
+### Manually Invoking The Command
+
+By default, the `artisan` method won't run the command immediately. 
+This can cause an error if you try to do some data assertions.
+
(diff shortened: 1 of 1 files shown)


sethsandaru:
@taylorotwell @driesvints please lemme know if I need to update the text 😄

taylorotwell:
I don't think this is accurate? For example, the following test passes on a empty, new Laravel project:

```php
$this->artisan('inspire');

$this->assertDatabaseMissing('users', ['email' => 'taylor@laravel.com']);
```

sethsandaru:
@taylorotwell assertDatabaseMissing would work normally fine because there won't be any insertion happened 🤔

sethsandaru:
So that's why we had the issue reported and this is the solution https://github.com/laravel/framework/issues/45811#issuecomment-1404767567

IMO it is fine to add the doc for `execute` too, since there is no place mentioned that (unless you are using PHPStorm and figure the method yourself)

PR #8484, opened 2023-01-27
[9.x] Adds doc for the `PendingCommand@execute`

sethsandaru:
Related to this https://github.com/laravel/framework/issues/45811, the method needs to be documented so it won't confuse people 

Resubmitting this, I added my answers too (check out the post below)

### DIFF
--- a/console-tests.md
+++ b/console-tests.md
@@ -34,6 +34,19 @@ Of course, all terminal commands typically exit with a status code of `0` when t
 
     $this->artisan('inspire')->assertFailed();
 
+### Manually Invoking The Command
+
+By default, the `artisan` method won't run the command immediately. 
+This can cause an error if you try to do some data assertions.
+
+In order to run the command, you need to invoke the `execute` method:
+
+    $this->artisan('users:create taylor@laravel.com')
+        ->assertSuccessful()
+        ->execute();
+
+    $this->assertDatabaseHas('users', ['email' => 'taylor@laravel.com']);
+
 <a name="input-output-expectations"></a>
 ## Input / Output Expectations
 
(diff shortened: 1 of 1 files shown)


sethsandaru:
Hi @taylorotwell , I added my answers here:

- https://github.com/laravel/docs/pull/8482#issuecomment-1406746295
- https://github.com/laravel/docs/pull/8482#issuecomment-1406747752

Lemme know if it makes sense

taylorotwell:
Please share your entire test.

sethsandaru:
@taylorotwell yeah sure, so for example, we had a command to create the user (`users:create email`)

If we do like this:

```php
$this->artisan('users:create taylor@laravel.com');

// this assert will fail
$this->assertDatabaseHas('users', ['email' => 'taylor@laravel.com']);
```

The flow of the code above would look like this: Create PendingCommand => Data assertion(s) - fail here => Execute Command

The assertion will always fail, because the command hasn't been executed yet (artisan will return a PendingCommand and only run when destructing the object, ref: https://github.com/laravel/framework/blob/9.x/src/Illuminate/Testing/PendingCommand.php#L475-L482).

So in order to run the command and then do some data assertion, we need to manually invoke the `execute` method (and we are lacking this in the documentation)

```php
$this->artisan('users:create taylor@laravel.com')->execute();

// this assert will pass
$this->assertDatabaseHas('users', ['email' => 'taylor@laravel.com']);
```

=> Create PendingCommand => Execute Command => Data assertions

sethsandaru:
cc @driesvints

taylorotwell:
Can you just literally paste the entire test method unmodified and with no explanation?

sethsandaru:
@taylorotwell yeah it works like that but won't if we assign `artisan` to a variable, sorries for the confusion above.

```php
$artisan = $this->artisan('users:create taylor@laravel.com');

// errors
```

I just changed the whole section into a simple warning, lemme know if it makes sense.

PR #8490, opened 2023-01-30
[8.x] Add notice pointing users to new Ably Laravel broadcaster

owenpearson:
Hey,
As [suggested by Taylor](https://github.com/laravel/framework/pull/43685#issuecomment-1267103189) we’ve recently released our own [laravel-broadcaster](https://github.com/ably/laravel-broadcaster) plugin, which we expect to be much more stable and reliable than the currently documented method of using Ably with laravel/echo and the Ably Pusher adapter. This PR (and the parallel PRs into other docs versions) updates the docs to let users know that this is the new recommended way to use Ably with Laravel. Please let us know if you have any questions or feedback :)

### DIFF
--- a/broadcasting.md
+++ b/broadcasting.md
@@ -108,6 +108,12 @@ The [laravel-websockets](https://github.com/beyondcode/laravel-websockets) and [
 <a name="ably"></a>
 ### Ably
 
+> **Important Note** 
+>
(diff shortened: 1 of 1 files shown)


taylorotwell:
We'll likely only merge docs into the current releases.

PR #8492, opened 2023-01-30
[10.x] Add documentation for new Ably Laravel broadcaster 

owenpearson:
Hey,
As [suggested by Taylor](https://github.com/laravel/framework/pull/43685#issuecomment-1267103189) we’ve recently released our own [laravel-broadcaster](https://github.com/ably/laravel-broadcaster) plugin, which we expect to be much more stable and reliable than the currently documented method of using Ably with laravel/echo and the Ably Pusher adapter. This PR (and the parallel PRs into other docs versions) updates the docs to let users know that this is the new recommended way to use Ably with Laravel. Please let us know if you have any questions or feedback :)

### DIFF
--- a/authentication.md
+++ b/authentication.md
@@ -420,11 +420,11 @@ You may also use HTTP Basic Authentication without setting a user identifier coo
 
     }
 
-Next, [register the route middleware](/docs/{{version}}/middleware#registering-middleware) and attach it to a route:
+Next, attach the middleware to a route:
 
     Route::get('/api/user', function () {
         // Only authenticated users may access this route...
-    })->middleware('auth.basic.once');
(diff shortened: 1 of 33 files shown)


taylorotwell:
Hey there! 👋

So my goal here was for Ably to maintain all related documentation to the Ably maintained drivers and I would simply link to them with a notice. I have added that notice here: https://github.com/laravel/docs/commit/3ebb71766688652de74cd1d5a4440a523cc39544

If I should fix anything about that note let me know - here is what it looks like:

CleanShot 2023-02-10 at 16 40 15@2x


Thank you for your work on maintaining this broadcaster and Echo client!

PR #8499, opened 2023-02-01
Add methods to collection docs

xurshudyan:
Docs for PR: https://github.com/laravel/framework/pull/45839


taylorotwell:
I honestly cannot wrap my head around what these methods even do or when I would use them. I feel like we need a more real-world, longer explanation for why they might be useful to end users.

PR #8505, opened 2023-02-03
[9.x] Add warning for request input default

bert-w:
This warning describes the usage of `$request->input('someinput') ?? 'somedefault'` vs `$request->input('someinput', 'somedefault')`. The former is a good alternative if you want empty input arguments to fall back to the given default.


taylorotwell:
Not sure this warrants documentation.

PR #8512, opened 2023-02-06
Fix `withoutOverlapping()` docs

tillkruss:
It's seconds, not minutes. https://github.com/laravel/framework/issues/45952#issuecomment-1418687326


decadence:
Then it's too small value for the default lock? If someone uses it on command that runs 30 minutes this method will not work for some cases (if another command starts on 25 minute).

tillkruss:
> Then it's too small value for the default lock? If someone uses it on command that runs 30 minutes this method will not work for some cases (if another command starts on 25 minute).

It's a weird default. I expected it to be 24 hours.

decadence:
I bet it was 24 hours but forgotten when Laravel went to seconds from minutes in Cache time

taylorotwell:
Eh, it appears appears to be minutes to me? We multiply that value by 60 when passing to the cache layer, giving 86,400 seconds - which is 24 hours.

CleanShot 2023-02-06 at 14 24 50@2x

PR #8516, opened 2023-02-07
Update migrations.md

Sairahcaz:
As mentioned [here](https://github.com/laravel/framework/issues/5517) likely will maybe never support composite primary keys. I thought it would be great if that immediately would be pointed out at the documentation. I had some troubles with that and it confused me a bit, that the Schema Builder supports composite primary keys but not Eloquent.


taylorotwell:
Hmm - don't think is totally necessary atm. Thanks though!

PR #8534, opened 2023-02-09
Clarify MySQL credentials

jasonmccreary:
As a new user to Sail (and Docker) it wasn't clear to me the credentials I should use when connecting to MySQL via TablePlus. While not a leap to use the values from my `DB_USERNAME` and `DB_PASSWORD` environment variables, I received "access denied" errors when working with the tables.

This additional sentence aims to clarify where the credentials come from and which to use when connecting to the MySQL instance directly.


taylorotwell:
Thanks 👍

PR #8536, opened 2023-02-09
[10.x] Use promoted properties

browner12:
using promoted properties makes the docs more succinct.

I did not switch to promoted properties when the properties were not explicitly defined in the code already.


driesvints (approved):

ankurk91:
Return types must be updated in docs too. Like Skelton

PR #8542, opened 2023-02-10
[10.x ] Note new type-hints for Cast contracts

jasonmccreary:
There are a few more of these contract changes for Validation Rules. I'm starting with these to ensure this is something you want to add and then follow your formatting for the others.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -38,6 +38,7 @@
 - [The `Redirect::home` Method](#redirect-home)
 - [The `Bus::dispatchNow` Method](#dispatch-now)
 - [ULID Columns](#ulid-columns)
+- [Contract type hints](#contract-type-hints)
 
 </div>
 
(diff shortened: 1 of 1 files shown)


nunomaduro:
@taylorotwell For info, this change is optional. People don't have to change their applications. So, not sure if this should stay on the upgrade guide.

jasonmccreary:
@nunomaduro, how is it optional? My understanding is if you implement an interface, you have to implement its type hints.

nunomaduro:
It's called Covariance and Contravariance: https://www.php.net/manual/en/language.oop5.variance.php.

```php
interface CastsAttributes {
    function set(object $object, mixed $value);
}

class LowerCase implements CastsAttributes {
    function set($object, $value) {
        var_dump($object, $value);
    }
}

(new LowerCase)->set(new stdClass, 'value');
// object(stdClass) {},  string(5) "value"
```

jasonmccreary:
Interesting, yeah, seems to only be if you define a type hint, then it needs to match. Otherwise, you can leave it off. Except for return types.

Oh, PHP.

PR #8557, opened 2023-02-13
[10.x] Removes named parameters

timacdonald:
As Laravel doesn't support named parameters, feels like we are sending mixed messages by documenting their usage.

Proposing we remove them from the docs.

Maybe this could be improved with a named constructor `Response::empty(403)`.

### DIFF
--- a/http-client.md
+++ b/http-client.md
@@ -571,7 +571,7 @@ You may use the `recorded` method to gather all requests and their corresponding
 
 ```php
 Http::fake([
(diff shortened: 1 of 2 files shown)


ankurk91:
The envelope mail docs is using named arguments already

taylorotwell:
There are certain named parameters I'm pretty committed to not changing. This is one. 👍

timacdonald:
@ankurk91 the envelope class explicitly opts into named parameters, so that makes sense there.

PR #8581, opened 2023-02-15
[9.x] Mail: add warning about alwaysTo disabling cc/bcc

VivienLN:
This little warning would have saved me some time. 

Didn't seem obvious to me and the only mentions I found about it were some old stackoverflow threads. I can't be the only one.

Tested in 9.x but it may also applies to laravel 10


taylorotwell:
Wrong branch - should be sent to `10.x`

PR #8582, opened 2023-02-15
Change Monolog 3 bump impact to medium and explain about 3rd-party se…

jonnott:
…rvice packages implication


browner12:
lol, when Bugsnag actually gets their Laravel 10 support released.

jonnott:
> lol, when Bugsnag actually gets their Laravel 10 support released.

Not to mention Rollbar..!

PR #8583, opened 2023-02-15
Add information about `shouldRender` method

trovster:
It seems quite a few people overlook the `shouldRender()` method, so I have added to the example of a component to improve visibility.


taylorotwell:
We would have to explain what it actually does.

decadence:
Thanks. Didn't know this exists.

PR #8584, opened 2023-02-15
[10.x] upgrade doc updated for removed registerPolicies method

ahmett:
This PR addresses laravel/framework#46132


u01jmg3:
@ahmett: it's not clear from the docs whether the method is called `registerPolicy` or `registerPolicies`. (I know `registerPolicies` is correct.)

PR #8598, opened 2023-02-19
Added PHP extensions to connect to a database

Verhoeckx:
For Laravel to work properly, you need a PHP extension to connect to a database. These extensions were missing in the documentation.


taylorotwell:
Honestly I think probably 99% of PHP installations include these things.

Verhoeckx:
> Honestly I think probably 99% of PHP installations include these things.

Not when you are working with containers! Then they are not automatically included.

PR #8610, opened 2023-02-22
[10.x] Add set column for ```Soft Deletes```

milwad-dev:


taylorotwell:
I think it's fine to not customize this.

PR #8618, opened 2023-02-25
we have to add @endsection instead of @show

jemmeli:
we have syntaxe error in the documentation add @endsection instead of @show


jemmeli:
syntaxe error add @endsection instead of @show

ankurk91:
Good catch, interestingly `@show` is working fine and does not generate any error

taylorotwell:
This is valid and necessary syntax. @show is a real concept in Blade.

PR #8619, opened 2023-02-25
chain ->getValue(DB::connection()->getQueryGrammar()) to DB::raw()

jonnott:
..calls in the docs, as per upgrade guide

### DIFF
--- a/queries.md
+++ b/queries.md
@@ -247,13 +247,13 @@ If you already have a query builder instance and you wish to add a column to its
 Sometimes you may need to insert an arbitrary string into a query. To create a raw string expression, you may use the `raw` method provided by the `DB` facade:
 
     $users = DB::table('users')
-                 ->select(DB::raw('count(*) as user_count, status'))
+                 ->select(DB::raw('count(*) as user_count, status')->getValue(DB::connection()->getQueryGrammar()))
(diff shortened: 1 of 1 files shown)


jonnott:
Hmmm .. after debugging this in my own apps, I think I've got the wrong end of the stick here .. I think (correct me if I'm wrong), you only need getValue() if you're adding the output of an expression into some of raw SQL or other string. If it's going straight into e.g. a select() or addSelect(), you can still just use plain DB::raw() ?

jonnott:
"Database "expressions" (typically generated via DB::raw) have been rewritten in Laravel 10.x to offer additional functionality in the future."

@driesvints @taylorotwell Can we get a sneak of what that functionality is?

driesvints:
Think @tpetry worked on this?

tpetry:
@jonnott You are getting the idea for them from my pr: [[10.x] database expressions with grammar-specific formatting](https://github.com/laravel/framework/pull/44784)

I am right now working on [tpetry/laravel-query-expressions](https://github.com/tpetry/laravel-query-expressions) to provide a lot of ready-made expressions. But its a lot of work to identify all the changes of different database versions and finding ways to harmonize their behaviour to work exactly the same on all of them.

PR #8620, opened 2023-02-26
[10.x] Adds more examples to time manipulation

DarkGhostHunter:
I thought the Time manipulation of the _Mocking_ section of testing could have more details on what is offered, and why you may use it.

### DIFF
--- a/mocking.md
+++ b/mocking.md
@@ -134,7 +134,7 @@ If you would like to [spy](http://docs.mockery.io/en/latest/reference/spies.html
 <a name="interacting-with-time"></a>
 ## Interacting With Time
 
-When testing, you may occasionally need to modify the time returned by helpers such as `now` or `Illuminate\Support\Carbon::now()`, or some of Laravel's generated timestamps like when creating Eloquent Models. Thankfully, Laravel's base feature test class includes helpers that allow you to manipulate the current time.
+When testing, you may occasionally need to modify the time returned by helpers such as `now` or `Illuminate\Support\Carbon::now()`, or some of Laravel's generated timestamps like on Eloquent Models. Thankfully, Laravel's base feature test class includes helpers that allow you to manipulate the current time.
 
 The `travel()` method can be used to freeze the time into the future or the past, and resume the current flow of time using `back()`.
 
--- a/mocking.md
+++ b/mocking.md
@@ -155,7 +155,7 @@ Alternatively, you may execute a callback while the time is frozen, like traveli
     $this->travel(5)->days(function () {
         // Test something five days into the future...
     });
-
+    
     $this->travelTo(now()->subDays(10), function () {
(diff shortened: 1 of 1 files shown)


mocking.md
```diff
@@ -134,32 +134,76 @@ If you would like to [spy](http://docs.mockery.io/en/latest/reference/spies.html
 <a name="interacting-with-time"></a>
 ## Interacting With Time
 
-When testing, you may occasionally need to modify the time returned by helpers such as `now` or `Illuminate\Support\Carbon::now()`. Thankfully, Laravel's base feature test class includes helpers that allow you to manipulate the current time:
+When testing, you may occasionally need to modify the time returned by helpers such as `now` or `Illuminate\Support\Carbon::now()`, or some of Laravel's generated timestamps like on Eloquent Models. Thankfully, Laravel's base feature test class includes helpers that allow you to manipulate the current time.
+
+The `travel()` method can be used to freeze the time into the future or the past, and resume the current flow of time using `back()`.
```
dammy001, line 139:
```suggestion
The `travel()` method can be used to freeze time into the future or the past, and resume the current flow of time using `back()`.
```

mocking.md
```diff
@@ -134,32 +134,76 @@ If you would like to [spy](http://docs.mockery.io/en/latest/reference/spies.html
...
-        $this->travel(5)->hours();
-        $this->travel(5)->days();
-        $this->travel(5)->weeks();
-        $this->travel(5)->years();
-
-        // Freeze time and resume normal time after executing closure...
-        $this->freezeTime(function (Carbon $time) {
-            // ...
-        });
+To give a concrete example, imagine that forums threads save internally their moment of last activity. We need to test that a thread becomes locked for inactivity after one week. We can move the time forward after the thread is created to test that.
```
dammy001, line 177:
```suggestion
To give a concrete example, imagine that forum threads save internally their moment of last activity. We need to test that a thread becomes locked for inactivity after one week. We can move the time forward after the thread is created to test that.
```

mocking.md
```diff
@@ -134,32 +134,76 @@ If you would like to [spy](http://docs.mockery.io/en/latest/reference/spies.html
...
+        $thread = Thread::factory()->create();
+        
+        $this->travel(1)->week();
+        
+        $this->assertTrue($thread->isLockedByInactivity());
+    }
 
-        // Travel to an explicit time...
-        $this->travelTo(now()->subHours(6));
+Imagine that the forum thread receives a reply three days after it was created. Logically, the thread would become locked after a week of that last reply. That reply can be created in using the convenient _wormhole_ callback.
```
dammy001, line 190:
```suggestion
Imagine that the forum thread receives a reply three days after it was created. Logically, the thread would become locked after a week of that last reply. That reply can be created using the convenient _wormhole_ callback.
```

PR #8622, opened 2023-02-27
[10.x] Add pennant helper

milwad-dev:
@timacdonald


taylorotwell:
I think the `feature` helper is actually more useful for checking features using it's shorthand syntax.

milwad-dev:
We can change how to works of `feature` helper in docs.

PR #8626, opened 2023-03-02
[10.x] Add `isMatch` method in helpers

milwad-dev:
Is related to this PR https://github.com/laravel/framework/pull/46303


taylorotwell:
Doesn't include the stringable example and isn't in the right order.

PR #8630, opened 2023-03-04
Document sync driver not supporting after_commit

dciprian-petrisor:
https://github.com/laravel/framework/issues/46296

Currently, the ```after_commit``` option is silently ignored by the ```sync``` queue driver, which may cause unexpected behaviour for developers.

This PR attempts to document that the ```after_commit``` option is not  supported, to avoid confusion.


taylorotwell:
I never put two "warning" blocks directly next to each other.

dciprian-petrisor:
The one above is a note, not a warning.

Suggestions on placement/message is welcome, I'd make the required changes.

PR #8634, opened 2023-03-06
[10.x] Add find and findMany methods

kevinb1989:


taylorotwell:
I'll try to work it in.

PR #8635, opened 2023-03-08
[10.x] Add Change Error Theme into `Error-Handling`

milwad-dev:
Suppose you want to change the theme of the error pages!
It is not written in any part of the documentation.
This section is a good place for those who want to change the theme of the error page.


taylorotwell:
This is probably better suited here: https://github.com/spatie/ignition

milwad-dev:
We are using Laravel!
When we want to change this error, many may not know where and which package to find.

PR #8637, opened 2023-03-08
[10.x] Add replace_placeholders and processors to logging

alanpoulain:
Related: https://github.com/laravel/framework/pull/46344 and https://github.com/laravel/laravel/pull/6139

Placeholders usage should be shown by default in the documentation.

See also @Crell's post: https://peakd.com/hive-168588/@crell/using-psr-3-placeholders-properly.


Crell:
Shouldn't it be left open until then, so it doesn't get lost/forgotten?

PR #8640, opened 2023-03-09
Pennant feature helper

sanskar-mk2:
I was quickly looking to get rich value of a feature in blade.
I found out that the pennant codebase has an undocumented helper method.

Let me know how I can improve on this if needed. Thank you.


taylorotwell:
I think we would likely want examples of all uses cases.

PR #8654, opened 2023-03-15
Remove duplicate service provider registration

hopkins385:
Hi guys, I just noticed there is a duplicate in the AppServiceProvider register method for the Telescope Service Provider. As I know, there is just one Telescope Service Provider, but there are two ways how to resolve the class (either by passing the full path into the closure or by just entering the classname and make use of the "use" statement). So I guess what happened here is that both ways have been added to the doc, but it does confuse, because as per current documentation the service provider would be registered twice? .. Anyways, please let me know if I miss something.

### DIFF
--- a/telescope.md
+++ b/telescope.md
@@ -74,7 +74,7 @@ php artisan telescope:install
 php artisan migrate
 ```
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Two different service providers. The one in the core of Telescope and your app level Telescope one where you place your configuration.

PR #8659, opened 2023-03-16
[10.x] Fix Missing Config Params on SFTP Driver

praxxys:
This will save people a few hours of searching as setting file and directory permissions for the SFTP driver is not well documented. This is particularly useful for people upgrading from SFTP v2 as config params have changed in SFTP v3.

Hope this helps.

### DIFF
--- a/filesystem.md
+++ b/filesystem.md
@@ -135,6 +135,7 @@ Laravel's Flysystem integrations work great with SFTP; however, a sample configu
         'privateKey' => env('SFTP_PRIVATE_KEY'),
         'passphrase' => env('SFTP_PASSPHRASE'),
 
+        // Settings for file/directory permissions, defaults to `private`
         'visibility' => 'public', // `private` = 0600, `public` = 0700
         'directory_visibility' => 'public', // `private` = 0700, `public` = 0755
 
--- a/filesystem.md
+++ b/filesystem.md
@@ -135,9 +135,9 @@ Laravel's Flysystem integrations work great with SFTP; however, a sample configu
         'privateKey' => env('SFTP_PRIVATE_KEY'),
         'passphrase' => env('SFTP_PASSPHRASE'),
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
If these options aren't set - what are the defaults?

praxxys:
Hi @taylorotwell, currently they default to `private` which is `0600` for files and `0700` for directories. I can revise to make it clear. Or do you think we should add a sub section for `File/Directory Permissions` to declutter?

praxxys:
Additional reference:

https://github.com/thephpleague/flysystem/blob/b01bb3920e90c9cc6c85ee08af8e7b98e1249e47/src/UnixVisibility/PortableVisibilityConverter.php#L21
```
    public function forFile(string $visibility): int
    {
        PortableVisibilityGuard::guardAgainstInvalidInput($visibility);

        return $visibility === Visibility::PUBLIC <------ if `$visibility` is unset, will return `private`
            ? $this->filePublic
            : $this->filePrivate;
    }

    public function forDirectory(string $visibility): int
    {
        PortableVisibilityGuard::guardAgainstInvalidInput($visibility);

        return $visibility === Visibility::PUBLIC <------ if `$visibility` is unset, will return `private`
            ? $this->directoryPublic
            : $this->directoryPrivate;
    }
```

Stack trace:

[upload()](https://github.com/thephpleague/flysystem-sftp-v3/blob/e8d38bda3d5ccffa78bf13b807ed2055c99982c4/SftpAdapter.php#L89)  ->  [setVisibility()](https://github.com/thephpleague/flysystem-sftp-v3/blob/e8d38bda3d5ccffa78bf13b807ed2055c99982c4/SftpAdapter.php#L196https://github.com/thephpleague/flysystem-sftp-v3/blob/e8d38bda3d5ccffa78bf13b807ed2055c99982c4/SftpAdapter.php#L196)

taylorotwell:
Thanks

PR #8662, opened 2023-03-16
controller argument names matter in case of route model binding, so m…

CurvesTech:
controller argument names matter in case of route model binding, so mentioned that exceptional case.


taylorotwell:
No plans to add this.

PR #8665, opened 2023-03-16
separated the strange looking directive from @class

CurvesTech:
image


taylorotwell:
I don't really think it's a big deal.

PR #8670, opened 2023-03-17
Document how to customize pagination information for resources

mortenscheel:
A Google search on "how to customize eloquent resource pagination metadata" shows that I'm not the only one who wants to know how to do this. Unfortunately none of the results provide the correct answer.

After a bit of digging, I discovered that [PaginatedResourceResponse](https://github.com/laravel/framework/blob/136ed754f2d80e5c7cd407a4262eb2eac478e7e8/src/Illuminate/Http/Resources/Json/PaginatedResourceResponse.php#L53-L56) allows you to override the default result. And as far as I can tell, this feature hasn't been documented.

### DIFF
--- a/eloquent-resources.md
+++ b/eloquent-resources.md
@@ -491,6 +491,24 @@ Paginated responses always contain `meta` and `links` keys with information abou
 }
 ```
 
+#### Customizing pagination information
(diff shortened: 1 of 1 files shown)


mortenscheel:
@taylorotwell you should probably delete the last line too, since the example is different.

PR #8680, opened 2023-03-24
Add missing ->deleteFileAfterSend(true) example

jozeflambrecht:
Was documented in 5.x, but somehow disappeared


taylorotwell:
I would prefer if we just renamed this method in our own extended class tbh... something like `thenDeleteFile`.

PR #8687, opened 2023-03-28
Document `@js` blade directive

CalebDW:
Hello!

This PR documents the [`@js` blade directive](https://github.com/laravel/framework/pull/39522) which was introduced in 8.x.

This should probably be merged into the `8.x`, `9.x`, `10.x`, and `master?` branches, but I'm not sure how y'all handle that...should I create a PR for each branch?

Thanks!

### DIFF
--- a/blade.md
+++ b/blade.md
@@ -170,8 +170,18 @@ The latest versions of the Laravel application skeleton include a `Js` facade, w
 </script>
 ```
 
+##### The `@js` Directive
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
We already document how to deal with JSON.

CalebDW:
So is the `@js` blade directive deprecated then?

Not sure why you'd want undocumented features unless they're going to be removed...

PR #8691, opened 2023-03-31
withPivotValue docs

iamgergo:
As I mentioned https://github.com/laravel/framework/pull/46559 here, this method is not documented.


taylorotwell:
I'm not sure this really explains what it does? Even I would not be sure I would understand the purpose of this method if I read this.

PR #8693, opened 2023-03-31
Minor Clarification on Eloquent Factory Creation

HeinHanekom:
PR adds an indication that setting the $model is not required when normal naming conventions are being followed as Laravel is able to resolve the model class based on the name of the factory class.


taylorotwell:
I think it pretty much says that in the text above. "If these conventions don't work for ..."

PR #8699, opened 2023-04-04
[10.x] Add multiple connection monitor in `queues`

milwad-dev:


taylorotwell:
I think that is fairly obvious from the example further up the page.

PR #8701, opened 2023-04-05
[10.x] Add Chipper CI to Dusk docs

fideloper:
Dusk is very commonly used in Chipper CI. We have [docs](https://chipperci.com/docs/testing/laravel-dusk-new/) on using Dusk, but it would be very handy if the official docs mentioned Chipper as well!


fideloper:
TY for the edits!

PR #8703, opened 2023-04-06
[10.x] Update eloquent.md updateOrCreate 

aidancrane:
This may be the worlds smallest change but I just wanted to give a hint that the `updateOrCreate` function uses `->fill()` rather than `->create()`. I know this is covered in the section above https://laravel.com/docs/master/eloquent#mass-assignment-exceptions but it would potentially save others from looking at the API docs.

I was confused when working on my model that values were not being saved because the attributes were not fillable.

https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Eloquent/Builder.php#L578


taylorotwell:
No plans to change this.

PR #8707, opened 2023-04-07
[10.x] Adjust count macro in Scout

joaopalopes24:
See https://github.com/laravel/scout/issues/717

Thank you @driesvints

### DIFF
--- a/scout.md
+++ b/scout.md
@@ -717,9 +717,9 @@ If you would like to define a custom Scout search builder method, you may use th
     public function boot(): void
     {
         Builder::macro('count', function () {
-            return $this->engine()->getTotalCount(
-                $this->engine()->search($this)
-            );
+            $this->limit = 10;
+
+            return $this->engine()->getTotalCount($this->raw());
         });
     }
 


scout.md
```diff
@@ -717,9 +717,9 @@ If you would like to define a custom Scout search builder method, you may use th
     public function boot(): void
     {
         Builder::macro('count', function () {
-            return $this->engine()->getTotalCount(
-                $this->engine()->search($this)
-            );
+            $this->limit = 10;
+
+            return $this->engine()->getTotalCount($this->raw());
```
driesvints, line 722: This doesn't seem correct? The previous statement of ->search should remain.

I also think it should be noted that this only needs to be done for Meilisearch.

joaopalopes24: Method `raw` already does the search as it was before.

See:

```
/**
 * Get the raw results of the search.
 *
 * @return mixed
 */
public function raw()
{
    return $this->engine()->search($this);
}
```

driesvints: Gotcha 👍

joaopalopes24: I transformed this:

```
return $this->engine()->getTotalCount(
    $this->engine()->search($this);
);
```

On this:

`return $this->engine()->getTotalCount($this->raw());`

Which is exactly the same thing. I just added the limit before.

driesvints: @joaopalopes24 I still think we should revert it for now so this PR can focus on the thing it's trying to explain.

taylorotwell:
Whole thing is confusing. Would rather just remove macro docs.

PR #8710, opened 2023-04-08
[10.x] Add correct `assertJson`

milwad-dev:


taylorotwell:
I think current docs are good enough.

PR #8711, opened 2023-04-09
Fix closure request guard register method

Nothehi:
I suggest moving the `viaRequest` function inside the `register` method instead of the `boot` method. This is because the `boot` method is called after the configuration is loaded, and if the `viaRequest` function is inside the `boot` method, it will result in an error stating that this driver is not defined in the configuration.


taylorotwell:
Need more information because IMO it should be in the `boot` method.

PR #8716, opened 2023-04-11
[10.x] Add `logs` into helpers

milwad-dev:
I added an example in under of `logs` section, If you don't need tell me to remove them!


taylorotwell:
Think I'll leave this one out.

PR #8719, opened 2023-04-13
Add section on job trimming

gbradley:
See https://github.com/laravel/horizon/issues/1263. 

This PR adds a section on Horizon's `trim` config options, drawing particular attention to the fact that `trim.pending` may remove jobs from the queue before they are handled, becoming more likely when applications utilise delayed dispatching or have long wait times.

### DIFF
--- a/horizon.md
+++ b/horizon.md
@@ -177,7 +177,7 @@ To prevent Horizon's job lists from continually growing, these lists will be "tr
 
 > **Note**
 > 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hey there - this is already documented in the configuration file with an explanation. 👍

gbradley:
@taylorotwell The trim notes are in the config, it just seems odd that devs need to dig into the config and notice that pending jobs will disappear after an hour. I get that its intended behaviour, but the feature makes Horizon inherently less reliable than the basic DB driver, so thought that warranted a call-out. Anyway, cheers for considering.

PR #8720, opened 2023-04-14
[10.x] Add `noActionOnDelete` into migrations

milwad-dev:


taylorotwell:
I think it's fine to not have this. It's the default behavior.

PR #8722, opened 2023-04-14
Explain how to translate built in features

jasonvarga:
I had to recently explain to someone that uses our package that translations for built in features like validation need to go in their app, and shouldn't be part of our package.

I was surprised to see no mention of community maintained translations. It's very easy to not know these sort of packages exist, especially for newbies, and the lack of a mention kind of implies that users are expected to manually translate them.

Laravel Lang seems to be the most popular so I've mentioned that.


driesvints:

PR #8724, opened 2023-04-14
[10.x] Add `assertSimilarJson` in http-tests

Mohammadreza-73:


taylorotwell:
This doesn't actually describe what "similar" means, which I think we should do if we document this method.

PR #8725, opened 2023-04-15
[10.x] improve description of random method in str helpers

ceyhun-celik:
I am learning Laravel helpers. Almost every method contains comment of return. But Str::random not contain comment of return and I need to know default length of random. So I did in my PR. Best regards.


taylorotwell:
IMO you should just always pass a value.

PR #8726, opened 2023-04-15
[10.x] Add binding `alias` in container

milwad-dev:
In Container one of the most important method is `alias` method, and there is not section about alias in service container!!


taylorotwell:
This isn't really designed for end users.

PR #8733, opened 2023-04-21
Str::isMatch example added

localusercamp:
`Str::isMatch` example added


taylorotwell:
We would probably want to do this while documenting the other match methods as well.

PR #8734, opened 2023-04-21
Naming Resource Route Parameters - edit doc

mannv:
```php
Route::resource('users', AdminUserController::class)->parameters([
    'users' => 'admin_user'
]);
```
when run command: `php artisan route:list --path=users`

image


taylorotwell:
I think it's just giving a concise example.

PR #8741, opened 2023-04-25
Updated docs to direct people to place RateLimiter in boot method

jrgilman:
The RouteServiceProvider::configureRateLimiting method no longer exists in the 10.1.* app skeleton, but the documentation still refers to it (it used to exist in 10.0.*). This can cause some confusion.

Old App Skeleton: https://github.com/laravel/laravel/blob/v10.0.7/app/Providers/RouteServiceProvider.php

New App Skeleton: https://github.com/laravel/laravel/blob/v10.1.0/app/Providers/RouteServiceProvider.php

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -665,7 +665,7 @@ Using the `Route::fallback` method, you may define a route that will be executed
 <a name="defining-rate-limiters"></a>
 ### Defining Rate Limiters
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
There were other changes to document as well. Just pushed my changes.

PR #8746, opened 2023-04-26
add warning regarding using refresh tokens

mvynhb:
See here for more details: https://stackoverflow.com/a/12885823


bernardwiesner:
I know we already have a big warning on passport docs about only using OAuth it if you are familiar with it, but many still fall for the refresh_token over-engineering.

PR #8748, opened 2023-04-26
[10.x] add warning to storage link and change nginx recommended settings

bernardwiesner:
Security concern that should be addressed in my opinion.


taylorotwell:
No plans to merge right now.

PR #8756, opened 2023-04-30
[10.x] Add more example for `withoutTimestamps` method in eloquent

milwad-dev:
I think in withoutTimestamps section can add more example for more understand!


taylorotwell:
I think it's fine.

PR #8761, opened 2023-05-02
[10.x] Add disable-middleware section in http-tests

milwad-dev:
It's a missing method in http-tests!


taylorotwell:
I don't really document it because I don't think it's a good idea. It's not a real world test.

PR #8771, opened 2023-05-03
[10.x] Add `assertDatabaseEmpty` in database-testing

milwad-dev:


taylorotwell:
Use count

driesvints:
@milwad-dev sometimes, some methods remain in the framework for BC reasons but we remove them from the docs because we don't want to actively advertise them anymore.

PR #8778, opened 2023-05-06
[10.x] Add `value` method in eloquent

milwad-dev:
Is a very helpful method in Eloquent 👍


taylorotwell:
Will try to work it in.

PR #8782, opened 2023-05-08
details on how to use systemd

devbro1:
Adding details on how to use systemd as an alternative to supervisor


taylorotwell:
Feels like something more suited for a blog post or something?

PR #8785, opened 2023-05-12
Fix makeSearchableUsing using $query instead of $models

francoism90:
I don't fully understand why this has been changed to a Collection. Also the `$query` variable is not included.

If this is correct, how does it work?

Thanks.

### DIFF
--- a/scout.md
+++ b/scout.md
@@ -472,7 +472,7 @@ Sometimes you may need to prepare the collection of models before they are made 
     /**
      * Modify the collection of models being made searchable.
      */
-    public function makeSearchableUsing(Collection $models): Builder
+    public function makeSearchableUsing(Collection $models): Collection
     {
         return $models->load('author');
     }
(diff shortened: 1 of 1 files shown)


francoism90:
Hmm, it indeed uses something like this:
```php
/**
     * Modify the collection of models being made searchable.
     *
     * @param  \Illuminate\Support\Collection  $models
     * @return \Illuminate\Support\Collection
     */
    public function makeSearchableUsing(BaseCollection $models)
    {
        return $models;
    }
```

So shouldn't the example be like this instead:
```
public function makeSearchableUsing(BaseCollection $models)
    {
        return $models->load(...); // or $models->loadMissing(...)
    }
```

francoism90:
Hmm, it seems `makeSearchableUsing` and `makeAllSearchableUsing` are two separated things I mixed up. Still I think `$query` should be `$models` instead. :)

PR #8786, opened 2023-05-12
[10.x] Add `sear` method into Cache

milwad-dev:
Is a missing method that helpful that other developers know about it.
In my opinion, @taylorotwell you can write better description about it.


decadence:
`sear` is alias for `rememberForever` which is already in docs

PR #8788, opened 2023-05-12
[10.x] Clarifies issues when passing the `name` attribute to components

nunomaduro:
The purpose of this pull request is to address concerns regarding passing the `name` attribute on components slots, which was reported here: https://github.com/laravel/framework/issues/47057. @taylorotwell If you believe that this issue should be addressed in Laravel 10 or 11, please inform me, and I will follow up.


taylorotwell:
I'm just fixing it.

PR #8789, opened 2023-05-13
[10.x] Add `Retrieving Multiple Models` section in eloquent

milwad-dev:
Is a feature that is not exists in docs.
We can change into like:

```php
Flight::find([1, 2, 3]);
```


driesvints:
Feel free to send this in once you're ready

PR #8790, opened 2023-05-14
[10.x] Add `assertThrows` into http-tests

milwad-dev:


taylorotwell:
I have no idea what the `$response` is for in your first example.

PR #8793, opened 2023-05-18
Change example nginx document root path

briankiewel:
On systems where SELinux is enabled, files placed into the previous example nginx root path would be blocked from being read by nginx. This update provides an example path that would work on these systems because `/srv/www/*` is in the default SELinux fcontext ruleset.

### DIFF
--- a/deployment.md
+++ b/deployment.md
@@ -57,7 +57,7 @@ server {
     listen 80;
     listen [::]:80;
     server_name example.com;
-    root /srv/example.com/public;
+    root /srv/www/example.com/public;
 
     add_header X-Frame-Options "SAMEORIGIN";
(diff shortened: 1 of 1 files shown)


taylorotwell:
This code block isn't necessarily meant to be applicable to all OSs everywhere. Just a starting point.

briankiewel:
Yeah, lol, I get that it's an example. Just figured if it could be a starting point that'd work more universally it might be helpful to some, especially newer devs who may not have as much or any experience with OS management. You see a lot of "just disable SELinux" articles out on the internet when in reality some small changes (like putting the code in a more standard location) would fix the issues they may run in to. I'll get off my soapbox now, but appreciate the consideration!

PR #8794, opened 2023-05-18
fix roadrunner installation instructions

NiclasvanEyk:
It seems like the `spiral/roadrunner` package does not ship the `rr` binary anymore, so the current installation instructions won't work. According to [the roadrunner documentation](https://roadrunner.dev/docs/intro-install/2023.x/en#composer) you need to require the additional `spiral/roadrunner-cli` package to download binaries.


NiclasvanEyk:
Just saw that this is fixed in the master version of the docs, so this will likely resolve itself with the next release of the docs.

PR #8798, opened 2023-05-19
[10.x] Add `unless` method in eloquent-resources

milwad-dev:


taylorotwell:
Honestly I think `when` is fine. Unless can become confusing imo.

PR #8801, opened 2023-05-20
[10.x] Add Rule::when() section in Validation

TWithers:
I noticed that there were no docs on the `Rule::when()` validator that was added in 8.55 in the current docs. I added a quick example of how to use it under Conditionally Adding Rules to help clarify how it can be used.

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -1847,6 +1847,25 @@ In the example above, the `email` field will only be validated if it is present 
 > **Note**  
 > If you are attempting to validate a field that should always be present but may be empty, check out [this note on optional fields](#a-note-on-optional-fields).
 
+<a name="conditionally-modifying-rules-using-when"></a>
+#### Adding or Modifying Rules Using When
+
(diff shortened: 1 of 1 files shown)


TWithers:
The use case for this is rule over straight PHP is when the rules don't have access to the request input or data that is being validated. 

Example: I am reusing a several rules in different validators in my app, so I have abstracted the rule array to a helper method that can be injected throughout my code. I can't pass request data into my helper method because in one place I am validating a Livewire component, in another I am validating a request, and in another I am just validating an array. Using `Rule::when`'s closure and the `Fluent $input` allows me to grab the data currently in the validator without needing to write different checks on the data that is passed into my helper method.

taylorotwell:
We sort of document a way to do the exact same thing as this using `sometimes` just below this. It feels sort of redundant.

TWithers:
It definitely does, and I can see how it would be confusing. I just know, in the project where I am using `Rule::when()` (which is quite similar to the scenario above), `when()` allows me to keep the rule logic with the rules. Using `Validator::sometimes()` would have me implementing that same logic in 3 different places, or writing a wrapper for the Validator for the specific scenario. Personal preference is to not do that, so finding `Rule::when()` in the source code was exactly what I needed for this situation.

Thanks again for taking the time to review!

PR #8818, opened 2023-05-26
[10.x] Add note to the "View creators" section

shaedrich:
The docs are very sparsly emphasizing the difference between view creators and composers so that has been a very common question in forums concerning that topic every since.


taylorotwell:
This shouldn't be a note.

PR #8822, opened 2023-05-28
Add preventsAccessingMissingAttributes to "Configuring Eloquent Strictness" section

decadence:
Hello.

I don't know why but documentation doesn't mention `preventsAccessingMissingAttributes` method which I find very handy. It helped me many times while developing. Use cases:

- you forgot to run `php artisan migrate`
- you forgot to add new column in `select`
- you forgot to add foreign key in `select` and eager loading doesn't work
- you dropped column from database but it's still in your code
- you have typo in attribute name somewhere in your code


decadence:
Too bad. But at least this feature exists.

PR #8824, opened 2023-05-29
[10.x] scheduling: withoutOverlapping: Make sure the cache driver configuration is not array or null

erlangparasu:
To work with `withoutOverlapping`


erlangparasu:
because `array` cache driver is not creating mutex file to detect overlapping

erlangparasu:
@taylorotwell please provide comments or suggestions

erlangparasu:
without this documentation or warning, it feel like bug in laravel

PR #8825, opened 2023-05-29
[10.x] Add the use of the sole and soleValue methods to the query builder.

kevinb1989:


taylorotwell:
I feel like this example should make clear what the real-world use case for such a method is.

PR #8826, opened 2023-05-29
[10.x] Add use of the explain method in the query builder.

kevinb1989:


taylorotwell:
Doesn't say anything about what information is returned or in what format.

PR #8830, opened 2023-05-31
Added clarification for using nullOnDelete() method with nullable() c…

benounnas:
This commit adds an important clarification to the Laravel documentation regarding the usage of the nullOnDelete() method in conjunction with the nullable() column modifier before the constrained() method.

The added information emphasizes the need to include the nullable() column modifier before calling constrained() when using nullOnDelete(). This ensures that the foreign key column associated with nullOnDelete() is allowed to be nullable in the database schema.

By providing this clarification, users of Laravel can now better understand the correct order of operations when using nullOnDelete() and avoid any potential confusion or issues.


taylorotwell:
The sentence right below this says the same thing.

PR #8837, opened 2023-06-03
[6.x] Redis changed from predis to phpredis

NickSdot:
Either the first or second sentence is wrong. Afaik, it must be `phpredis`.


taylorotwell:
It's correct.

PR #8838, opened 2023-06-03
Update Redis client

rafaelstz:
It doesn't work with phpredis anymore, so I updated it to predis, I believe it's a change required by the third-party package `predis/predis` needed to run Redis through PHP.

It might be applicable just when we're using the `predis/predis` package so that we could add a note about it instead of changing the references.


taylorotwell:
> It doesn't work with phpredis anymore, so I updated it to predis

This needs explanation.

PR #8839, opened 2023-06-04
[10.x] Added example of using Axios to inject X-CSRF-TOKEN header in XCSRF section

ludoguenet:


taylorotwell:
I've never had to do this?

PR #8842, opened 2023-06-07
Update pint.md

benfreke:
feat: adds information on the laravel preset

Specifically, the changes that ChatGPT advised are different between PSR-12 and the "Laravel" preset available through Pint.

I found this issue, https://github.com/laravel/pint/issues/117, and thought it would be a simple documentation add as per the comments.

### DIFF
--- a/pint.md
+++ b/pint.md
@@ -5,6 +5,7 @@
 - [Running Pint](#running-pint)
 - [Configuring Pint](#configuring-pint)
     - [Presets](#presets)
+        - [Laravel Preset](#laravel-preset)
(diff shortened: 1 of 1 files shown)


HugoHeneault:
Hi there, 

Why is this closed? We were wondering what are the diff and this could have is place on doc?

Thanks

JohnnyWalkerDigital:
It's kind of insane that they give us a preset by default, but refuse to explain what those rules are. However something like this would be better suited to the Laravel Pint project than Laravel's documentation. This is too granular and not something the reader would be looking for there.

PR #8853, opened 2023-06-10
[10.x] remove duplication from register()

mwaqasiu:


taylorotwell:
Please see previous PRs. This is NOT duplicated. They are different classes.

PR #8858, opened 2023-06-12
Update validation.md

michael-lurquin:
Correction of the namespace for the use of SQL queries when using the "Rule::unique" function (with "where").

Fix message : `Argument #1 ($query) must be of type Illuminate\Database\Eloquent\Builder, Illuminate\Database\Query\Builder given`

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -1768,7 +1768,6 @@ Sometimes, you may wish to ignore a given ID during unique validation. For examp
 
 To instruct the validator to ignore the user's ID, we'll use the `Rule` class to fluently define the rule. In this example, we'll also specify the validation rules as an array instead of using the `|` character to delimit the rules:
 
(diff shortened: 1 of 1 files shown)


michael-lurquin:
It may not be an error, but when I put "use Illuminate\Database\Eloquent\Builder", it gives me the error `Argument #1 ($query) must be of type Illuminate\Database\Eloquent\Builder, Illuminate\Database\Query\Builder given` whereas when I put "use Illuminate\Database\Query\Builder", it works.

The error message doesn't make sense, or is it just me?

I use it in the following code:

`Rule::unique('pages')->where(function(Builder $query) {
    return $query->where([
        tenant_id' => $this->tenant_id,
        slug' => $this->slug,
    ]);
})->ignore($this->id),`

PR #8862, opened 2023-06-14
Add Attachment Macro usage

hungthai1401:
This PR includes the Attachment macro usage


taylorotwell:
The example should likely be realistic - always creating attachments from the same path doesn't make sense in my opinion.

PR #8864, opened 2023-06-14
Fix parameter order in signed URL expiry example

jnym-:
Puts the route params and expiry date parameters in the correct order in the signed URL with expiry date example.


taylorotwell:
CleanShot 2023-06-14 at 22 27 04@2x

PR #8869, opened 2023-06-20
Add information about `soketi` service available

olsza:
Hello all,
I noticed that in the SAIL documentation there is missing information that `soketi` can be given to `with`, which can be seen when verifying in the file https://github.com/laravel/sail-server/blob/master/routes/web.php#L23
So I added this information

A view of what the /routes/web.php file looks like:
image

A preview of how it will look in the documentation:
Zrzut ekranu z 2023-06-21 09-50-00


taylorotwell:
We are working on this atm.

PR #8878, opened 2023-06-23
Create Your Own Helpers.md

rmunate:
Standard Creation and Use of Helpers within (Laravel PHP Framework)

### DIFF
--- a/helpers.md
+++ b/helpers.md
@@ -8,6 +8,7 @@
     - [Lottery](#lottery)
     - [Pipeline](#pipeline)
     - [Sleep](#sleep)
+- [Create Your Own Helpers](#create-helpers)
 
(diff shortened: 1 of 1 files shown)


rmunate:
Standardize Laravel Helpers #47471
https://github.com/laravel/framework/discussions/47471

I have worked for many years with Laravel, and I have seen that in almost all projects the Helpers are created in different ways in different paths and folders, I have seen people who start using the Framework that when they see the syntax of a Helper similar to that of a PHP own function try to use the function on another system. After seeing this I wanted to standardize the process of creating Helpers, I would like to provide the feature and that a standard for creating helpers for developing systems, easy to use and implement, can be included in the official documentation.

PR #8880, opened 2023-06-25
[10x] Add example to create queued event listener

dammy001:


taylorotwell:
You can just add the interface after.

PR #8893, opened 2023-07-02
[10.x] Add missing `casts`

milwad-dev:


taylorotwell:
It is the same as array.

PR #8898, opened 2023-07-08
Add warning about docker context sail

basvdhoeven:
Is related to this https://github.com/laravel/sail/issues/81#issuecomment-1626401679

Installing Sail when using the `desktop-linux` docker context (which is the default for Docker Desktop on Linux) results in permission issues on for example the Storage folder.

### DIFF
--- a/sail.md
+++ b/sail.md
@@ -62,6 +62,9 @@ Finally, you may start Sail. To continue learning how to use Sail, please contin
 ./vendor/bin/sail up
 ```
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
We would to tell you how to actually do that.

basvdhoeven:
Thanks for the feedback @taylorotwell, I've created a new PR: https://github.com/laravel/docs/pull/8901.

PR #8913, opened 2023-07-12
LazyLoadingViolationException note

cosmastech:
https://github.com/laravel/docs/pull/8912

Per the closing of the last, I have added this as regular text.

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -1850,7 +1850,7 @@ public function boot(): void
 }
 ```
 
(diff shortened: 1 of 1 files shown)


cosmastech:
> Actually going to hold off on this.

@taylorotwell is there an expectation that this may change in 11.x?

When we came to realize that this is how it functioned, we were a bit dismayed. Our team discussed setting `$preventsLazyLoading = true` on a BaseModel class that our models extend from so that this is much stricter.

Would love to see this either more configurable (something like `Model::preventLazyLoading('all')`) or to have the `count()` requirement removed from `Builder@hydrate()` as it is unintuitive.

I know that last option is probably a considerable breaking change.

PR #8926, opened 2023-07-27
[10.x] Add Soketi as available service

leandrowferreira:
This PR only adds soketi as available Sail service during installation.


taylorotwell:
We are working on some things in this regard.

PR #8933, opened 2023-08-01
Fixed incorrect code in one of the examples of multiselect()

benjamincrozat:


devajmeireles:
Hey! @benjamincrozat , could you fix other things too, please? I was going to open a PR, but we can use yours! :D

1: We have `select` example inside `multiselect` examples:

CleanShot 2023-08-01 at 2 15 38

2: The `search` should returns the array after `pluck`. Otherwise, an exceptions is thrown:

CleanShot 2023-08-01 at 2 16 54

CleanShot 2023-08-01 at 2 17 21

taylorotwell:
@devajmeireles that is already fixed - thanks all

PR #8937, opened 2023-08-03
Update eloquent-relationships: $withCount by default

justinkekeocha:
Loading relationship counts by default when a model is retrieved


taylorotwell:
I honestly do not recommend this.

justinkekeocha:
care to give a reason why? Already guessing because it can be easily done with collections?

PR #8940, opened 2023-08-04
Update upgrade.md

mgralikowski:
Upgrade to Laravel 10 forced to move all attributes from `$dates` to `$casts`. It was a breaking change as we were struggling with a lot of errors related to forms.

It turned out that it was caused by [the change](https://github.com/mgralikowski/framework/commit/0d7ac95d98986171ea2419cf8c35768e380e1595) in the getDates() method.

Example of usage of this method: https://github.com/LaravelCollective/html/blob/64ddfdcaeeb8d332bd98bef442bef81e39c3910b/src/Eloquent/FormAccessible.php#L31

Generally, I think the name of this method now is misleading a bit, as it is limited only to timestamps. 
I am not the only one who noticed that -> https://github.com/laravel/framework/pull/42587#issuecomment-1462502416

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -191,6 +191,8 @@ protected $casts = [
 ];
 ```
 
+In the previous version, the Eloquent `getDates()` method was returning a list of attributes including timestamps and all attributes defined in `$dates` property. After moving dates to the `$casts` you should expect only two timestamps or an empty array.
+
 <a name="relation-getbasequery-method"></a>
 #### Relation `getBaseQuery` Method
 


taylorotwell:
It still returns a list of "date" fields.

mgralikowski:
@taylorotwell, no. It doesn't as there is no more `$dates` attribute available.

Before the update, on average it returned me 3-6 dates (Carbon objects), after the upgrade it is limited to only 2 timestamps. I have an application with ~25 models where `$dates` was used, so after an update, I had a lot of `Call to a member function format() on string` errors and I spent 2 hours to find out what is the reason. I just wanted to save people time.

Sadly, in my case, there is easy no fix for this, as getDates is used by a third-party library. But in case using this method directly in-app code, thanks to this information, I would know, that I have to expect different results, and probably prepare my own method to "get dates" in a model, as you did not accept my [PR](https://github.com/laravel/framework/pull/47960#issuecomment-1666179337).

PR #8944, opened 2023-08-07
Update grammatical error in team name validation message

phojie:
The original message had a minor grammatical error, using the plural form "characters" instead of the correct singular form "character."


taylorotwell:
But that may be actually what Laravel returns in this case?

PR #8950, opened 2023-08-10
[10.x] Update Grammer Fix an with a in artisan.md

PeterMMM:


PeterMMM:
Update Grammar Fix an with a in artisan.md

PR #8955, opened 2023-08-11
Update billing.md to reflect the code sample for syncing customer data

tdondich:
Update billing.md to reflect the code sample for syncing customer data is actually a closure directly on the Billable model. Update the previous text to reflect that.


taylorotwell:
I think the listener is the closure in this case.

PR #8957, opened 2023-08-13
[10.x] Describe optional parameters for scheduled invokable objects

vHeemstra:
Page: [Task Scheduling - Defining Schedules](https://laravel.com/docs/10.x/scheduling#defining-schedules)

This adds a small explanation with example about the optional second argument to the `call` method with invokable objects, including the need for it to be an associative array  with keys matching the parameter names of the invokable object.

### DIFF
--- a/scheduling.md
+++ b/scheduling.md
@@ -55,6 +55,20 @@ In addition to scheduling using closures, you may also schedule [invokable objec
 
     $schedule->call(new DeleteRecentUsers)->daily();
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
You can just do:

```php
$schedule->call(fn () => (new DeleteOldComments)(['before' => '6 months']))->monthly();
```

vHeemstra:
@taylorotwell Maybe my explanation was not clear enough (sorry).

This PR isn't about added the functionality. (It is already there.) This PR is about including an example of this functionality in the documentation.

PR #8959, opened 2023-08-13
10.x] Update Nginx config

TheAndrey:
I have improved the Nginx configuration example.

Added the necessary parameter checking existing of a PHP file. Without it, when requesting a URL http://example.com/index2.php we will get an empty page with the message "File not found" from the FastCGI backend. This event will also be recorded in the Nginx error log.

✅ With the addition of a file existence check, we will immediately receive a 404 error page from Nginx.

### DIFF
--- a/deployment.md
+++ b/deployment.md
@@ -76,6 +76,7 @@ server {
     error_page 404 /index.php;
 
     location ~ \.php$ {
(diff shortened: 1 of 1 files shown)


taylorotwell:
You can customize your own Nginx file as needed.

PR #8979, opened 2023-08-24
Document limiting client access to scopes in Passport

axlon:
https://github.com/laravel/passport/pull/1682

### DIFF
--- a/passport.md
+++ b/passport.md
@@ -41,6 +41,7 @@
     - [Assigning Scopes To Tokens](#assigning-scopes-to-tokens)
     - [Checking Scopes](#checking-scopes)
 - [Consuming Your API With JavaScript](#consuming-your-api-with-javascript)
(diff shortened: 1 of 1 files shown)


bram-pkg:
I just found the PR which merged this functionality after a long search for existing functionality. Any updates on this documentation PR?

hafezdivandari:
@axlon why not update the migrations on Passport and add `scopes` column to the `oauth_clients` table by default?

axlon:
> @axlon why not update the migrations on Passport and add `scopes` column to the `oauth_clients` table by default?

@hafezdivandari this is definitely possible and might even be preferable now with strict mode, I didn't opt to do this at the time because grant_types was already an opt-in feature within passport that worked the same way (it also wasn't a part of the schema by default)

driesvints:

PR #8994, opened 2023-08-29
[10.x] reset password page security tweak

craigpipeandpiper:
It is commonly accepted security best practice on a modern system that a reset password page should be non-committal about if a user existed or not. This is to prevent simple attacks against the reset password page to determine if a user is registered on a system or not by email address or username. Such attacks could then be combined with other attacks to break into user accounts.

As such i have changed the way the reset page returns its status to always return success. This would need to be combined with further tweaks to change the wording, i am not capable of making this change as it would involve changes to I18N content.

Please view this PR as a start of a longer discussion as to how to handle this simple tweak. I don't expect this is fine as-is, but wanted to broach the issue to test the water and see if this is something which should be adjusted to improve the security of projects built with Laravel.

### DIFF
--- a/passwords.md
+++ b/passwords.md
@@ -84,7 +84,7 @@ Before moving on, let's examine this route in more detail. First, the request's 
 
 The `sendResetLink` method returns a "status" slug. This status may be translated using Laravel's [localization](/docs/{{version}}/localization) helpers in order to display a user-friendly message to the user regarding the status of their request. The translation of the password reset status is determined by your application's `lang/{lang}/passwords.php` language file. An entry for each possible value of the status slug is located within the `passwords` language file.
 
-It is important to note that this endpoint will always return a success message even if the password entered is invalid. This is by design, so that an attacker cannot brute force the endpoint to determine if an email address is valid.
+It is important to note that this endpoint will always return a success message even if the email address entered is invalid. This is by design, so that an attacker cannot brute force the endpoint to determine if an email address is valid.
 
 > **Note**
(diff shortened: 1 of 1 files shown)


craigpipeandpiper:
Was recommended on the Discord to ping @valorin who may be able to weigh in and indicate if this is required or not.

taylorotwell:
I just leave this up to the user. In the default Laravel starter kits you can just try to register and see if an email is taken because it is unique.

valorin:
This is definitely the start of a **much bigger** discussion, but one that would require some big changes to Laravel's starter kits, etc, and I'm not sure if the trade-off is worth it?

Although it is a commonly accepted practice to use generic language for password resets to prevent enumeration of email addresses, this only works if the application doesn't provide alternate methods of identifying email addresses. Laravel's defaults use the email address as a unique username, which makes it trivial to enumerate email addresses from the user registration form too - the validator will kindly tell you if the email address has been taken or not. I actually prefer this method to using the reset form because it won't trigger any emails to existing users and validation failures are less likely to be logged than password reset failures.

The starter kits would need to be changed to require unique usernames instead of emails and/or use an email bounce as part of the registration flow to confirm access to the email without exposing existence of an account. Most apps don't bother with this because enumeration isn't typically a risk in itself - i.e. knowing someone has an account doesn't tell you much, and you can still conduct credential stuffing attacks regardless. The exception is if your app is a sensitive or highly personal service, where knowing someone has an account can share private information, but in these cases you'd want to build your own custom auth flow to protect your users anyway.

In terms of building apps for most purposes, I don't really see the need to add in that complexity by default. We'd be better served by making MFA easier to implement (and looking at things like Passkeys), as they'll make brute-force and credential stuffing attacks much harder to perform.

PR #8997, opened 2023-08-30
Fix Slack notification routing

WckD123:
Need to edit the example for sure.

The new version (^3.0) of the `slack-notification-channel` package, fails if we provide the route as a slack hook `(https://hooks.slack.com/services/...)` instead of a channel `#slack-channel`.

Apart from the example fix, I have also added a note for people who would be upgrading. This is exactly the issue I faced. The upgrade seems to be backwards compatible except when you start using block-notation. Then, you also need to update the route type from a URL/URI to `#sample-channel`

### DIFF
--- a/notifications.md
+++ b/notifications.md
@@ -301,7 +301,7 @@ Sometimes you may need to send a notification to someone who is not stored as a 
 
     Notification::route('mail', 'taylor@example.com')
                 ->route('vonage', '5555555555')
-                ->route('slack', 'https://hooks.slack.com/services/...')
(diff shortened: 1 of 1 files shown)


WckD123:
Just for context: https://github.com/laravel/slack-notification-channel/blob/9408050fb7fdb3ba45acb7a27a92c5cce98637de/src/SlackNotificationRouterChannel.php#L50

The first two conditions work well with the older version (where we had to provide slack hook urls). The third one is the new code. It has different functions and different objects (ones that don't support block notations)

PR #9011, opened 2023-09-06
Better explanation of the need of DBAL library for column type migrations

glamorous:
I noticed while trying to change type of timestamp to datetime in migrations that it didn't work in MySQL. According to the current documentation, it's only needed for SQLLite while the "warning" is saying what you can do if you use the package, but that's under the SQLite header, which people will "ignore" because they use another db engine. This change will hopefully help other developers to better understand if they need the package or not.

If I need to change something or there's a better explanation, I'm open for it 😉

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -1034,12 +1034,15 @@ When modifying a column, you must explicitly include all of the modifiers you wa
         $table->integer('votes')->unsigned()->default(1)->comment('my comment')->change();
     });
 
-<a name="modifying-columns-on-sqlite"></a>
-#### Modifying Columns On SQLite
+<a name="the-doctrine-dbal-library"></a>
+#### The Doctrine DBAL library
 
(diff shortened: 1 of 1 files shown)


glamorous:
Hi,

This migration isn't working in Latest laravel version with MySQL 5.7 when you don't have the DBAL package installed (which is optional)

```
public function up(): void
    {
        Schema::table('charging_sessions', function (Blueprint $table) {
            $table->dateTime('started_at')->default(null)->change();
            $table->dateTime('stopped_at')->nullable()->change();
        });
    }

    /**
     * Reverse the migrations.
     */
    public function down(): void
    {
        Schema::table('charging_sessions', function (Blueprint $table) {
            $table->timestamp('started_at');
            $table->timestamp('stopped_at')->nullable();
        });
    }
```

The current documentation only suggest to install the package for usage with SQLite.

PR #9015, opened 2023-09-08
[10.X] Document update/save/delete or fail methods

peterfox:
Based on the following PRs that don't appear to be documented:

https://github.com/laravel/framework/pull/38784
https://github.com/laravel/framework/pull/11202
https://github.com/laravel/framework/pull/38592


peterfox:
@taylorotwell totally understand. I'm curious should the methods not be deprecated in a future release if they're not documented? Seems odd to keep them in but also no encourage their use.

alexgaal:
> @taylorotwell totally understand. I'm curious should the methods not be deprecated in a future release if they're not documented? Seems odd to keep them in but also no encourage their use.

Laravel has a lot of functionality which is not documented (yet). Not all of them are deprecated, it's mostly just another way of solving a problem.

PR #9028, opened 2023-09-14
Improve deduplicating reporting setup example

timacdonald:
So much easier to just set the property in the exception handler, which is where you are probably looking for this feature. Keeps everything colocated.

## Before

Screenshot 2023-09-14 at 1 11 32 pm

## After

Screenshot 2023-09-14 at 1 11 51 pm


timacdonald:
Agree. Updated here and PR'd here: https://github.com/laravel/framework/pull/48413

Going to draft this until we tag that change.

PR #9030, opened 2023-09-14
[10.x] Prompts: Adds `spin`function documentation

manelgavalda:
This PR adds a new section for the `spin` function.

### DIFF
--- a/prompts.md
+++ b/prompts.md
@@ -10,6 +10,7 @@
     - [Multi-select](#multiselect)
     - [Suggest](#suggest)
     - [Search](#search)
+    - [Spin](#spin)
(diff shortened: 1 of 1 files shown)


taylorotwell:
Will let @jessarcher review

manelgavalda:
@jessarcher 

Just as a side note: I went with this example because I think it is a common case, but in case you want to go with a `sleep()` to make it simpler I will update it.

jessarcher (approved):
I've added a note about the requirement on the `pcntl` extension to get the animated version of the spinner.

I've also removed the `Http` import from the code snippet to be consistent with the other prompts, which don't show the imports for any of the classes used purely for example purposes.

PR #9037, opened 2023-09-18
update Collection::MergeWithKeys documentation

tobya:
A small additional note to clarify that a Key Value pair or an empty array should be returned.  

Specify that returning null will cause an error to be raised.


taylorotwell:

PR #9039, opened 2023-09-19
Add entry for `Str::of` to list of available methods for strings

peterhauke:
There does not appear to be a direct description of the `Str::of` function. It appears under [`str()`](https://github.com/laravel/docs/blob/a2cf776692b78089c6e06464fd2bcfad31604f85/strings.md?plain=1#L1188).

I am not sure whether a separate description for `Str::of` is necessary or not as the description for `str()` is clear. However, a link in the available methods sections might help people find this description faster.


taylorotwell:
I think it would be a bit weird to link somewhere that doesn't show `Str::of`. We would probably want a dedicated section for it. 👍

PR #9046, opened 2023-09-23
Adjust documentation for `decimal` migration method

LuisLeguizamon:
Add information about the `$unsigned` parameter for the `decimal` migration method

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -518,7 +518,7 @@ The `date` method creates a `DATE` equivalent column:
 <a name="column-method-decimal"></a>
 #### `decimal()` {.collection-method}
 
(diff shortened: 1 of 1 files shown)


hafezdivandari:
According to MySQL [docs](https://dev.mysql.com/doc/refman/8.0/en/numeric-type-syntax.html):

> As of MySQL 8.0.17, the UNSIGNED attribute is deprecated for columns of type [FLOAT](https://dev.mysql.com/doc/refman/8.0/en/floating-point-types.html), [DOUBLE](https://dev.mysql.com/doc/refman/8.0/en/floating-point-types.html), and [DECIMAL](https://dev.mysql.com/doc/refman/8.0/en/fixed-point-types.html) (and any synonyms); you should expect support for it to be removed in a future version of MySQL. Consider using a simple CHECK constraint instead for such columns.

taylorotwell:
See above.

PR #9052, opened 2023-09-25
Warning about order of background tasks

shaneniebergall:
I learned the hard way that tasks declared toward the top of the Kernel still execute before those at the bottom, so if there are any 'runInBackground' tasks that you expect to run asynchronously, that won't be the case unless they're at the top of the file.

Updating documentation so that others are aware.


taylorotwell:
I'm not sure we need to add a note for this.

PR #9057, opened 2023-09-27
Add section on custom model events in Eloquent

CodeWithKyrian:
This commit introduces a new section in the Eloquent documentation about custom model events. It covers both how to define these events and also how to listen to them in an observer class.

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -1575,6 +1575,49 @@ When models are being created within a database transaction, you may want to ins
         }
     }
 
+<a name="custom-model-events"></a>
+### Custom Model Events
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
No plans to document this. That's now really how I would do custom model events.

CodeWithKyrian:
Oh. That's fine. If you don't mind me asking, how then would you do custom model events? It's part of the framework but nowhere on the documentation where it's mentioned, does that mean it's unimportant?? The absence in the docs means few know about it and those who do might not be using it the best way it was intended. So what's that best way please?

PR #9067, opened 2023-10-03
Update sail.md

joseph-d:
Added the command to run the Sail installer on machines that don't have php installed locally


taylorotwell:
I don't really know why running `sail:install` would be necessary. That command is only run once at the beginning of a project.

PR #9072, opened 2023-10-06
Update broadcasting.md

Naghal:
Add a note for Laravel sanctum users to make the authorization route work with sanctum

### DIFF
--- a/broadcasting.md
+++ b/broadcasting.md
@@ -542,6 +542,10 @@ The `Broadcast::routes` method will automatically place its routes within the `w
 
     Broadcast::routes($attributes);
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Not enough information.

Naghal:
To provide clarification on why I made this PR:
This PR aims to hint Laravel Sanctum users that they need to move the `Broadcast::routes()` to their `api` route file in order to allow Laravel echo to be able to correctly authorize private channels. `auth:sanctum` middleware is also required.
This is because as the documentation states, `Broadcast::route()` will automatically place its routes within the `web` middleware group.

PR #9078, opened 2023-10-12
add note for receivers and mailables

torbentschechne:
The automatic retrieval of the recipient does not work if a `Mailable` is being used. It works for a MailMessage. This causes confusion and should be mentioned in the docs.


taylorotwell:
Not sure this is clear enough to know what you mean.

PR #9083, opened 2023-10-16
[10.x] Update events.md

siarheipashkevich:
Removed useless argument and import


taylorotwell:
I think it's to show you it's available if you need it.

siarheipashkevich:
@taylorotwell got it, but example above already has this ability.

PR #9087, opened 2023-10-18
[10.x] Clarify options to overwrite all keys instead of only the 'hosts' in the read/write configuration

PrinsFrank:


taylorotwell:
I think the existing wording is pretty sufficient?

PR #9090, opened 2023-10-20
[Pennant] update section regarding class based feature naming

Treggats:
I was working with a class based feature and was looking if I could change the name that was used.
Because I want to use the name in javascript, and the FQN was a bit hard to work with.

I came across this issue https://github.com/laravel/pennant/issues/67 and saw it was undocumented. Hence this pull request to add it to the docs.

## Changes
- update the class based feature section regarding using a property for naming the feature
- update the Blade directive section regarding using a class based feature FQN


taylorotwell:
I'll let @timacdonald take a stab at documenting this.

Treggats:
> I'll let @timacdonald take a stab at documenting this.

Sure no problem.

PR #9092, opened 2023-10-22
Remove reference to "null" log driver that doesn't exist

amadeann:
There's no "null" driver for logging. There is only a "null" channel.

### DIFF
--- a/logging.md
+++ b/logging.md
@@ -54,7 +54,6 @@ Name | Description
 `daily` | A `RotatingFileHandler` based Monolog driver which rotates daily
 `errorlog` | An `ErrorLogHandler` based Monolog driver
 `monolog` | A Monolog factory driver that may use any supported Monolog handler
-`null` | A driver that discards all log messages
 `papertrail` | A `SyslogUdpHandler` based Monolog driver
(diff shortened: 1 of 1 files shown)


browner12:
Before we get rid of this, can we add a section to discuss how to send logs to the "null" stack? It's still helpful information to know.

amadeann:
@browner12 
I added this in my AppServiceProvider to get the `null` log functionality working:

```
use Monolog\Handler\NullHandler;
use Monolog\Logger as MonologLogger;

...

if (($logManager = $this->app->make('log')) instanceof LogManager) {
    $logManager->extend('null_driver', function ($app, array $config) {
        return new MonologLogger('null_driver', [
            new NullHandler,
        ]);
    });
}
```

but I think this is too much of a workaround for the docs. And then, having the driver named "null" is problematic, since there's no way I know of to define string `null` in `.env` file.

browner12:
I wasn't suggesting a code change, just a docs blurb about how to use the "null" channel.

amadeann:
I am not sure what note would be appropriate. This section
https://laravel.com/docs/10.x/logging#writing-to-specific-channels
already covers the topic of logging to a specific channel.

PR #9093, opened 2023-10-23
Add a warning about MySQL behaviour in not nullable timestamps

frankmichel:
Added a warning about this behaviour:
https://ma.ttias.be/laravel-mysql-auto-adding-update-current_timestamp-timestamp-fields/

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -827,6 +827,9 @@ The `timestamp` method creates a `TIMESTAMP` equivalent column with an optional 
 
     $table->timestamp('added_at', $precision = 0);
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't totally want to document MySQL specific things.

frankmichel:
Not sure why you don't think that mentioning this isn't helpful for Laravel users. Since it will cause unexpected behaviour if not used correctly.

That being said, the migrations docs page is full of MySQL and/or SQLite/PostgreSQL specific warnings and clarifications.

PR #9100, opened 2023-10-26
Update documentation for `splice` method in collections.md

Robert-Smagge:
Adds a warning that `splice` modifies the collection (like the warnings that are already there for `forget` and `transform`)


taylorotwell:
Splice and slice don't serve exactly the same purpose so I'm not sure we can recommend it as a direct alternative. Also, our code examples show already that the existing collection is modified.

PR #9123, opened 2023-11-07
Updates Vite docs - `detectTls` -> `valetTls`

corbinmj16:
The `PluginConfig` now uses `valetTls` to Utilize the valet TLS certificates.


timacdonald:
We deprecated `valetTls` when we introduced Herd support.

https://github.com/laravel/vite-plugin/pull/233

`detectTls` is the correct key to use here.

PR #9129, opened 2023-11-10
Mention the optional numeric argument `@break()` / `@continue()` accept

u01jmg3:
After seeing this [StackOverflow comment](https://stackoverflow.com/questions/73014262/how-to-continue-outer-loop-from-inner-loop-inside-laravel-blade-template-larav#comment-128965201), I thought it was worth adding this info to the docs.

Wording loosely taken from the PHP docs for [`break`](https://www.php.net/manual/en/control-structures.break.php) / [`continue`](https://www.php.net/manual/en/control-structures.continue.php).


taylorotwell:
Will try to work this in.

PR #9131, opened 2023-11-13
Incorrect number update

semiherdogan:
Hi,
found a typo and updated it.


browner12:
While the total count **is** 10, the sequence will result in 5 of each option.

PR #9132, opened 2023-11-13
Meaning less JSONP

monsur22:
There is meaning less **JSONP** in responses.md file. I think it will be **JSON**. 
So if my thinking is correct, please merge this request.
Thank you.


browner12:
when you use the `->withCallback()` method, a JSONP response is returned, even though they both return the `JsonResponse` object.

PR #9133, opened 2023-11-13
docs: explain how to compute only one timestamp for model

ezef:
I spent some time searching for it and currently not explained in the docs. It will save some time for other people


taylorotwell:
I recommend people use both if they use them at all.

ezef:
Well, the thing is that for some cases both are not needed. In my case I have an Interaction model that I know it will be created but will never be updated, because it will be used to generate analytics. 

I think if the framework is supporting this behavior of deactivating one timestamp, It would be nice to have it on the docs

PR #9135, opened 2023-11-14
[10.x] Fix confusing namespace in Castables samples

Anton5360:


siarheipashkevich:
@Anton5360 maybe better to use `App\Casts\ ...` ?

https://laravel.com/docs/10.x/eloquent-mutators#value-object-casting

PR #9138, opened 2023-11-15
[10.x] Scout Database Driver toSearchableArray keys warning

edh649:
the toSearchableArray function allows changing of values to be searched using providers. The database driver does not respect these changes and instead only respects the keys returned as a filter for columns to search by (https://github.com/laravel/scout/blob/10.x/src/Engines/DatabaseEngine.php#L170).


taylorotwell:
I'm not sure I would understand what this means if I was reading it in the docs.

PR #9145, opened 2023-11-20
[10.x] Set languages on code blocks

jordihelp:
Many code blocks did not indicate of what language the code is based on.
- Original code blocks

    $foo->method()

- Modified code blocks
```php
$foo->method()
```


antimech:
Have you tested the build result?

taylorotwell:
The default is PHP.

PR #9146, opened 2023-11-20
[11.x] Use Gate Facade due to updated BaseController

Jubeki:
Mainly to make you aware of changes to be made due the skeleton slimming, where traits are removed from the Base Controller:
See https://github.com/laravel/laravel/pull/6188/files#diff-6eba513804f3952156d999eea3fb21199659f395d0092da71a10868b72bf59ce

- `$this->authorize(...)` will not be available, but can be replaced with `Gate::authorize(...)`
- `$this->authorizeResource(...)` will not be available without the trait, the method is currently not implemented in `Illuminate\Auth\Access\Gate` therefor marked as `TODO`

### DIFF
--- a/authorization.md
+++ b/authorization.md
@@ -579,6 +579,7 @@ Like the `can` method, this method accepts the name of the action you wish to au
     use App\Models\Post;
     use Illuminate\Http\RedirectResponse;
     use Illuminate\Http\Request;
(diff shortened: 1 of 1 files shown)


nunomaduro:
Thank you. This is already fixed.

PR #9149, opened 2023-11-22
Added section about using auth in global scopes

eskayamadeus:
It has not been previously documented that auth is not accessible in global scopes if used without first being checked. This commit adds the missing detail. 

**Many developers have run into this issue but it is still not documented:**
1. https://github.com/laravel/framework/issues/22316
2. https://www.reddit.com/r/laravel/comments/f4ni65/laravel_authhasuser_always_false_in_global_scope/
3. https://laravel.io/forum/06-30-2016-global-scope-auth
4. https://stackoverflow.com/questions/68256830/laravel-how-to-use-auth-user-inside-the-global-scope

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -1261,6 +1261,37 @@ Eloquent also allows you to define global scopes using closures, which is partic
         }
     }
 
+<a name="using-auth"></a>
+#### Using Auth in Global Scopes
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
I mean it just depends on where and when the query is executed?

PR #9156, opened 2023-11-29
Cursor pagination items warning

cosmastech:
I'm not entirely sure this belongs to the existing warning, but I spent the last few hours trying to debug an issue that was reported and it's not obvious why this is a problem.

Say you have a database with two tables

**conversations**
| id  | created_at |
| ------------- | ------------- |
| 25  | 2023-01-01 00:00:00  |

**messages**
| id  | conversation_id | message | created_at |
| ------------- | ------------- | ------------- | ------------- |
| 98 | 25 | Hello! | 2023-11-29 13:45:12  |

And you want to build a cursor like this 
```php
$query = Conversation::query()
    ->select('conversations.*')
    ->join('messages', 'conversation.id', 'messages.conversation_id')
    ->orderBy('messages.created_at')
    ->orderBy('conversations.id');

$query->cursorPaginate()->nextCursor()->encode(); // <-- base64 string
```

Will build a cursor with
```php
[
    '_pointsToNextItems' => true,
    'conversations.id' => 25,
    'messages.created_at' => '2023-01-01 00:00:00', // <-- this should be 2023-11-29 13:45:12
]
```

But the value is actually the conversations.created_at value, not the messages.created_at value.

### DIFF
--- a/pagination.md
+++ b/pagination.md
@@ -126,7 +126,7 @@ You may create a cursor based paginator instance via the `cursorPaginate` method
 Once you have retrieved a cursor paginator instance, you may [display the pagination results](#displaying-pagination-results) as you typically would when using the `paginate` and `simplePaginate` methods. For more information on the instance methods offered by the cursor paginator, please consult the [cursor paginator instance method documentation](#cursor-paginator-instance-methods).
 
 > **Warning**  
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hmm, I'm not sure I would totally understand this note. Can it be re-worded? 😅

cosmastech:
@taylorotwell took another stab at it. 😅

Kudos to you and all of the other docs writers who are capable of writing things succinctly and in a way which is digestible for newcomers.

PR #9173, opened 2023-12-04
Add section about streaming reads with the s3 driver

adi64:
This PR is the spiritual successor to https://github.com/laravel/docs/pull/8371.

The `s3` filesystem driver doesn't *really* stream reads by default, even when calling e.g. `readStream`. Instead, it copies the whole file to a temporary directory and then returns a stream to that temporary file. This is because streaming reads are disabled by default:

https://github.com/laravel/framework/blob/v10.34.2/src/Illuminate/Filesystem/FilesystemManager.php#L244

This was introduced in https://github.com/laravel/framework/pull/34001, but without an explicit reason. I also asked about this behavior here, but without luck: https://github.com/laravel/framework/discussions/49232.
I suspect that it was disabled by default because Flysystem changed the default behavior to using streams, and that could've broken backwards compatibility in Laravel 7.x, and it was never touched since then.

### DIFF
--- a/filesystem.md
+++ b/filesystem.md
@@ -193,6 +193,12 @@ Typically, after updating the disk's credentials to match the credentials of the
 
     'endpoint' => env('AWS_ENDPOINT', 'https://minio:9000'),
 
+<a name="streaming-reads"></a>
+#### Streaming Reads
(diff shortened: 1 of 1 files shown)


taylorotwell:
Will try to work this in.

PR #9175, opened 2023-12-04
[10.x] clarify need for backed Enum

browner12:
the validation rule currently requires a backed Enum if the value being validated is a `string` or `int`.

I think it would be nice if the rule also handled pure Enums when passed a `string`, but that's a separate issue. I'm going to attempt a PR for that in `laravel/framework`, but until then this clarification in the docs should help.


taylorotwell:
Sorry, not sure I follow. Passing an int or string value to what?

PR #9181, opened 2023-12-06
Add PostgeSQL for Pulse

Jubeki:
I saw that @jessarcher added PostgreSQL support for Pulse, therefor this limitation can be updated. (Maybe it can even be removed? Not sure how it looks with MariaDB which is MySQL compatible and for not first party supported databases like MongoDB)

### DIFF
--- a/pulse.md
+++ b/pulse.md
@@ -28,7 +28,7 @@ For in-depth debugging of individual events, check out [Laravel Telescope](/docs
 ## Installation
 
 > **Warning**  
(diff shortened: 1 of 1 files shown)


taylorotwell:
@jessarcher this ok?

jessarcher:
@taylorotwell Yeah, I think this makes sense. No support for MSSQL or SQLite currently so I think it makes sense to keep the warning for now.

PR #9184, opened 2023-12-08
Update broadcasting.md

ictgtvt:
Add guide for using custom guards.

### DIFF
--- a/broadcasting.md
+++ b/broadcasting.md
@@ -541,6 +541,10 @@ The `Broadcast::routes` method will automatically place its routes within the `w
 
     Broadcast::routes($attributes);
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think we would need an explanation for *why* you want to pass an auth guard here.

ictgtvt:
Hi @taylorotwell, I just found that when we use 2 guards (Example: `web` guard as default as `admin` guard for admin dashboard: Pusher private channel authentication needs have `auth:admin` in Broadcast::routes.

If we don't do that, this code can not retrieve the authenticated user info:
https://github.com/laravel/framework/blob/10.x/src/Illuminate/Broadcasting/Broadcasters/Broadcaster.php#L340

PR #9185, opened 2023-12-08
Document global Pint installation

caendesilva:
Since this is officially supported (https://github.com/laravel/pint/issues/70, https://github.com/laravel/pint/issues/63) I figure it's only helpful to document how to install Pint globally.


taylorotwell:
Only if the Composer global bin directory is in your PATH.

PR #9195, opened 2023-12-14
[10.x] Add note about disabling pulse

driesvints:
This hopefully will stop a lot of incoming issues 😅

### DIFF
--- a/pulse.md
+++ b/pulse.md
@@ -78,7 +78,7 @@ Pulse is not meant to be run during testing and as such should be disabled. You 
 <env name="PULSE_ENABLED" value="false"/>
 ```
 
-Additionally, if you're using sqlite during testing, you should prevent the Pulse migrations from running by adding this check in the `up` method:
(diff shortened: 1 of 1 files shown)


jbrooksuk (approved):

pulse.md
```diff
@@ -69,6 +69,28 @@ Many of Pulse's configuration options can be controlled using environment variab
...
+<a name="disabling-pulse-during-testing"></a>
+#### Disabling Pulse During Testing
+
+Pulse is not meant to be run during testing and as such should be disabled. You can easily achieve this by setting the `PULSE_ENABLED` environment variable to `false` in your `phpunit.xml` file:
+
+```xml
+<env name="PULSE_ENABLED" value="false"/>
+```
+
+Additionally, if you're using sqlite during testing, you should prevent the Pulse migrations from running by adding this check in the `up` method:
```
antimech, line 81:
```suggestion
Additionally, if you're using SQLite during testing, you should prevent the Pulse migrations from running by adding this check in the `up` method:
```

taylorotwell:
Fixing this in Pulse itself if possible.

PR #9198, opened 2023-12-14
Improve wording on the `whereJsonContains` method and SQLite support

danieleambrosino:
I find the wording "This feature [whereJsonContains] is not supported by SQLite database versions less than 3.39.0" misleading. This wording suggests that the `whereJsonContains` method can be used by merely using a compatible version of SQLite, which is not true; while SQLite does support JSON functions since version 3.39.0 (as already stated in the previous paragraph), the SQLite Grammar does not support the `whereJsonContains` clause regardless of the version of SQLite.

### DIFF
--- a/queries.md
+++ b/queries.md
@@ -491,7 +491,7 @@ Laravel also supports querying JSON column types on databases that provide suppo
                     ->where('preferences->dining->meal', 'salad')
                     ->get();
 
-You may use `whereJsonContains` to query JSON arrays. This feature is not supported by the SQLite grammar yet:
+You may use `whereJsonContains` to query JSON arrays. This method is not supported by SQLite:
 
     $users = DB::table('users')
                     ->whereJsonContains('options->languages', 'en')
--- a/queries.md
+++ b/queries.md
@@ -491,7 +491,7 @@ Laravel also supports querying JSON column types on databases that provide suppo
                     ->where('preferences->dining->meal', 'salad')
                     ->get();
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hmm, what happens if you do it? Wouldn't it fall back to the base grammar's `whereJsonContains`? What error are you actually receiving?

danieleambrosino:
If you try to build a query adding a `whereJsonContains` constraint while using the SQLite driver, a `RuntimeException` is thrown with the message "This database engine does not support JSON contains operations."
The exception is thrown [here](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Query/Grammars/Grammar.php#L613).
This is due to the fact that SQLiteGrammar does not provide any implementation of the protected method `compileJsonContains`, hence it falls back to the implementation provided by its parent `Illuminate\Database\Query\Grammars\Grammar`, which just throws the aforementioned exception.

danieleambrosino:
I also just found out [this issue](https://github.com/laravel/framework/issues/49217) opened a couple of weeks ago in the laravel/framework repo regarding the same topic, that's still open and flagged as "help wanted".
This kind of query is actually feasible in SQLite nowadays. I have a decent amount of experience dealing with unstructured data in SQLite. I'd be glad to contribute submitting a PR with a proposal of implementation.

taylorotwell:
@danieleambrosino got it - thanks. yeah, open to PRs if it's not too hard to fix!

PR #9208, opened 2023-12-20
Add manual trigger for model broadcasting

hussaini:


taylorotwell:

hussaini:
Hi @taylorotwell 

The code that I mentioned in the docs is actually part of `Illuminate\Database\Eloquent\BroadcastsEvents`, I don't think I need to release as package.

I think it might be useful to let people know that model broadcast event can be manually triggered instead of need to 'touch' the model so that it fire the event.

In my case, I have a model and it's relation, lets name it `A` and `B` respectively. I make some changes to `B` , but `A` doesn't fire any event since it doesn't receive any changes. Hence I need to manually trigger model broadcast event, which is in my case I just use `->broadcastUpdated()` to fire `updated` event.

PR #9217, opened 2023-12-21
added note that the db Artisan command does not work on windows

Mavv3006:
Currently when you run the `db` Artisan command on a windows computer the CLI throws a `Symfony\Component\Process\Exception\RuntimeException` stating that 'TTY mode is not supported on Windows platform.'

This PR will document this limitation in the docs.


taylorotwell:
Would need to explore this further before adding this.

PR #9220, opened 2023-12-22
Update deployment.md

bhanwarpsrathore:
Intl extension is required for Number helper


taylorotwell:
That's a very small part of the framework, not a core requirement.

bhanwarpsrathore:
I use Laravel Herd for development and it was working fine, then it broke on production as 'intl' extension was not there. It has to be somewhere I think :)

PR #9229, opened 2023-12-29
[10.x] Adjust documentation to use new Laravel Sail supervisord environment substitution support

bram-pkg:
This should be merged after https://github.com/laravel/sail/pull/645 is released


bram-pkg:
@taylorotwell here is the PR which documents the new environment variable.

What do you think about also assuming people have the `sail` alias? Would clean up this document a bit, I think.

PR #9231, opened 2023-12-31
[10.x] Remove the duplicate registration line for Telescope.

kevinb1989:


michaeldyrynda:
I don't think this is a duplicate registration because there are two providers; the vendor one and the one in your `app/Providers` folder. 

This change will only use the vendor one.

kevinb1989:
I believe we have two different TelescopeServiceProvider classes and got to register both. Close this now.

PR #9237, opened 2024-01-03
[9.x] Update upgrade guide (HandleCors)

frankykubo:
Everytime I am upgrading from 8.x, I need to google about the need of removing *fruitcake/laravel-cors* package from app's composer.json, because it is not able to update packages this way. Also, after successfull update, changing the _HandleCors_ middleware in Kernel.php is needed.

In this P.R., I have added a few lines that mentions this process and easy steps which will help users migrate from 8.x to 9.x without need of googling this issue.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -103,6 +103,16 @@ In addition, return types were added to methods implementing PHP's `SessionHandl
 
 </div>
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'll have to rewrite the wording to match the rest of the documentation.

PR #9253, opened 2024-01-10
Update logging.md

mortenscheel:
The context code example strongly implies that `{id}` in the message string will be replaced with the value from the context. But that is _not_ the case for log channels that use the default monolog driver.

I'm not sure if this is the right place to put it, but as far as I can tell, the `replace_placeholders` config setting isn't documented anywhere.


taylorotwell:
Then it should be fixed in the default laravel/laravel configuration IMO.

PR #9255, opened 2024-01-11
Update sanctum.md

TodoshiTakato:
Corrected and improved the "Issuing API Tokens" block where Laravel PHP code is given as an example

### DIFF
--- a/sanctum.md
+++ b/sanctum.md
@@ -130,7 +130,7 @@ Sanctum allows you to issue API tokens / personal access tokens that may be used
 To begin issuing tokens for users, your User model should use the `Laravel\Sanctum\HasApiTokens` trait:
 
     use Illuminate\Database\Eloquent\Model;
-    use Illuminate\Contracts\Auth\Authenticatable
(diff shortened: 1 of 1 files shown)


taylorotwell:
Why?

TodoshiTakato:
Because User model class should first extend default Laravel model
'Illuminate\Database\Eloquent\Model',
then implement
'Illuminate\Contracts\Auth\Authenticatable' interface.

I think Class can not extend Interface. At least PHPStorm IDE is saying that 'Class reference expected',
and advices to "Change 'extends Authenticatable' to 'implements Authenticatable' "

Jubeki:
In the laravel skeleton it also uses `extends Authenticatable` which is an alias:

```php
use Illuminate\Foundation\Auth\User as Authenticatable;
```

See https://github.com/laravel/laravel/blob/2254e8e301cfcbf8983c2f7cfa8b7c63549b73c7/app/Models/User.php#L7

TodoshiTakato:
Oh, okay, thank you very much, I was wrong.

PR #9260, opened 2024-01-15
[10.x] Adding Missing Content of `Configuring Eloquent Strictness`

devajmeireles:
I used `9.x` as a base to include the missing content in `10.x`


taylorotwell:
I don't really like this feature so I removed it.

PR #9264, opened 2024-01-18
Update billing.md - setting Billing anchor for a subscription checkout

Digitwellsolutions:
It is possible to set the billing cycle for a new subscription checkout. 
I have created a new section to describe this, but I did not update the warning in the previous section.
The language is probably not consistent with the existing. 
Please feel free to modify or re-write.


taylorotwell:
I'll try to work this into the flow of the documentation.

PR #9269, opened 2024-01-19
[10.x] updates for the new Settler box and Homestead version

browner12:
- Mailhog is gone. Mailpit is the default and only option now.
- Ubuntu is now version 22.04
- PostgreSQL is no version 16
- The "mysql" feature flag never actually did anything, so we drop it.
- Add PHP 8.3

@svpernova09 can you confirm the Mailpit config has not changed?

### DIFF
--- a/homestead.md
+++ b/homestead.md
@@ -74,7 +74,7 @@ Homestead runs on any Windows, macOS, or Linux system and includes Nginx, PHP, M
 - MySQL 8.0
 - lmm
 - Sqlite3
-- PostgreSQL 16
+- PostgreSQL 15
 - Composer
 - Docker
 - Node (With Yarn, Bower, Grunt, and Gulp)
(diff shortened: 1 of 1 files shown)


svpernova09:
> @svpernova09 can you confirm the Mailpit config has not changed?

Confirmed the configuration did not change, just the binary.

homestead.md
```diff
@@ -73,14 +74,14 @@ Homestead runs on any Windows, macOS, or Linux system and includes Nginx, PHP, M
 - MySQL 8.0
 - lmm
 - Sqlite3
-- PostgreSQL 15
+- PostgreSQL 16
```
svpernova09, line 77: PostgreSQL 15 https://github.com/laravel/settler/blob/master/scripts/amd64.sh#L750

browner12: fixed

PR #9274, opened 2024-01-22
[10.x] Add `windows_os` to helpers

milwad-dev:


taylorotwell:
This is more for internal usage.

PR #9277, opened 2024-01-23
[10.x] bump php requirement

franciscobrba:
composer require php 8.2 to fetch all dependencies.
ngnix boilerplate also require php 8.2.

simple as that


franciscobrba:
@taylorotwell hey, just closed? the change seems pretty reasonable

PR #9279, opened 2024-01-24
[10.x] Note on using the `Sanctum:ignoreMigrations` method in sanctum.md

nasirhumbatli:


taylorotwell:
I'm not sure this note is needed.

nasirhumbatli:
When we create a Laravel project, it would be more appropriate if Sanctum's default migrations are not automatically exported to the `database/migrations` folder. Then, if we don't want to use Sanctum's default migrations, we can use `Sanctum::ignoreMigrations` method and we don't need to remove Sanctum's default migrations from `database/migrations` folder.

Kamarbandi:
@taylorotwell  I completely agree with @nasirhumbatli . Having Sanctum's default migrations not automatically exported to the database/migrations folder when creating a new Laravel project would offer more flexibility. The ability to use Sanctum::ignoreMigrations method is a great solution, eliminating the need to manually remove default migrations from the database/migrations folder if they are not needed. This approach enhances customization and simplifies the setup process. 👍

PR #9280, opened 2024-01-24
Add mention of required Typesense configuration in Scout config

jasonbosco:
Based on feedback from here: https://github.com/laravel/scout/issues/796#issuecomment-1908591943

### DIFF
--- a/scout.md
+++ b/scout.md
@@ -177,51 +177,7 @@ public function toSearchableArray()
 }
 ```
 
-You should also define the Typesense collection schema in your application's `config/scout.php`, which describes the data types of each field that Typesense will use to validate incoming data:
-
-```php
-// In config/scout.php
-[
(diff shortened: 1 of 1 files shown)


taylorotwell:
@jasonbosco I personally would like to see a dedicated Laravel page in the Typesense documentation that you can maintain on your end with more detailed instructions on how to actually update the schema definition. Right now, imo, that is the worst part of the DX story in this integration. We could then link to that page from our docs and not have to maintain them on our side so closely and with such detail.

jasonbosco:
@taylorotwell Yeah, I'm down to do that. The next time someone asks for more details than is currently in the Laravel docs, I'll put together a dedicated page in the Typesense docs which addresses the question and provides more detailed instructions. I'll then open a PR in this repo, with a link to that dedicated page.

PR #9284, opened 2024-01-28
Remove laravel-websockets

epixian:
The `laravel-websockets` package does not work out of the box with Laravel 10, and seems to have been abandoned, judging from the [numerous issues](https://github.com/beyondcode/laravel-websockets/issues) raised.

Removed the reference entirely.


epixian:
@freekmurze @mpociot is the package still alive or no?

PR #9287, opened 2024-01-30
[10.x] Explain UTC behaviour when serialising dates

driesvints:
Currently, it's still not clear that dates will always be displayed as UTC when being serialised in Eloquent models. This extra sentence should make this crystal clear. See https://github.com/laravel/framework/issues/49899


taylorotwell:
I'm not sure I totally follow. 🤔

driesvints:
@taylorotwell the timezone you set in your app.php will be applied but the serialised date is ALWAYS in UTC.

PR #9293, opened 2024-01-31
Corrected order of some functions to alphabetically.

kasperhartwich:
Some functions were our of alphabetic order. Some was actually in different orders from contact to TOC.


taylorotwell:
Some of them are intentionally kept by their "related" counterparts... for example empty and is isNotEmpty.

PR #9295, opened 2024-02-01
Fix conditional clauses examples

glamorous:


taylorotwell:
What is being fixed? The value really is injected in as an arg. Did you check the implementation of `when`?

glamorous:
Hi Taylor,

I was following the implementation and it didn't work as expected, so in my opinion docs were wrong and the way I proposed is the normal way of "injecting" something in the closure from outside... . In my testcase I used an integer with value "2" and inside the closure the value was passed as "true".

To give you an example:

```
$customer_id = 2;

AccessRight::query()
	->whereNull('manual')
	->when($customer_id !== null, function (Builder $query, $customer_id) {

		$query->where('customer_id', $customer_id);
	})
	->dump()
```

I expected that the query would have 'customer_id = 2' but instead you get 'customer_id = true'. See:

```
"select * from [accessRight] where [manual] is null and [customer_id] = ?"
array:1 [
  0 => true
]
```

It is only working that the result of the expression is injected. So in the example above the `!== null` should be deleted to make it work like documentated. For me it was a waste of time and unnecessary debugging. I wanted to help people to have the other approach documentated as well, the one that is always working :-)

But now it's clear for me 😄 it's injecting the exact same thing that you provide in the `when`-expression, in my case the result "true".

PR #9307, opened 2024-02-05
[10.x] Assigning a validation message info in upgrade docs

peterfox:
# Why

Found during a recent upgrade. Appears to be very obscure but a breaking change that others might come across when upgrading from 9 to 10.


taylorotwell:
Dunno why it's in the routing section.

peterfox:
My apologies @taylorotwell , I confused myself when making the edit. Will reapply to correct section if that's okay?

PR #9308, opened 2024-02-06
Add documentation for Passport route prefix value `path`

selfsimilar:
Passport's route prefix (default `oauth`) is set in `config/passport.php` but is not documented anywhere, and is not in the publishable config file in the Passport project. I have opened a [PR to add `path` to the config file](https://github.com/laravel/passport/pull/1714), but the official Passport docs should also document this config value.


taylorotwell:
Not really sure this should be customized.

PR #9320, opened 2024-02-09
[10.x] Update Laravel installer to 5.5

nshiro:


taylorotwell:
Reverting this version number. We went like 12 years without this I don't think we need it.

PR #9323, opened 2024-02-12
Add maintenance mode warning in Queues Documentation

achyutkneupane:
My colleague got stuck with the queue not working problem because he was in maintenance mode.

After looking into the issue, we found out the cause and saw [Queues](https://laravel.com/docs/10.x/queues) documentation doesn't have the maintenance mode warning.

So, here it is.


Dipesh79:
Thank You @achyutkneupane

PR #9334, opened 2024-02-14
[10.x] Fix security vulnerability in Cashier Stripe quickstart example

zepfietje:
Currently the payment status is not checked.
This means the user can start a Stripe checkout session, get the session ID from the URL, and manually construct the checkout success route, marking the order as paid without paying.


taylorotwell:
Thanks!

PR #9342, opened 2024-02-15
fix: wording for increment and incrementEach

ahadcove:
When using increment/decrement you can update only one row or the full table.
When using incrementEach/decrementEach you can only update all rows even if only one row is selected.

This documentation adds that difference to the docs in case users run into this issue in the future like i did


taylorotwell:
I don't really follow this.

PR #9345, opened 2024-02-16
Update notifications.md

fredbradley:
I'm 99% sure that the documented namespace has been mistyped. The correct name space is `Illuminate\Notifications\Messages\SlackMessage`


fredbradley:
No, I'm probably wrong. I see there are two SlackMessage.php files - I need to better understand why my app is using one rather than the other.

PR #9359, opened 2024-02-19
[10.x] Add warning for `trashed` event in Eloquent

milwad-dev:
Many developers don't know that when they want to use the `trashed` event, they need to use the `softDeleted` event.

https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Eloquent/SoftDeletes.php#L171


taylorotwell:
I'm not sure this needs a big box in the docs.

PR #9365, opened 2024-02-20
[11.x] Add eager load limit to Eloquent

milwad-dev:
Just need to change the description!


taylorotwell:
Not sure this needs to be documented. It just works.

PR #9367, opened 2024-02-20
Update database query builder page

hamannjames:
I believe it would be valuable to add some clarity around what `update` can do. It can only update `fillable` properties, but this is not clear. I added that clarity.

### DIFF
--- a/queries.md
+++ b/queries.md
@@ -939,7 +939,7 @@ In the example above, Laravel will attempt to insert two records. If a record al
 <a name="update-statements"></a>
 ## Update Statements
 
-In addition to inserting records into the database, the query builder can also update existing records using the `update` method. The `update` method, like the `insert` method, accepts an array of column and value pairs indicating the columns to be updated. The `update` method returns the number of affected rows. You may constrain the `update` query using `where` clauses:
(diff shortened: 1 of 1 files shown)


decadence:
This is not true for Query Builder. When you do DB::update, model and fillable are not used.

hamannjames:
@decadence you are right! But when you call the update method on an eloquent class, it does require the properties to be fillable. I still find this confusing. I have looked over the docs and cannot find a good place this is explained.

I am going to close this, but still recommend it is made more clear in the docs. Is there a place you know of that better explains the `update` method being called by a model object that I am missing?

decadence:
You can really get confused here :) `DB::update` or `Model::update` update records massively without events and other Eloquent features. When you call specific model instance `update` method, it just calls `fill` and `save` on that instance and that requires field to be fillable.

I didn't find `update` method in Eloquent documentation, only [this](https://laravel.com/docs/10.x/eloquent#updates). Mass Updates section is only about first case.

First: https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Query/Builder.php#L3434
Second: https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Eloquent/Model.php#L984

PR #9370, opened 2024-02-21
[10.x] Update all references of Builder to the contract on the Eloquent: Relationships page

freerkminnema:
Upgrading from v10.44.0 to v10.45.0, our code broke because we were using code similar to this:

```php
use Illuminate\Database\Eloquent\Builder;

public function posts(?string $tag = null): HasMany
{
    return $this->hasMany(Post::class)
        ->when($tag, fn (Builder $query, string $tag) => $query->where('tag', $tag));
}
```

This caused:

> Unknown error TypeError: App\\Models\\{closure}(): Argument #⁠1 ($query) must be of type Illuminate\\Database\\Query\\Builder, Illuminate\\Database\\Eloquent\\Relations\\HasMany given

It appears this PR was the cause: https://github.com/laravel/framework/pull/50124

Looking through the PR description, this is actually exactly the edge case we are experiencing:

> The only edge case I can think of is someone receiving the `Builder` instance inside these methods callbacks, but that was deprecated in favor of the `BuilderContract`, so its a far stretch.

I have resolved the issue locally by typing the received variable to this `BuilderContract`, but I was unaware of the existence of the _BuilderContract_ and that this should be used in favor of the direct `Builder` import.

Since the documentation about relationships almost exclusively, save one instance, refers directly to the `Builder` namespace, I wanted to submit this PR to guide future developers to using the `BuilderContract` instead.

### DIFF
--- a/eloquent-relationships.md
+++ b/eloquent-relationships.md
@@ -1264,7 +1264,7 @@ where user_id = ? and active = 1 or votes >= 100
 
 In most situations, you should use [logical groups](/docs/{{version}}/queries#logical-grouping) to group the conditional checks between parentheses:
 
-    use Illuminate\Database\Eloquent\Builder;
+    use Illuminate\Contracts\Database\Eloquent\Builder;
 
     $user->posts()
             ->where(function (Builder $query) {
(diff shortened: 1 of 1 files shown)


siarheipashkevich:
@freerkminnema should I use contract builder for all my conditionable methods?

freerk-mpluskassa:
> @freerkminnema should I use contract builder for all my conditionable methods?

That's what I did to prevent any such change as https://github.com/laravel/framework/pull/50124 from impacting our codebase in the future.

freerkminnema:
> @freerkminnema should I use contract builder for all my conditionable methods?

That's what I did to prevent any such change as https://github.com/laravel/framework/pull/50124 from impacting our codebase in the future.

taylorotwell:
That PR was reverted.

PR #9371, opened 2024-02-21
[11.x] Documents `once` helper

nunomaduro:
This pull request documents `once` helper.


decadence:
It may be worth specifying that this function will cache different result for different objects

nunomaduro:
@decadence Not sure we want to elaborate on all details. Lets see what @taylorotwell thinks!

PR #9373, opened 2024-02-21
Document `login()` via `user()`

weshooper:


weshooper:
@taylorotwell thanks for looking. I've added a description above, is there anything else I can do to help make it clearer?

PR #9375, opened 2024-02-21
[11.x] Add section for `make:enum` command

milwad-dev:
I add it to helpers page because I can't find a good place for that.


taylorotwell:
I will look into it.

PR #9380, opened 2024-02-22
Fix broken link to PHPUnit docs

weshooper:
Though, since there's no way of linking to 'latest' (I just looked into that possibility) it may be better to remove the link as it'll break again soon enough.


nunomaduro:
We don't support PHPUnit 11 on Laravel 10.

PR #9381, opened 2024-02-22
[11.x] Fixes session table command

nunomaduro:


taylorotwell:
@nunomaduro I think a few other commands received this treatment.

PR #9396, opened 2024-02-24
[11.x] Add per second to rate limit

milwad-dev:


taylorotwell:
We will document this. Thanks.

PR #9416, opened 2024-02-27
Documenting how to customize Laravel 10. directory paths

giandomenicodisalvatore:
I struggled to find info about how to move the `storage` directory away from the `root`, to the docker volume mount point. 
Some authors suggest bash scripting, others prefer .dockerignore + bind mounting, while others just do it manually.

Reading the code in  `/laravel/framework/src/Illuminate/Foundation/Application.php` I found many *undocumented gems*, that I think may be useful in common scenarios, like deploying the app to shared hostings.

Surprisingly, the Laravel way is so clean and easy that I would like it to be shared in the documentation, *given the right caveats* that I may have missed for my limited experience.

This is the [gist](https://gist.github.com/giandomenicodisalvatore/b0e4310363af2e26fabc82cfc7f0a077).

Thank you for your remarkable work and dedication.

### DIFF
--- a/structure.md
+++ b/structure.md
@@ -26,6 +26,8 @@
     - [The `Policies` Directory](#the-policies-directory)
     - [The `Providers` Directory](#the-providers-directory)
     - [The `Rules` Directory](#the-rules-directory)
+- [Customizing Framework Directories](#customizing-framework-directories)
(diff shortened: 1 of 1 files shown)


taylorotwell:

PR #9440, opened 2024-02-29
[11.x] Adjusts upgrade guide regarding publishing migration

nunomaduro:


upgrade.md
```diff
@@ -471,3 +480,42 @@ Then, in your application's `config/sanctum.php` configuration file, you should
...
+```shell
+php artisan vendor:publish --tag=passport-migrations
+```
+
+In addition, the password grant type is disabled by default. You may enable it by calling the `enablePasswordGrant` method in the `boot` method of your application's `AppServiceProvider`:
+
+    public function boot(): void
+    {
+        Passport::enablePasswordGrant();
+    }
```
driesvints, line 521: I don't think we need these extra sections on sanctum, passport and telescope. We should direct people to these package's upgrade guide's instead. Doing the above will only duplicate these upgrade steps in two different places. By just pointing people to the package upgrade guides we can add more steps to those guides in the future and don't need to update two different places.

upgrade.md
```diff
@@ -67,18 +67,27 @@ You should update the following dependencies in your application's `composer.jso
 <div class="content-list" markdown="1">
 
 - `laravel/framework` to `^11.0`
-- `laravel/sanctum` to `^4.0`
+- `laravel/sanctum` to `^4.0` (If installed)
 - `laravel/telescope` to `^5.0` (If installed)
+- `laravel/passport` to `^12.0` (If installed)
```
driesvints, line 72: I don't think we need `(If installed)` here next to these packages since it's a bit of a given to only do this if you have these packages installed.

PR #9444, opened 2024-03-02
Update collections.md (unless method)

dilshadafser:
it should be  evaluates to 'false'


nicolus:
The current documentation seems correct to me, as is shown in the example below.

dilshadafser:
image
No, here should be 'false' instead of 'true'

dilshadafser:
> The current documentation seems correct to me, as is shown in the example below.

No, here should be 'false' instead of 'true'   as in screenshot given below

PR #9446, opened 2024-03-04
Adding clarity about the lifetime of asynchronous process calls

brendanpetty:
It appears possible to spawn a process and leave it running, after the Laravel life cycle ends. But this is not actually possible and processes die silently. Need to ->wait() before returning a response to keep those processes alive.


taylorotwell:
It is actually possible using something like:

`Process::start('foo 2>&1 &');

PR #9459, opened 2024-03-06
Better document SQL Server settings in config/database.php

HenkPoley:


HenkPoley:
Wanted to better document how to enable MultiSubnetFailover, this 'missing' option had caused some confusion on my side.

And added the rest of the options just in case.

taylorotwell:
Can we instead just give an example of how setting any configuration option you want using `DATABASE_URL`? I think that is more practical. Then we can remove all of the array based configuration stuff here and link out to Microsoft documentation for all available DATABASE_URL options.

taylorotwell:
Please mark as ready for review when it's ready for me to take a look again.

HenkPoley:
I've slashed the documentation as requested. Is the array syntax deprecated?

🤔 I did find a bug in \Illuminate\Support\ConfigurationUrlParser->getDatabase()

https://github.com/laravel/framework/blob/6c27ca43dbcfd232813ca5aba8ad3c80405ba9b1/src/Illuminate/Support/ConfigurationUrlParser.php#L96..L101

```
    /**
     * Get the database name from the URL.
     *
     * @param  array  $url
     * @return string|null
     */
    protected function getDatabase($url)
    {
        $path = $url['path'] ?? null;

        return $path && $path !== '/' ? substr($path, 1) : null;
    }
```

It receives an `$url['path']` array item like `Server=your_server_address;Database=your_database_name`

But [`substr($path, 1)`](https://www.php.net/substr) is of course not the way to extract the name `your_database_name`. But it ends up just chopping of the first letter 'S' as `erver=your_server_address;Database=your_database_name` (or whatever other ordering you had in the connection string.

This should be more appropriate:

```
    /**
     * Get the database name from the URL.
     *
     * @param array|string $url
     * @return string|null
     */
    protected function getDatabase($url)
    {
        $path = $url['path'] ?? null;

        $pairs = explode(';', $path);

        $databaseName = null;

        // Iterate over each pair
        foreach ($pairs as $pair) {
            // Split the pair into key and value
            list($key, $value) = explode('=', $pair, 2); // Limit to 2 elements to ensure it splits only on the first '='

            // Check if the key matches 'Database', case-insensitively
            if (strcasecmp($key, 'Database') === 0) {
                $databaseName = $value;
                break; // Stop the loop once we've found the database name
            }
        }

        return $databaseName;
    }
```

I'll submit patches. on laravel/framework.

HenkPoley:
Until _something_ is fixed in the handling of DATABASE_URL (for `sqlsrv` ?) please don't merge my most recent documentation update, the previous revision is fine though.

driesvints:
https://github.com/laravel/docs/pull/9388

driesvints:

PR #9476, opened 2024-03-11
Update eloquent.md - UUID/ULID Migration

lmpclk:
The documentation surrounding the UUID's and ULID's refers the reader to ensuring that they have the equivalent UUID/ULID primary key in their migration. Within the documentation however it states 'You should also ensure that the model has a ...', I expect this should read 'You should also ensure that the migration has a...'?

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -205,7 +205,7 @@ Eloquent requires each model to have at least one uniquely identifying "ID" that
 
 Instead of using auto-incrementing integers as your Eloquent model's primary keys, you may choose to use UUIDs instead. UUIDs are universally unique alpha-numeric identifiers that are 36 characters long.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think it's OK. If anything it should read "ensure the table"

lmpclk:
Just to assist with clarity for those with less experience; myself included! Perhaps including something such as:

If you would like a model to use a <UUID/ULID> key instead of an auto-incrementing integer key, you may use the `Illuminate\Database\Eloquent\Concerns\<HasUuids/HasUlids>` trait on the model. Of course, you should ensure that the model has a <[UUID equivalent primary key column](/docs/{{version}}/migrations#column-method-uuid) / [ULID equivalent primary key column](/docs/{{version}}/migrations#column-method-ulid)> by including this in the respective migration.

PR #9493, opened 2024-03-13
fixes version in composer install

matz0r:
Just tried to install it from scratch and was confused it installs laravel 10 while copying the command from the docs.

Should be 11 instead of 10 i guess.


matz0r:
fixed now by nuno.

PR #9505, opened 2024-03-15
fix(Sanctum): Fix config filename.

shehi:
`Broadcast::routes()` statement has no effect when placed inside `routes/api.php` config file. It needs to be placed inside `routes/channels.php` config file.


taylorotwell:
That's not quite right either. Fixed! Thanks.

shehi:
FYI my solution worked. :) Problem is, Broadcasting still auto-adds a route to `web` route group, which I discovered after sending this PR several minutes ago. Came back to see how to override that behaviour and saw your changes. You are fast!

PR #9506, opened 2024-03-15
feat: Add `prefix` value to `Broadcast::routes()` usage.

shehi:
The subsequent `Echo`-related code uses `/api/broadcasting/auth` endpoint for authentication. It'd make more sense to show how `/api` route prefix is added.


shehi:
@taylorotwell , you enhanced my previous PR and applied it. Subsequently, this one became somewhat obsolete. Please add the changes suggested in this PR to your fix as well.

PR #9509, opened 2024-03-15
Broadcasting - Remove duplicated content

ricocrivelli:
Correct the issue in the documentation where the Pusher Channels chapter is duplicated and mixed with information from the Reverb.


taylorotwell:
Fixed by another PR. Thanks.

PR #9522, opened 2024-03-21
Fix recipient name order in on-demand notification example

asamofal:
The current documentation incorrectly states the order of the email address and recipient name in the arrays passed to the `Notification::route` and `Notification::routes` methods.

The `route` parameter expects the key in the array to be a name and the value to be an address. Otherwise, there will be an exception:
> Email address "Barrett Blair" does not comply with the addr-spec of RFC 2822.

I've tested it with Laravel 10.x but this example for `route` method in docs is available since v8.x, do we need multiple PRs for every branch then or ...?


taylorotwell:
Copied example in docs - sent perfectly fine:

image

PR #9524, opened 2024-03-21
[11.x] Add new entry on 'hashing' documentation

elzdave:
- Added 'Allowing to Rehash Password Using Different Algorithm' to the 'hashing' documentation

### DIFF
--- a/hashing.md
+++ b/hashing.md
@@ -1,13 +1,18 @@
 # Hashing
 
-- [Introduction](#introduction)
-- [Configuration](#configuration)
-- [Basic Usage](#basic-usage)
+- [Hashing](#hashing)
+  - [Introduction](#introduction)
+  - [Configuration](#configuration)
+  - [Basic Usage](#basic-usage)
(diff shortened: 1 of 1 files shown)


taylorotwell:
Not sure I understand all the other changes.

elzdave:
Pardon me, but the addition of the new `verify` option on `hashing` config file, contained in the Laravel v11.x, can affect the behaviour expected by the developer.

On the previous version (v9.x-v10.x), while the `verify` option were evaluated at Argon/Bcrypt hasher class, the default `hashing` config file did not contain the `verify` option at `hashing.bcrypt` and `hashing.argon` options, so the hasher using default predefined value of `false`. On v11.x, if we published the `hashing` config file, it now contain the `verify` option set to `true`. While this option seems harmless on previous version, it will raise exception if the developers change the hash driver on the fly, along with `rehash_on_login` set to `true`.

As this `verify` option didn't mentioned anywhere in the documentation, would it be nice to add some documentation on how this option work?

Just my 2 cents :)

PR #9534, opened 2024-03-26
Added the correct command to generate the queue batches table

jalexmelendez:
Changed the 'php artisan make:queue-batches-table' command to 'queue:batches-table'


taylorotwell:
CleanShot 2024-03-26 at 08 18 45@2x

jalexmelendez:
You´re right! I was using an older version of the framework, thought i was using v11...

PR #9535, opened 2024-03-26
Re-add excludeUnvalidatedArrayKeys() to validation.md

Grldk:
I just noticed that the mention of `excludeUnvalidatedArrayKeys()` has disappeared from the docs in `9.x` and newer, but could not find any reason for this omission. As far as I can see the code still works as documented, so I've re-added the paragraph from the `8.x` docs.


taylorotwell:
It already defaults to excluding them.

PR #9550, opened 2024-04-02
Modify-db-connection-name-example-for-multiple-connections

RoyHridoy:
In multiple db connections or not default db connection's example, I think, usages of `mysql` or other db name instead of `sqlite` will be easy to understand cause 11.x uses `sqlite` as default db.


taylorotwell:
It's fine.

PR #9559, opened 2024-04-04
Fix an invalid `sometimes` validation

mahmoudmohamedramadan:
I think that the field must not have the following validations together: `sometimes`, `required`, and `email`. Yeah, It's confusing! because `sometimes` validation works when we present or do not present the field but, in case we did not present a value the request will hit the `required` validation and then will fire a validation error like so with `email` validation.

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -1917,7 +1917,7 @@ Alternatively, you may use the `exclude_unless` rule to not validate a given fie
 In some situations, you may wish to run validation checks against a field **only** if that field is present in the data being validated. To quickly accomplish this, add the `sometimes` rule to your rule list:
 
     $v = Validator::make($data, [
-        'email' => 'sometimes|required|email',
+        'email' => 'sometimes',
     ]);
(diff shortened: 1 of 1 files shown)


validation.md
```diff
@@ -1917,7 +1917,7 @@ Alternatively, you may use the `exclude_unless` rule to not validate a given fie
 In some situations, you may wish to run validation checks against a field **only** if that field is present in the data being validated. To quickly accomplish this, add the `sometimes` rule to your rule list:
 
     $v = Validator::make($data, [
-        'email' => 'sometimes|required|email',
+        'email' => 'sometimes',
```
meduzen, line 1920: I am regularly confused by the official example from the docs, too (and I am [not the only one](https://stackoverflow.com/questions/35839414/laravel-sometimes-vs-sometimesrequired)), thou I think the edit should have been be:

```suggestion
        'email' => 'sometimes|email',
```

PR #9571, opened 2024-04-10
[11.x] Add `integer` section to HTTP Requests

avosalmon:
Add documentation for retrieving integer input values from requests using [$request::integer()](https://github.com/laravel/framework/blob/d63d2b1f50b6bbf8beb88b7b4ad5a4d230244443/src/Illuminate/Http/Concerns/InteractsWithInput.php#L348-L358) method.
This method is quite useful but not documented.


taylorotwell:
No plans to document this atm. Feels simpler to just cast to int.

PR #9574, opened 2024-04-11
[11.x] removes string backed enum type we also support int

monurakkaya:


taylorotwell:
Drafting until released.

PR #9587, opened 2024-04-17
number helpers require php intl extension

saeedvaziry:
I couldn't find in the docs that the `Number` helper requires the PHP `intl` extension


taylorotwell:
An exception is thrown informing you of this.

PR #9589, opened 2024-04-18
[11.x] Document when model is included in a collection applying `toArray()` in collections.md

iamwebwiz:
This PR adds sample responses for when a model is included in a collection and the `toArray()` method is applied to it.

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -2770,14 +2770,40 @@ The `toArray` method converts the collection into a plain PHP `array`. If the co
 
     $collection->toArray();
 
+    // ['name' => 'Desk', 'price' => 200]
+
+When a model is included in the collection:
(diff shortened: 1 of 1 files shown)


taylorotwell:

iamwebwiz:
Hey @taylorotwell , this is not a new code to the framework, I think that having this written explicitly in the docs makes it easier for devs to visualize what they will get, and more value-added if the dev is pretty new to Laravel. I would've found this part of the doc helpful as at 9.x when I needed it, so, I thought it would be helpful for others.

The bug was in the initial part of the doc where the response of the collection was expressed as a multidimensional array when the result is not. This PR addresses that plus the addition I made to it.

PR #9605, opened 2024-04-25
Clarify timeout value unit in `Http::timeout()` method

bymaul:
Updated the documentation for the `timeout` method in HTTP client to explicitly state that the timeout value is specified in seconds.


taylorotwell:
We already do?

> The `timeout` method may be used to specify the maximum number of seconds to wait for a response.

PR #9607, opened 2024-04-28
[11.x] Add note for `make:provider` command

milwad-dev:
When the user uses the `make:provider` command to create a provider, the provider is automatically registered in the `providers.php` file.


milwad-dev:
I already mentioned this tip in the section for registering providers.

PR #9610, opened 2024-04-29
Update sail.md, add mariadb

robertsilen:
Add mariadb to sail doc, with the logic that MariaDB is mentioned elsewhere, e.g. https://laravel.com/docs/11.x/database


taylorotwell:
I think it's OK.

robertsilen:
@taylorotwell Was this closed without merging for some reason? Wondering when the change will show up in e.g. https://github.com/laravel/docs/commits/11.x/

PR #9616, opened 2024-04-30
Fixing method name and example code under Price Preview section

haider00125:
Fixing the method name from `productPrices` to `previewPrices` and example code in **Price Preview** section.


haider00125:
Hey @taylorotwell 
Same issue exists in 10.x as well. Should i open this for 10.x too?

PR #9619, opened 2024-04-30
[11.x] Add `Exceptions` to facade table

milwad-dev:


taylorotwell:
Can you add it without changing the entire table so the diff is easier to read?

PR #9629, opened 2024-05-03
Update Reverb for nginx sub directory

devfake:
I hope that the change is acceptable. I spent several hours on this problem, although the solution was relatively simple.


taylorotwell:
This feels very specific to Nginx.

PR #9642, opened 2024-05-07
Sail alias ignores args for local sail

ulcuber:


ulcuber:
@taylorotwell 

From `man bash` -> ALIASES:

> There is no mechanism for using arguments in the replacement text. If arguments are needed, a shell function should be used (see FUNCTIONS below).

Alias is just a prefix string substituted instead of first word in command.

So `sh vendor/bin/sail` part receives all arguments while `sh sail` part receives nothing

For example the final expression for `sail up` would be like `([ -f sail ] && sh sail) || (sh vendor/bin/sail up)`. So instead of expected `./sail up` it would be just `./sail`

ulcuber:
It was fixed before in newer version by:
https://github.com/laravel/docs/pull/9338

PR #9650, opened 2024-05-11
[11.x] Add documentation for DeleteWhenMissingModels attribute

Neol3108:


taylorotwell:
We're going to leave the property there for now because it's consistent with the rest of the framework. I think we're exploring a more consistent approach to attributes before we document just one.

shaedrich:
This might be it: #9967

PR #9652, opened 2024-05-13
Pulse thresolds

timacdonald:
Document pattern based threshold support.

https://github.com/laravel/pulse/pull/340

### DIFF
--- a/pulse.md
+++ b/pulse.md
@@ -304,12 +304,40 @@ The `SlowJobs` recorder captures information about slow jobs occurring in your a
 
 You may optionally adjust the slow job threshold, [sample rate](#sampling), and ignored job patterns.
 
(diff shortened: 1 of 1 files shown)


pulse.md
```diff
@@ -304,17 +304,45 @@ The `SlowJobs` recorder captures information about slow jobs occurring in your a
...
+],
+```
+
+If no regex patterns match the request's URL, then the `'default'` value will be used.
+
 You may also configure URL grouping so that similar URLs are grouped as a single entry. For example, you may wish to remove unique IDs from URL paths or group by domain only. Groups are configured using a regular expression to "find and replace" parts of the URL. Some examples are included in the configuration file:
 
 ```php
-Recorders\OutgoingRequests::class => [
+Recorders\SlowOutgoingRequests::class => [
```
timacdonald, line 345: Fixed an incorrect class here.

PR #9658, opened 2024-05-13
[11.x] Add `Rule::array()` in validation

milwad-dev:
This method is beneficial, and I think adding it to the documentation is necessary.

The description needs to update!


taylorotwell:
I'll try to work on this. 👍

PR #9660, opened 2024-05-13
Update queues.md

tm69:
updated command for batchables queue table


taylorotwell:
CleanShot 2024-05-13 at 16 52 24@2x

PR #9663, opened 2024-05-17
updating the laravel socialite docs of redirection url for stateless api

iamsuzon:


taylorotwell:
Not really a full explanation of why you might want to do this.

PR #9664, opened 2024-05-17
Add section on how to handle pivot models with custom table names

edwwaarrdd:
Add a section to the docs on the option to pass your pivot model class as the $table parameter in the belongsToMany() function as discussed here: https://github.com/laravel/framework/issues/51410


taylorotwell:
Will try to incorporate this.

PR #9665, opened 2024-05-19
[11.x] docs(queue): Add docs for forcing the default connection

victorelec14:
Added documentation to specify that the `queue:work` command can be forced to use the `default` connection. Two examples were provided: one using `null` and another using an empty string.

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -1695,6 +1695,12 @@ By default, the `queue:work` command only processes jobs for the default queue o
 php artisan queue:work redis --queue=emails
 ```
 
+It is also possible to force the command to use the `default` connection in this way:
+```shell
+php artisan queue:work null --queue=emails
+php artisan queue:work "" --queue=emails
+```
+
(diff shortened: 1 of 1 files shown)


victorelec14:
Hi @taylorotwell ,

I think it would be a good idea to document this. On some deployment websites (or in scripts), there are times when the "connection name" is a required field and cannot be left empty. Because of this, some people create a new connection without realizing that they can enter "null" as the connection name to use the default value.

For example, in the case of Ploi, it doesn’t allow you to leave it empty, but very few people know that entering "null" will select the default connection. Even the developer of Ploi wasn’t aware of this until I opened a support ticket.

image


This is not the first time I have seen this issue. It also happens in scripts that pass an empty string "" or use different commands depending on whether there is a value or not.

I think it would be good to reconsider this.

Thank you.

Docs: https://laravel.com/docs/11.x/queues#specifying-the-connection-queue

PR #9667, opened 2024-05-21
[11.x] Fix code styling in packages.md

KamilWojtalak:
Fix code styling at packages.md, just to be consistent


taylorotwell:
We never use spaces for concatenation. If it's inconsistent we should prefer no spaces.

PR #9672, opened 2024-05-23
Added info on how to reset multiple connections when using the RefreshDatabase trait

hn-0:
In testing, it's possible to refresh multiple connections by setting the test class' `connectionsToTransact` property to an array of connection name strings. I've added some info to highlight this, as I noticed that someone asked about it [on the forums](https://laracasts.com/discuss/channels/laravel/multi-database-tests).


taylorotwell:
Will try to work this in the more typical fashion.

PR #9684, opened 2024-05-31
Change Carbon/Dates Likelihood Of Impact to High

piotr-lottestyle:
Hi. I'm reopening [pull request #9678](https://github.com/laravel/docs/pull/9678) due to significance of this, please reconsider this:

In my opinion this is very significant change:

After changes this code broke our appllicaiton:

`return $time->diffInMinutes(now(), true) > 10;`

because `diffIn` methods returns float instead of integer

Hotfix:

`return ((int) $time->diffInMinutes(now(), true)) > 10;`

I was in process of upgrading Laravel from 6 up to 11 and this single change had "Medium" likelihood impact in docs but it was serious for our project and we spent lot of time looking for problem and we had problem with our tests. 

Please consider changing Likelihood to High to increase awareness of this problem. This small change may save lot of time for other developers.

https://github.com/briannesbitt/Carbon/issues/2986
https://laravel.com/docs/11.x/upgrade#carbon-3

@taylorotwell Laravel codebase is in fact affected by this:

This code makes no sense:

https://github.com/illuminate/cache/blob/master/Repository.php#L554

Duration is always > 0

Please reconsider changing it to high. It may affect many people. Changing single word in docs can save lot of time for hundreds of teams as it would save our time. I readed Laravel Upgrade 11 notes carefuly but due to high time pressure in our projects I had to rush things and I didn't put enough weight to parts of documentation with Likelikehood impact scored "Medium". Such simple change of  wording "Medium" to  "High" could put developers on alert at least. It can save some frustration. And time.

getSeconds

The more i think about it gets worse. Dates function are crucial to applications. it's very difficult to detect such problems because we are dealing with time. They may surface up after long time.

Security, caching, sessions and more.

I had bad experience with floating-point problems but this seems far much worse.

Carbon 2: `diffInSeconds() returns 0`

Carbon 3: `diffInSeconds() returns 0.00001`

So if `$duration = now()->diffInSeconds($someTimestamps)`

So every statement in every application with `$duration > 0` will break.

Example above (from Laravel codebase) it might be trivial but might cause huge loss of time for many users. Impact might be huge and cause lot of stress to many dev teams

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -412,7 +412,7 @@ public function scalar($query, $bindings = [], $useReadPdo = true);
 <a name="carbon-3"></a>
 #### Carbon 3
 
-**Likelihood Of Impact: Medium**
+**Likelihood Of Impact: High**
 
 Laravel 11 supports both Carbon 2 and Carbon 3. Carbon is a date manipulation library utilized extensively by Laravel and packages throughout the ecosystem. If you upgrade to Carbon 3, be aware that `diffIn*` methods now return floating-point numbers and may return negative values to indicate time direction, which is a significant change from Carbon 2. Review Carbon's [change log](https://github.com/briannesbitt/Carbon/releases/tag/3.0.0) for detailed information on how to handle these and other changes.
 


piotr-lottestyle:
To support this I found some random code in google:

https://coveralls.io/builds/16955852/source?filename=src%2FTraits%2FCaching.php#L163

Screenshot_2024-05-31_13-31-38

piotr-lottestyle:
And another one:

https://forum.laravel-livewire.com/t/livewire-duplicatelogin-middleware/3765/2


Screenshot_2024-05-31_13-33-44


A lot of people can be affected by this. This is just few first examples from google

michaeldyrynda:
Given Laravel 11.x supports either Carbon 2.x or 3.x, does downgrading to Carbon 2.x sufficiently address the issue?

Medium impact seems a reasonable severity, given the upgrade is optional.

piotr-lottestyle:
@michaeldyrynda there is huge risk some developer teams can miss it causing potential hazard in security etc.

taylorotwell:
It's already in the table of contents as a change you need to review. 

CleanShot 2024-05-31 at 11 25 42@2x

pb30:
>  there is huge risk some developer teams can miss it causing potential hazard in security etc.

This is true for everything in the upgrade guide. DBAL removal is listed as "Likelihood Of Impact: Low". But if your application happens to use those DBAL methods then obviously your personal impact is high. For you the actual impact was high, for me across 5 projects it wasn't an issue, none of them used the methods.

Not every application uses `diffIn*`, so it is not as universally applicable as say modifying columns in migrations, which is likely very common amongst any type of application.

Nor is upgrading Carbon even suggested in the "Upgrading Dependencies" section. You mentioned you used Shift, and Shift even posts an extra comment highlighting this change:

> ⚠️ Laravel 11 has a nesbot/carbon dependency of ^2.72|^3.0. Carbon 3.0 contains some breaking changes, most notably to the diffIn* methods. Shift detected you are using these methods within your application. While you may try to downgrade to Carbon 2, Shift recommends you review the [Carbon migration notes](https://carbon.nesbot.com/docs/#/upgrade-3.0) and complete the upgrade to Carbon 3.0.

It sucks to get bit when upgrading, but there's only so much upgrade docs can do, especially if you are rushing.

PR #9685, opened 2024-05-31
11.x - Add Controller Middleware section to upgrading

davidrushton:
In Laravel 10.x the logic was `$this->middleware` for controllers, but this is causes a "Call to undefined method" exception in Laravel 11.x.

Added a section to upgrading docs to highlight this change with code example for new style and link to docs.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -208,7 +208,7 @@ public function dump(...$args);
 
 **Likelihood Of Impact: Medium**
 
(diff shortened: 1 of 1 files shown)


browner12:
yes and no. this is only if you update your scaffolding, which the docs recommend against.

as long as your `Controller` still extends `Illuminate\Routing\Controller`, you're still okay to use the `middleware()` method in Laravel 11.

PR #9687, opened 2024-06-02
Nginx: Parse all PHP files using Octane to avoid reavealing php sourcecode

colorando-de:
I had no chance to chek this on my side yet, but @fideloper suggested to update the Nginx config proposal according to the discussion and suggestion in the this octane issue: https://github.com/laravel/octane/issues/900#issuecomment-2140520600

With the current documented config, PHP files in public folder other than index.js will just be downloaded as sourcecode. Instead of this, they should be parsed via Octane.

### DIFF
--- a/octane.md
+++ b/octane.md
@@ -283,12 +283,12 @@ server {
 
     charset utf-8;
 
-    location /index.php {
-        try_files /not_exists @octane;
-    }
-
     location / {
         try_files $uri $uri/ @octane;
+    }
+
+    location ~ \.php$ {
+        include octane.conf;
     }
 
     location = /favicon.ico { access_log off; log_not_found off; }


fideloper:
Hi!

I might be missing something (i’m checking on my phone) but I think this doesn’t really make sense as it’s referencing/including file `include octane.conf;` which doesn’t seem part of the documented example. IIRC that include like also replaced the content within the `location @octane` section, which I didn’t see reflected here.

fideloper:
(this is closed just to keep things clean, feel free to respond)

colorando-de:
@fideloper I agree, but as you noted, documenting the octane section extraction is difficult. It seems separate from the original issue. 

Can we open a new issue to separate the octane section into an extra file? This way, we can address the security fix now and discuss the other topic further.

PR #9705, opened 2024-06-13
Update broadcating.md

KakkoiDev:
Hi there, thank you very much for Laravel, it's a great project!

I noticed while implementing Broadcasting that the docs indicated to install `laravel-echo` and `pusher-js` with: `npm install --save-dev laravel-echo pusher-js`.

However, I think this is a mistake because those packages will not be included into the production build.

So I replaced them with this command instead: `npm install --save laravel-echo pusher-js`
Actually `--save` is optional and the command `npm install laravel-echo pusher-js` would do exactly the same thing.

For more information on the difference between `--save` and `--save-dev`: https://stackoverflow.com/a/42206389/12818567

Have a great day!


taylorotwell:
I think this actually works fine.

CelDaemon:
@KakkoiDev I'm aware of this issue already being closed, but `--save-dev` would still be accurate as the vite build step is considered to be part of the dev dependencies, those packages are no longer needed after the build step and are therefore considered dev dependencies.

PR #9708, opened 2024-06-15
Correct the composer import for Resend and clairfy the configuration changes

ssmusoke:


taylorotwell:
Can you resubmit this with a description? The Resend Laravel SDK is a bit different and we just use the plain PHP client for our lightweight integration.

PR #9713, opened 2024-06-19
Update upgrade.md to include Scout dependency

HarryLee186:

PR #9714, opened 2024-06-19
Update upgrade.md to include Scout dependency

HarryLee186:
- Apologies, I put the wrong release in on [this PR](https://github.com/laravel/docs/pull/9713), this fixes that


taylorotwell:
@driesvints is this necessary?

driesvints:
Yeah, v10 is the minimum one. Although I don't feel we should list each package per se.

PR #9716, opened 2024-06-20
Update artisan.md

ekadesign:
By default is not allowed to use withCommands because of configure() method
You have to call builder and build your application instance manually. Because configure() doesn't provide any data to App instance

e.g.
public static function configure(?string $basePath = null)
    {
        $basePath = match (true) {
            is_string($basePath) => $basePath,
            default => static::inferBasePath(),
        };

        return (new Configuration\ApplicationBuilder(new static($basePath)))
            ->withKernels()
            ->withEvents()
            ->withCommands()
            ->withProviders();
    }

Also I didn't found any other ways to register it as was written in document

### DIFF
--- a/artisan.md
+++ b/artisan.md
@@ -655,19 +655,31 @@ Sometimes, you may need more manual control over how a progress bar is advanced.
 <a name="registering-commands"></a>
 ## Registering Commands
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Works fine for me.

ekadesign:
@taylorotwell argument $commands in `withCommands` method will be empty any time, no matter what do you send to it while using configure. Checked twice. If you want to I just can create issue

PR #9718, opened 2024-06-20
Add crucial note about folder permissions to deployment.md

enderandpeter:
This PR is related to #9710. @taylorotwell expressed that he was not in disagreement with the veracity of this info, but did not think [structure.md](https://laravel.com/docs/11.x/structure) was a good place for this and was unsure if this was important enough to mention.

Please see #9710 for the **Problem**, **Solution**, and **Notes** sections that explain why these docs should inform devs there are two folders, `storage` and `bootstrap/cache`, that the framework needs to write into.

Also, take a look at [this guide on setting up permissions on a Linux web server](https://serverfault.com/questions/357108/what-permissions-should-my-website-files-folders-have-on-a-linux-webserver) that explains more about secure user and folder permissions in such an environment.

I've been working with this framework for some time. Over and over I see junior and senior devs set these folders to `777`. Enough is enough! 🙂


enderandpeter:
Awesome! Thank you so much!

PR #9719, opened 2024-06-21
Describe the purpose and function of `queue:work --memory`

spawnia:
The implementation of `--memory` is actually really clever. I found it to be very useful to prevent potential memory leaks. However, its description does not fully describe what it does.

My first interpretation of its function was that it sets a limit on the amount of memory that individual jobs may use. Thus, I set it to a value that equals [the PHP INI setting `memory_limit`](https://www.php.net/manual/de/ini.core.php#ini.memory-limit). This setting essentially made the option useless, as the worker now kept running until it was right at the cusp of using up all memory. Subsequently, the next job to start processing would fail.

I found that it is probably best to just leave the default, unless your queue worker uses a lot of memory while idle. If this were the case, the queue worker would always stop after processing a single job, causing unnecessary restarts. For most applications, 128MB should be sufficient.

In conclusion, I think it is valuable to describe this option in more detail. Even though most users will not have to modify it, there is potential for misunderstanding is purpose, leading to a misconfiguration that actually makes your queue worker less reliable.

I also updated the description of the option in https://github.com/laravel/framework/pull/51873.

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -1710,6 +1710,15 @@ The `--max-jobs` option may be used to instruct the worker to process the given 
 php artisan queue:work --max-jobs=1000
 ```
 
+<a name="handling-memory-leaks"></a>
+#### Handling Memory Leaks
+
+The `--memory` option specifies the amount of memory in megabytes that an idle queue worker may use before stopping. This acts as a failsafe if your jobs leak memory and continually drive up memory usage. The default value of 128 may be modified if your application generally uses a lot of memory while idle:
+
+```shell
+php artisan queue:work --memory=256
+```
+
 <a name="processing-all-queued-jobs-then-exiting"></a>
 #### Processing All Queued Jobs and Then Exiting
 


taylorotwell:
Hmm, this is described in `php artisan help queue:work`.

spawnia:
`php artisan help queue:work` gives me the following information:

```
--memory[=MEMORY]      The memory limit in megabytes [default: "128"]
```

This description does not tell me many critical pieces of information:
- What exactly is limited? (I assumed it was the queue jobs, which was the wrong assumption to make)
- What happens when the memory limit is hit?
- When would I want to change the default?

I think this lack of detail may lead to misuse of the option, which can be actually harmful and contrary to its purpose. This is why I believe improving the documentation on it is valuable.

driade:
Hi. I support @spawnia proposal. It's 10 years working with Laravel and didn't know the option. It would be very nice if it'd be documented.

PR #9720, opened 2024-06-21
Add docs for withAggregate method

gigante762:
We are missing a very good feature, the withAggregate method


taylorotwell:
No plans to document this atm.

PR #9722, opened 2024-06-26
Add Note for Laravel Installer when using with Valet

askdkc:
Related to this issue: https://github.com/laravel/installer/issues/347

Since the Laravel Installer added the TLD check feature in version 5.8.0, it starts asking for the user password every time you run `laravel new myproject`. This confuses some users because the installer suddenly asks for the user password without explaining the reason. Adding this note will address that issue.

### DIFF
--- a/installation.md
+++ b/installation.md
@@ -72,6 +72,9 @@ composer global require laravel/installer
 laravel new example-app
 ```
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't think I really want this right in the install docs.

PR #9724, opened 2024-06-27
Add transaction isolation level configuration info into database.md

savinmikhail:
Me and my team were able to find how to set the transaction isolation level only by exploring the internal codebase, so adding it to docs might be beneficial.


taylorotwell:
I will need to format this to bring it inline with the rest of the docs when I find the time. 👍

PR #9742, opened 2024-07-06
[11.x] Add `pairs` method to Number

milwad-dev:


taylorotwell:
Hmm, I think we may need a better explanation here. 🤔

milwad-dev:
Yes, I don't have any idea about the `pairs` method explanation. I wrote a full description of it.

PR #9753, opened 2024-07-12
[11.x] Fixes incorrect reference to validate and replaces it with verify

john-shipp:
In the upgrade guide, validate should be replaced with verify for the Sanctum upgrade section.


taylorotwell:
Validate is correct.

PR #9754, opened 2024-07-12
Update typo make:queue-table > queue:table

adnane-ka:
Update typo `make:queue-table > queue:table`


maru0914:
@adnane-ka 
I think `make:queue-table` is correct.
See https://github.com/laravel/docs/pull/9554

PR #9755, opened 2024-07-13
[11.x] Add `Controller` & `Request` to structure

milwad-dev:


taylorotwell:
I think it's OK to just describe this under HTTP.

PR #9757, opened 2024-07-14
Add examples to retrieve translations from a path or a package

carstenjaksch:
As a new user, I had a rough time figuring out how to retrieve translation strings in case of the added examples.

I think the documentation was misleading, because it mentioned the dot notation only.

### DIFF
--- a/localization.md
+++ b/localization.md
@@ -150,6 +150,14 @@ You may retrieve translation strings from your language files using the `__` hel
 
     echo __('messages.welcome');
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I've never really seen people use translations this way.

carstenjaksch:
@taylorotwell Filament does that. And I think all possible scenarios should be documented.

Isn't that the goal of the already great Laravel docs?

PR #9760, opened 2024-07-15
Add API Throttling middleware in Routing

TENIOS:
I'd like to add a Throttling API section to Routing as is done with Redis to make it easier for beginners and not implement their own designs. I found this method myself by looking at how `getMiddlewareGroups()` works in `withMiddleware()` method. I don't know exactly how to describe it, but I made a base for creating this section. Please flesh it out as it should be or create a new PR.

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -845,6 +845,19 @@ Rate limiters may be attached to routes or route groups using the `throttle` [mi
         });
     });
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Would need to fit style of rest of docs. Will try to take a look. 👍

PR #9773, opened 2024-07-18
Add Intl Extension to Deployment

TENIOS:
This extension is required for the [Number](https://laravel.com/docs/11.x/helpers#numbers-method-list) methods and for Laravel Cashier (Stripe/Paddle) packages and it is also used for validation `dns` and `spoof` in [`email` rule](https://laravel.com/docs/11.x/validation#rule-email).


taylorotwell:
Not required to use the entire framework.

PR #9777, opened 2024-07-18
Add documentation for getOrPut Collection method

antergos98:
Hi Taylor and the rest of the Laravel team.

I noticed that there was no documentation for the awesome getOrPut method on the docs so here is my contribution. I used this method a couple of times and I think it's really convenient and wish people who read the docs can discover it.

I tried to document it the best that I can while keeping it short like the other code blocks on the same page.

Thank you!


taylorotwell:
I personally don't feel collections should be used to represent single objects so leaving this one our for now. 😅

PR #9789, opened 2024-07-25
Update routing.md

webuxmotion:
Fix the semantic error


webuxmotion:
I confused the sign with the sign that precedes the type

PR #9794, opened 2024-07-26
Update routing hyperlinks

TENIOS:
I deleted the unnecessary link and tweaked the existing ones a bit.


taylorotwell:
I don't really want to break links if avoidable.

TENIOS:
The file uses a second tag `a` which does not work and is not used anywhere.

PR #9796, opened 2024-07-27
Update required SQLite version

gdarko:
Laravel 11 does not work correctly if older SQLite3 is installed.

As documented in [master](https://laravel.com/docs/master/database#introduction), v3.35.0+ works well.


hafezdivandari:
Laravel 11.15+ has extended support for SQLite 3.26+, please check: #9768

PR #9810, opened 2024-08-07
Include Str::transliterate method

Rattone:


taylorotwell:
Can you keep the docs in alphabetical order?

Rattone:
I tried to respect more hidden conventions and I didn't notice the alphabetical order :-)

The method description could be improved. The method used behind the scenes is described like this in the voku package:

_Returns an ASCII version of the string. A set of non-ASCII characters are replaced with their closest ASCII counterparts, and the rest are removed unless instructed otherwise._

Do I need to document the $unknown and $strict parameters as well?

PR #9811, opened 2024-08-08
Added info on ignoring case for "contains" and "containsAll" methods

adhuham:
Hi!
I've added information and code examples on how to ignore the case when using "contains" and "containsAll" methods.

Thank you!


taylorotwell:
Can you use named arguments?

taylorotwell:
Please mark as ready for review when the requested changes have been made.

PR #9820, opened 2024-08-14
fix: avoid page overflow with long table rows

hichemfantar:
continuation of #8448


hichemfantar:
@taylorotwell was i supposed to merge this into master to make sure it continues into future versions or is everything good?

PR #9823, opened 2024-08-15
Fix trustProxies usage in Sail configuration

cbaconnier:
The example doesn't work since the TrustProxies Middleware doesn't expect the wildcard (`*`)  to be in an array: 

https://github.com/laravel/framework/blob/0b85e157ca53d7002aead978d597e4a1ad37c944/src/Illuminate/Http/Middleware/TrustProxies.php#L70

### DIFF
--- a/sail.md
+++ b/sail.md
@@ -460,9 +460,7 @@ sail share
 When sharing your site via the `share` command, you should configure your application's trusted proxies using the `trustProxies` middleware method in your application's `bootstrap/app.php` file. Otherwise, URL generation helpers such as `url` and `route` will be unable to determine the correct HTTP host that should be used during URL generation:
 
     ->withMiddleware(function (Middleware $middleware) {
-        $middleware->trustProxies(at: [
-            '*',
-        ]);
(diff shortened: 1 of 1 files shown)


taylorotwell:
That named argument doesn't even accept a string in its type hint.

cbaconnier:
@taylorotwell You may have overlooked too fast. It does accept string (see link below) and using the wildcard in an array will simply (silently) not work, because the check expect the string `*` and not an array containing `*`

https://github.com/laravel/framework/blob/0b85e157ca53d7002aead978d597e4a1ad37c944/src/Illuminate/Foundation/Configuration/Middleware.php#L622-L640

https://github.com/laravel/framework/blob/0b85e157ca53d7002aead978d597e4a1ad37c944/src/Illuminate/Http/Middleware/TrustProxies.php#L151-L160

cbaconnier:
Also, it uses the correct syntax in the requests documentation: https://laravel.com/docs/11.x/requests#trusting-all-proxies

PR #9834, opened 2024-08-21
Remove canSelectAll example from multiselect

pelmered:
It looks like this parameter was reverted and replaced with a keyboard shortcut. See the PR here: https://github.com/laravel/prompts/pull/147

This option does not exist in the helper signature at all. Se: https://github.com/laravel/prompts/blob/main/src/helpers.php#L59

### DIFF
--- a/prompts.md
+++ b/prompts.md
@@ -421,16 +421,6 @@ $categories = multiselect(
 );
 ```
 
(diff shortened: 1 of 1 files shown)


pelmered:
> Whoops, maybe I should have updated the PR description to say the implementation has changed. Thanks @pelmered!

No, problem. Thank you for this feature!

PR #9850, opened 2024-08-26
[6.x] Use classes instead of IDs

timacdonald:
Collection and helper function heading tags are currently broken for several older versions of the docs. This PR targets `6.x`.

On server render, the heading tags all contain the same literal `id="collection-method" attribute; IDs are meant to be unique.

```html
<h4 id="collection-method"><code>random()</code></h4>
```

When the JavaScript kicks in, the ID attribute is updated its corresponding `<a name="...">` link, however the intended styling is then broken because it targets the `id` attribute, e.g., `#collection-method { ... }`.

- Matches the later documentation versions
- Gets the intended styling applied
- Has no flash of styled / unstyled content

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -199,17 +199,17 @@ For the remainder of this documentation, we'll discuss each method available on 
 ## Method Listing
 
 <style>
-    #collection-method code {
+    .collection-method code {
         font-size: 14px;
     }
 
-    #collection-method:not(.first-collection-method) {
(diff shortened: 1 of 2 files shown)


Rattone:
@timacdonald but is it normal that the error is still present in the v6 documentation? All the method names have the same id and the anchor links don't work. 
The same problem is also present in previous doc versions and in the past this behavior worked

PR #9871, opened 2024-08-30

BernardinoSlv:
the example is not correct as it is only allowing comments on posts from banned users


taylorotwell:
The example is correct.

PR #9876, opened 2024-09-03
Add section for customizing email verification link expiration

Rattone:


taylorotwell:
I think it's OK as is. 👍 We never link to H4s.

PR #9880, opened 2024-09-06
Fix HasOneThrough examples

johankrijt:
Fixed the HasOneThrough examples in the Eloquent Relationships section, as the relations currently are the other way around compared to the database structure.


taylorotwell:
It seems fine to me?

johankrijt:
> It seems fine to me?

@taylorotwell If you look at the schema it shows an `owner` can only have one `car`, and a `car` can have only one `mechanic`, while the relationship and written description is the other way around as it show a `mechanic` has a has-one-through relationship `carOwner`. So either the model and description, or the schema is wrong.

PR #9881, opened 2024-09-06
Document `Response` "stream" and "streamJson" methods

pedroborges:
Support for [streamed JSON Response](https://github.com/laravel/framework/pull/49873) was added in 10.x but it hasn't been documented yet, neither was the `stream` method.

I referenced Symfony's [streamed response docs](https://symfony.com/doc/current/components/http_foundation.html#streaming-a-response) for this contribution.


taylorotwell:
Thanks!

PR #9895, opened 2024-09-13
Add hint about InvokeDeferredCallbacks middleware for old Laravel structure

hasancanboga:
This pull request updates the documentation to include a hint about the potential absence of the InvokeDeferredCallbacks middleware if the project was upgraded from an earlier Laravel version, therefore not having the current Laravel structure, causing the defer() helper to not work.


taylorotwell:
Done.

PR #9914, opened 2024-09-24
Alternative configuration for Typesense schema collections and searchable parameters

robinsxv:
Add an alternative to configure Typesense schema collections and searchable parameters directly in the model instead of in the `config/scout.php`


taylorotwell:
Incorrect formatting.

PR #9922, opened 2024-10-01
changed "one-to-many" header on eloquent-relationships page

Lijahide:
added " / Has Many" to the "One to Many" header and corresponding TOC line.


Lijahide:
I was doing some training with a jr programmer and going over the documentation for Eloquent relationships.
We referenced this page many times and the Jr programmer struggled to find the right section for "has many".

I'm not asking you to take responsibility for this isolated problem. But I do think that it helps to add clarifying language to this section; and it certainly does not hurt.
I think it also makes sense to add this change because the "One to Many (inverse) / Belongs To" section already has this clarifying addition.

PR #9923, opened 2024-10-01
feat: add generic return type examples to Eloquent relationships

canvural:
Hi 👋🏽 

This PR adds generic return types to Eloquent relation examples. Laravel 11 recently got generic types for relationships. So this PR shows how people can define those types.


taylorotwell:
Our syntax highlighting on the website doesn't handle this totally well. We'll need to fix that first.

PR #9925, opened 2024-10-02
Add task scheduling to upgrade.md

FlorianLepelley:
Unnoticed during the migration to L11, task scheduling stopped working in the application as it was using the `app/Console/Kernel.php` file that is no longer supported.
Adding a warning in the upgrade guide should help others avoiding the troubles.


decadence:
That's not true AFAIK

taylorotwell:
Probably missed a step during migration.

PR #9930, opened 2024-10-04
Add warning about restoring default Meilisearch index settings

sgilberg:
I discovered the hard way that if I overwrote a default setting for an index, e.g. `'proximityPrecision' => 'byAttribute'` and used the artisan command to sync it, removing that setting and syncing again doesn't automatically restore the default, I have to explicitly set `'proximityPrecision' => 'byWord'` for that default to be restored. I'm not sure if my wording here captures that in the clearest way so feel free to edit as you see fit.

### DIFF
--- a/scout.md
+++ b/scout.md
@@ -323,6 +323,9 @@ After configuring your application's index settings, you must invoke the `scout:
 php artisan scout:sync-index-settings
 ```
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Let me think about this one!

PR #9932, opened 2024-10-07
Add FrankenPHP Configuration Instructions

olliescase:
This PR adds a new section to the Octane documentation explaining how to configure PHP settings for FrankenPHP. It instructs users to create a php.ini file in the project root to override system-wide PHP settings and includes a reminder to restart the FrankenPHP worker for changes to take effect.

### DIFF
--- a/octane.md
+++ b/octane.md
@@ -137,6 +137,23 @@ services:
 
 You may consult [the official FrankenPHP documentation](https://frankenphp.dev/docs/docker/) for more information on running FrankenPHP with Docker.
 
+<a name="frankenphp-configuration"></a>
+## Configuring FrankenPHP
+
+When using FrankenPHP with Laravel, note that the default system `php.ini` does not apply to the FrankenPHP worker environment. To configure PHP settings, create a `php.ini` file in your project's root directory:
+
+```ini
+# php.ini
+memory_limit = 512M
+upload_max_filesize = 100M
+post_max_size = 100M
+```
+
+FrankenPHP will use these settings for your Laravel application, allowing environment-specific configurations without affecting system-wide PHP settings.
+
+> [!NOTE]
+> Restart your FrankenPHP worker for changes to take effect.
+
 <a name="roadrunner"></a>
 ### RoadRunner
 


Cosnavel:
This really should be added to the docs. @olliescase and I searched hours to find a solution to increase frankenphp's memory limit

octane.md
```diff
@@ -137,6 +137,23 @@ services:
 
 You may consult [the official FrankenPHP documentation](https://frankenphp.dev/docs/docker/) for more information on running FrankenPHP with Docker.
 
+<a name="frankenphp-configuration"></a>
+## Configuring FrankenPHP
+
+When using FrankenPHP with Laravel, note that the default system `php.ini` does not apply to the FrankenPHP worker environment. To configure PHP settings, create a `php.ini` file in your project's root directory:
```
dunglas, line 143: This is not exact. The system default is taken into account, but FrankenPHP uses the default system path (hardcoded in the PHP source code) which is `/lib/php.ini`. Most distributions change this default path to `/etc`, hence the confusion.

If you use the static binary (default with Octane), you can use `/lib/php.ini` or the root directory of the project.  If you use the Docker image, you should use `$PHP_INI_DIR`. If you compile PHP by yourself, you can specify the path you want by passing ` --with-config-file-path` option to `./configure`.

This is documented here: https://frankenphp.dev/docs/config/

olliescase: Thanks @dunglas, I've rewritten to hopefully be a little more precise on this matter. Are you able to take a look and check if you think it covers it?

octane.md
```diff
@@ -137,6 +137,27 @@ services:
...
+To adjust PHP configuration, create a `php.ini` file in one of these locations and define your settings. For example, to increase the memory limit and file upload size, you can add the following to your `./php.ini` file:
+
+```ini
+# php.ini
+memory_limit = 512M
+upload_max_filesize = 100M
+post_max_size = 100M
+```
+
+If you're using the Docker image, you can set the `$PHP_INI_DIR` environment variable. FrankenPHP will then scan and load configuration files from `$PHP_INI_DIR/*.ini`.
```
dunglas, line 154:
```suggestion
If you're using the Docker image, you must store your `php.ini` in the directory specified in the `$PHP_INI_DIR` environment variable. FrankenPHP will then scan and load configuration files from `$PHP_INI_DIR/*.ini`.
```

taylorotwell:
I feel like this maybe doesn't belong in our documentation?

olliescase:
Sorry, been a pretty intense week. I disagree that this doesn't belong in the Laravel docs, because Laravel is explicitly supporting FrankenPHP. At the very least I think this should come down to a mention in the documentation. This is a vital piece of knowledge that everyone using Laravel with FrankenPHP should know (and beyond - be warned of).

I understand that its fairly logical to assume that the installed PHP's configuration won't apply to the standalone binary of FrankenPHP, but this is a facet that people making the switch won't necessarily realise the implications of.

Switching from Swoole to FrankenPHP is made out in the Laravel documentation to be as simple as doing the `install` command and changing the environment variable. This extra step that isn't needed for swoole (as it is just an extension installed on the system php).

This can cause some serious problems that are often not visible in dev/testing and only come to fruition in production when you deal with a more intensive load

dunglas:
For the record, we're working with PHP Cli Team to improve the default experience: https://github.com/crazywhalecc/static-php-cli/pull/553

Once this PR is merged, maybe a simple note hinting that another `php.ini` than the system one is used would be enough?

olliescase:
Yeah, I think that would be good - this is perhaps a little verbose for the Larvel documentation - I can agree on that. But a small note.

Maybe an additional flag to the install command to be `--ini` that in combination with `--server=frankenphp` also creates an empty `php.ini` in the root of the laravel project? 🙂 

A template along the lines of:
```ini
# Place your project-specific FrankenPHP ini variables here. Server-level configuration can be done in /usr/local/etc/php/php.ini
# For more information, see https://frankenphp.dev/docs/config/
```
(haven't invested much time towards an elegant wording)

I think this could make the process more elegant and also serve as an extra tidbit of in-code documentation to point people in the right direction.

Cosnavel:
@taylorotwell think it can be closed

PR #9933, opened 2024-10-08
Document solution for multiple rate limits on same segment

sgilberg:
I discovered that if I had both a per-minute and per-day rate limit on the same segment, only the first limit worked. It looks like I'm not the only one who has encountered this, and I noticed [this note](https://github.com/laravel/framework/issues/46290#issuecomment-1449910366) on the [suggested workaround here](https://github.com/laravel/framework/issues/46290#issuecomment-1447427560) inviting a PR to the docs, so here you go!

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -829,12 +829,12 @@ If needed, you may return an array of rate limits for a given rate limiter confi
         ];
     });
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Thanks!

PR #9938, opened 2024-10-10
Add output example for once helper function in documentation

aceraven777:
This PR adds an output example for the `once` helper function in the Laravel documentation. The example demonstrates how the `once` function caches results within the same instance and ensures that the method is executed only once per instance.


taylorotwell:
I think it's ok.

PR #9939, opened 2024-10-10
The code examples had 3 seconds not 30

alnutile:
I think this was a mistake?


taylorotwell:
It just says the default is 30, then the example shows how to set a different timeout.

alnutile:
ah I see makes sense

PR #9942, opened 2024-10-11
[11.x] explain using `@push` with layout components

browner12:
I recently converted a site from "template inheritance" layouts to "component" layouts. previously, I could have `@push`s declared anywhere within my view, and they would work fine. now I've noticed they **will not** work when declared after the closing tag of my layout component.

I'm not sure if this is intended behavior, a technical limitation, OR something that could/should be fixed.

If it's one of the first 2, I thought we should document it.

### DIFF
--- a/blade.md
+++ b/blade.md
@@ -1721,6 +1721,26 @@ If you would like to prepend content onto the beginning of a stack, you should u
 @endprepend
 ```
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
This feels kinda edge case. Let me think about how to word it. 🤔

PR #9946, opened 2024-10-11
Update installation.md

ginger-tek:
The PHP Installation commands for Windows can be simplified, as newer versions of PowerShell/Windows now provide more concise and readable commands and aliases, as well as set the TLS version to 1.2 at minimum by default.

An alternative to the proposed change could be `winget install php`, thanks to the Windows Package Manager. However, if the user has a version of Windows that does not include winget, that might hinder or otherwise exclude users without it. Some users may also just not want PHP managed by winget, which is understandable

### DIFF
--- a/installation.md
+++ b/installation.md
@@ -70,7 +70,7 @@ If you don't have PHP and Composer installed on your local machine, the followin
 ```
 
 ```shell tab=Windows PowerShell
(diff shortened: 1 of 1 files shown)


ginger-tek:
Strange. I do know that the built-in PowerShell on base Windows installations, depending on the OS version, is older (version 5). But most anyone developing on a Windows machine will have installed PowerShell 7+ and Windows Terminal anyway, which might be the issue.

ginger-tek:
Valid. I hope Windows fixes or upgrades the built-in powershell sometime soon, as the offering versions are often the root cause whenever it comes to running powershell scripts across people's machines.

PR #9950, opened 2024-10-12
[12.x] Add dedicated page for "Macros & Mixins"

shaedrich:
Fix laravel/framework#53123

### DIFF
--- a/macroable.md
+++ b/macroable.md
@@ -90,7 +90,7 @@ $translated = $collection->toLocale('es'); // [tl! focus]
 <a name="response-macros"></a>
 ## Response Macros
 
-If you would like to define a custom response that you can re-use in a variety of your routes and controllers, you may use the `macro` method on the `Response` facader:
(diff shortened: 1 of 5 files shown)


macroable.md
```diff
@@ -0,0 +1,211 @@
...
+// …
+
+$collection = collect(['first', 'second']); // [tl! focus]
+$translated = $collection->toLocale('es'); // [tl! focus]
+```
+
+<a name="response-macros"></a>
+## Response Macros
+
+If you would like to define a custom response that you can re-use in a variety of your routes and controllers, you may use the `macro` method on the `Response` facader:
```
kapersoft, line 93:
```suggestion
If you would like to define a custom response that you can re-use in a variety of your routes and controllers, you may use the `macro` method on the `Response` facade:
```

shaedrich: Thanks! 👍🏻

PR #9951, opened 2024-10-12
readable return type for command

Nir-An:
exit const for command

Use constants to improve code readability and prevent mistakes with PHP return types and command exit codes. This approach ensures consistency and clarity throughout the code.

 https://tldp.org/LDP/abs/html/exitcodes.html


taylorotwell:
I don't think this helps readability that much.

PR #9959, opened 2024-10-16
Updated namespace events namespace to reflect documentation

RRosalia:
In this specific section we are talking about Event Discovery with the usage of Listeners.


taylorotwell:
We are importing an event though?

PR #9962, opened 2024-10-18
[11.x] Added section for compatibility for Docker Desktop for Linux.

RyanPaiva56:
There's been a handful of issues posted about file access problems with docker on Linux. After reading through all of the issues on Laravel Sail and then other related issues posted to Docker Desktop for Linux, the actual issue is that Docker Desktop runs in a VM and doesn't have write permissions unless you explicitly set them and add the host user to that group. 

The current warning on the Laravel docs doesn't resolve the issue, basically you're just disabling Docker Desktop (haha) by switching contexts, and so I removed the previous stop-gap and replaced it. Docker Desktop will only work within its desktop-linux context because it's hard coded to do so within a VM. Docker Desktop also automatically changes the context from default to desktop-linux on start and stop, so changing the context is going to be an annoying thing you have to do every time forever. The fix added via this PR makes a permanent fix until Docker adds the ability for Docker Desktop to handle additional contexts, which they don't appear to have any plans for and cite security and stability as the reasons they don't want to touch it.

Also, other solutions basically have you run sail as root or disable rootless docker, neither of which are good. The solution in this PR reasonably limits the amount of files given write access without invoking root.

I also moved it to the top under intro just below where the docs state other compatibilites. It doesn't make sense for it to be farther down the page because you can't start sail without errors unless you do this first.

This PR also removes some whitespace that appears unnecessary.

Credit to @Ernestopheles for the original solution: https://github.com/cytopia/devilbox/issues/963#issuecomment-1438939653

This PR likely resolves these issues:

https://github.com/laravel/sail/issues/646
https://github.com/laravel/sail/issues/548
https://github.com/laravel/sail/issues/459
https://github.com/laravel/sail/issues/457
https://github.com/laravel/sail/issues/435
https://github.com/laravel/sail/issues/372
https://github.com/laravel/sail/issues/149
https://github.com/laravel/sail/issues/81


rojtjo:
> This PR also removes some whitespace that appears unnecessary.

These actually add line breaks in Markdown.
See: https://www.markdownguide.org/basic-syntax/#line-breaks

RyanPaiva56:
Thanks, it didn't show up on my preview but I'll take care of it and resubmit.

MansourWolou:
Hi @taylorotwell i'm having the same issues that are supposed to be resolved by this PR. But it looks like the commits are unmerged. @RyanPaiva56  are your commits merged in the current laravel 11.x ?

RyanPaiva56:
I closed it because there's no way to make it compatible. Don't use docker desktop.

PR #9965, opened 2024-10-19
[11.x] Move contract docs to "Architecture Concepts" section in navigation

shaedrich:
from the Digging Deeper section to right after the Facades docs, since these two concepts are closely related (hence the section "Contracts vs. Facades" on the Contracts docs page

Fixes laravel/framework#53123


taylorotwell:
I personally don't think contracts should be an emphasized part of Laravel anymore at all.

PR #9966, opened 2024-10-19
[11.x] Add `enum_value` helper

milwad-dev:


crynobone:
The function is marked as `@internal` and shouldn't be documented.

milwad-dev:
But I think it is good to use it!

crynobone:
https://stackoverflow.com/a/49846424

`@internal` is meant to be used only within the Framework and may have breaking changes in minor or patch release. This is not suited to be documented

PR #9967, opened 2024-10-19
[11.x] Reference docs for attributes

shaedrich:
Example: https://symfony.com/doc/current/reference/attributes.html

Fixes laravel/framework#53126

Similar to: https://laravel.com/docs/11.x/contracts#contract-reference

Not sure if the section from the [service container docs](https://laravel.com/docs/11.x/container#contextual-attributes) should be moved there, too, but if wanted, that could alternatively done in a follow-up PR.

### DIFF
--- a/attributes.md
+++ b/attributes.md
@@ -1,0 +1,58 @@
+# Attributes
+
+- [Introduction](#introduction)
+- [Attributes reference](#attributes-reference)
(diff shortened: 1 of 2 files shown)


taylorotwell:
Don't really have an appetite for this atm.

PR #9983, opened 2024-10-23
Add note to Eloquent.md to clarify UUID/ULID usage with Sanctum

ThomCM:
Updated the documentation to provide clarification on using UUIDs or ULIDs with Laravel Sanctum, specifically recommending the use of ulidMorphs() or uuidMorphs() methods to configure the tokenable_id column correctly.


taylorotwell:
This may be better placed in Sanctum or just considered obvious knowledge since column types would need to match.

PR #9984, opened 2024-10-23
[11.x] new allowed filename for anonymous index components

browner12:
we can now use either `index.blade.php` or `{directory}.blade.php` for our anonymous index components.

https://github.com/laravel/framework/pull/52669


browner12:
I chose to omit commenting on the fact the `index.blade.php` file takes precedence over the directory named file. I went back and forth on it, but it seems like an edge case, and didn't even want to put it in people's heads that they could have both.

open to adding it.

browner12:
i'm down with discouraging the use of `index.blade.php`, and now we just have consistency across Class based and anonymous components.

PR #9991, opened 2024-10-26
[11.x] Update the section for custom notification channels.

kevinb1989:
I try to implement a custom channel with Twilio, following the current instructions and it didn't work.

I follow the instructions for [caches](https://laravel.com/docs/11.x/cache#adding-custom-cache-drivers), and it works!

I read the source code and believe that this works similar to custom cache, email or queue drivers.


taylorotwell:
You would need to provide more info on what exactly didn't work. Error messages, reproduction steps, etc.

PR #9999, opened 2024-10-28
[11.x] remove "unique" SQL injection warning

browner12:
I don't believe this warning is actually valid anymore. Best I can tell, using the default `DatabasePresenceVerifier`, we are executing an aggregate "count" query, which use Prepared Statements with variable binding.

the note was added 5 years ago [here](https://github.com/laravel/docs/commit/8c63ad24fdb6750e0372e4ee9170ac2b119c377f)

the query is built [here](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Validation/DatabasePresenceVerifier.php#L46)

for the example of updating a user and ensuring the email remains unique, the generated query is

```sql
select count(*) as aggregate from `users` where `email` = ? and `id` <> ? and `users`.`deleted_at` is null
```

we could also go the route of not removing this completely, but softening the language to something like "*may* be vulnerable..." in case we're worried about people using alternate "Presence Verifiers".

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -1860,9 +1860,6 @@ To instruct the validator to ignore the user's ID, we'll use the `Rule` class to
         ],
     ]);
 
-> [!WARNING]  
-> You should never pass any user controlled request input into the `ignore` method. Instead, you should only pass a system generated unique ID such as an auto-incrementing ID or UUID from an Eloquent model instance. Otherwise, your application will be vulnerable to an SQL injection attack.
-
 Instead of passing the model key's value to the `ignore` method, you may also pass the entire model instance. Laravel will automatically extract the key from the model:
 
     Rule::unique('users')->ignore($user)


browner12:
would love secondary confirmation that we are, in fact, using Prepared Statements for the query.

taylorotwell:
I think the string concatenation here where the ignore value is placed in the `unique` rule string may be vulnerable to manipulation if it's uncontrolled user input.

CleanShot 2024-10-29 at 08 55 34@2x

browner12:
@taylorotwell we are still fine here. let's say someone tries to add in "DROP TABLE `users`" or something. the generated validation string is:

```
"unique:users,NULL,"DROP table `users`",id" 
```

the only place this gets used is in the [database verifier](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Validation/DatabasePresenceVerifier.php#L51) using a `where()` clause, which is a prepared statement.

you can test this with a simple example

```php
$validation = Validator::make(
    ['author_id' => 'value'],
    ['author_id' => Rule::unique('posts')->ignore('DROP TABLE `users`')],
);

$validation->validated();
```

if you run this on a route, and then view the executed queries you'll get:

Screenshot 2024-10-29 at 12 45 53 PM

PR #10002, opened 2024-10-29
Docs for \@bool directive

david-valdivia:
Documentation for \@bool directive on blade.


taylorotwell:
I need to think about the best place for this.

PR #10004, opened 2024-10-29
Logging: JSONL example for CustomizeFormatter

flexponsive:
Changes the existing `CustomizeFormatter` example to format logs as individual JSON objects per line (JSONL) using Monolog's `JsonFormatter`, instead of using a custom string format as before.

Especially with the inclusion of logging context since Laravel 11, JSONL is a really powerful log format to enable ad hoc queries and better log analysis. 

For example, get the last 50 logs with a specific tenant id in the context:

```bash
tac laravel.jsonl | jq 'select(.context.tenant_id == 12)' | head -n 50
```


taylorotwell:

PR #10015, opened 2024-11-05
Add beforeResolving and afterResolving methods to container docs

ahmadreza1383:
I've added the beforeResolving and afterResolving methods alongside the existing resolving method. This keeps the documentation concise while making developers aware of these additional methods.


taylorotwell:
Will try to work these in a bit more cohesively. 👍

PR #10017, opened 2024-11-07
Fix artisan command in queues.md

maheenncodes:
the command for generating queues table is wrong


taylorotwell:
It's not. This is correct on Laravel 11.

PR #10019, opened 2024-11-07
[11.x] Add PHP 8.4

Jubeki:
Laravel 11 and 12 will support PHP 8.4.

Sail already has a PHP 8.4 runtime, but it should not be the default until PHP 8.4 is released.

Laravel Homestead does not support PHP 8.4 yet. (At least I saw no commit indicating that)

Depends on https://github.com/laravel/framework/pull/52633


valpuia604:
L-10 should also support 8.4? https://github.com/laravel/framework/pull/53612/files

crynobone:
@valpuia604 Not exactly, I just proposing that it can run on PHP 8.4 if Taylor approved this. Officially Laravel 10 is only on security releases and does not accept new features/bugfixes.

PR #10022, opened 2024-11-09
[11.x] Add on demand `Mail::build()` function in docs.

pushpak1300:
This MR add new `Mail::build()` feature in docs.

https://github.com/laravel/framework/pull/53411


taylorotwell:
I think this is a bit more of an edge case feature that the vast majority of people will never need. Need to think about where it makes sense to document it.

PR #10028, opened 2024-11-15
[11.x] Document toMailUsing usage for password reset

fsylum:
Added the undocumented usage of `toMailUsing` with `PasswordReset` notification class.


taylorotwell:
I prefer to just document one simple way of doing this.

PR #10031, opened 2024-11-19
[11.x] add `collapseWithKeys` method to Collection

shane-zeng:
`collapseWithKeys` Collection methods that have been added in https://github.com/laravel/framework/pull/52347


taylorotwell:
I think it would be better to make the example a bit more real world.

PR #10032, opened 2024-11-19
[11.x] document new cache driver

browner12:
documentation for [this PR](https://github.com/laravel/framework/pull/53428).

should not be merged until after next release so my [bugfix](https://github.com/laravel/framework/pull/53561) is in there and it actually works.

I struggled with the wording on this. maybe it'd be better to just show an example `config/auth.php` entry?


taylorotwell:
Before we document this could we implement that automatic hashing of the keys so we don't need to worry about prefixes @browner12 ?

browner12:
@taylorotwell was that suggested somewhere? I'm not seeing it.

Also, I'm a little confused. Are you suggesting to hash the "prefix.email" combo? How would we retrieve it from the cache then?

PR #10035, opened 2024-11-20
Update sail.md

Rian190:
Add a docker command to install the composer dependencies and install sail in an existing application without the need of Composer and Php on the host.


Rian190:
Its not working yet as expected. I'm closing this until I have a good version

PR #10036, opened 2024-11-20
Install sail using docker

Rian190:
I ran into the situation where I had to intall sail in some projects which all had their own php version. Instead of juggling the php version of my OS, I created a docker command to help me out.

This command can be useful for others in the community as well


taylorotwell:
Going to hold off on this for now.

PR #10039, opened 2024-11-20
Update notifications.md

GoranBosevski:
The command that creates the notifications table is wrong, instead php artisan make:notifications-table, should be php artisan notifications:table


taylorotwell:
You're not on latest Laravel 11.

GoranBosevski:
Correct, thank you.

PR #10042, opened 2024-11-27
Fix 404 url for Linux install

brunogoossens:
The documentation contains a link to a 404 page.


taylorotwell:
We just need to fix the 404, not remove it from the docs.

PR #10049, opened 2024-12-02
Update laravel 10.x upgrade guides

imanghafoori1:
We had such a thing in our app but it was not clear that it stops working in Laravel 10 from the upgrade guides.


taylorotwell:
Will take a look - we would need to confirm a full list of methods.

PR #10051, opened 2024-12-04
Typo corrected in http-client.md

vokamut:


taylorotwell:
The defaut **is** 30 seconds.

PR #10056, opened 2024-12-08
Include encrypted:json as documented cast

ok200paul:
Small PR to include encrypted:json as a cast option in the docs.

We're building an API that takes in json in the POST but encrypts/decrypts it in the model. There was a little confusion on my side whether we needed to instruct client applications to send arrays (encrypted:array) rather than json for this to work. Looking through the framework code showed encrypted:json as an option, so this PR just exposes this to folks like me.


taylorotwell:
You should use `encrypted:array`. I don't think we document the JSON cast because it's just the same as array.

PR #10059, opened 2024-12-09
Update cache.md

jonkerw85:
The documentation incorrectly states that `stores.dynamodb.attributes.key` refers to the partition key for the DynamoDB table, when it actually holds the AWS access key (`AWS_ACCESS_KEY_ID`). This change clarifies that the table should simply have a string partition key named `key`, without any reference to the `stores.dynamodb.attributes.key` configuration item.


taylorotwell:
Docs are correct. Confirmed in source code.

jonkerw85:
Hi Taylor,

Thanks for taking a look! I dug deeper and noticed that stores.dynamodb.attributes.key is being used in two different
contexts:

- For AWS authentication as the AWS_ACCESS_KEY_ID (via newDynamodbClient), which aligns with its usage across other services like S3, SQS, and SES.
- As the attribute name for the DynamoDB table's partition key, as suggested in the documentation.

This dual purpose can be a bit confusing, as it conflates authentication credentials with table schema definition. It might be worth clarifying this in the docs or considering separate configuration keys to reduce ambiguity.

Let me know what you think!

PR #10062, opened 2024-12-11
[11.x] Storage:link source and target directory statement updated

ChetanTechnource:
Storage:link source and target directory statement updated as **public/storage is the target** and **storage/app/public is the source**.

Example : 
```
ln -s /var/www/html/laravel11-x/storage/app/public /var/www/html/laravel11-x/public/storage
```


ChetanTechnource:
@taylorotwell Please review if this is helpful.

PR #10065, opened 2024-12-11
Add a section on disabling SSL certificate verification

adhuham:
Add a sub-section on using the `withoutVerifying` method to disable SSL certificate verification.


taylorotwell:
This probably wouldn't receive its own H3.

PR #10080, opened 2024-12-17
[11.x] Fix Octane Docs Example Related with FrankenPHP + HTTPS

adrianobnu:
Fixed the SUPERVISOR_PHP_COMMAND in the example, because it's missing the flag of `server=frankenphp`.


taylorotwell:
This isn't really necessary if your `config/octane.php` file uses a server of `frankenphp`?

PR #10081, opened 2024-12-18
[12.x] Use tap() in middleware example

shaedrich:


shaedrich:
First, you invent it, then you don't want to use it … ¯\\_ (ツ)_/¯

joshmanders:
@shaedrich there's no reason to use it here.

shaedrich:
@joshmanders Isn't it, though? I'd say, it is the ideal use case.

PR #10085, opened 2024-12-22
Add note about Sanctum/Axios URI Decoding

aMytho:
When a sanctum/SPA XSRF token is sent to the client, it is URL encoded. It needs to be decoded before being sent back to the server. The examples with axios shows the settings that will automatically add the token to the header. However, they don't mention that axios also decodes the token.

The current example makes it seem like you can pass in the token without decoding the value. This is false and it caused me 2 days of troubleshooting.

I added a fetch example with a note about the decoding. This should help to reduce confusion when using a library other than axios. If you don't think that a fetch example is appropriate, it can be removed. However, I think we need to include the info about the token encoding.

(This is similar to https://github.com/laravel/docs/pull/7579)

### DIFF
--- a/sanctum.md
+++ b/sanctum.md
@@ -321,7 +321,7 @@ axios.get('/sanctum/csrf-cookie').then(response => {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Will think about this one but still not quite happy with how it sits in the documentation. I think the prior attempt you linked is a bit better in approach but just needed a little formatting. Feel free to take a stab at that if you want!

PR #10091, opened 2024-12-28
update api:install command

amir-astarabadi:
introduces the --passport option to the api:install command,


taylorotwell:
I would prefer to not have this.

PR #10099, opened 2025-01-05
Added Event Registration Information About Queue::exceptionOccurred 

WDC:
I added information about the `Queue::exceptionOccurred` method on the facade that is slightly more-verbose than the Queue documentation page's existing `Queue::failing` example. I believe both are equally helpful, but in different situations.


taylorotwell:
No plans to add this right now.

PR #10101, opened 2025-01-06
[11.x] use the word "Blog" instead of "Post"

browner12:
the word "post" can have multiple contexts in the world of web development, so in order to impove searchability, let's avoid using "Post" as a Model/Noun.

this will allow people searching for things related to the HTTP verb "POST" and route related things in these docs a little easier, without losing any clarity in the rest of the page.

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -53,12 +53,12 @@ To learn about Laravel's powerful validation features, let's look at a complete 
 
 First, let's assume we have the following routes defined in our `routes/web.php` file:
 
-    use App\Http\Controllers\BlogController;
+    use App\Http\Controllers\ArticleController;
 
-    Route::get('/blog/create', [BlogController::class, 'create']);
-    Route::post('/blog', [BlogController::class, 'store']);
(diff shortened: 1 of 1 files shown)


LucaRed:
I like the idea but I'd use another word, such as `Article`.

Creating, editing, deleting, etc. articles is more intuitive than "blogs".

joshmanders:
At first I was like "this is a bit too pedantic for an example in the docs" for the same reason @LucaRed said, but your second point of removing the noise from docs searching for actual framework things makes me agree.

I'd go with `Articles` over `Blogs` too for intuitiveness.

taylorotwell:
Yeah I would call it Article or something so we're operating on a resource.

browner12:
"Blog" has been renamed to "Article" per suggestion.

taylorotwell:
I dunno, on second thought I may just leave this as is. Not sure it has been a big complaint or headache. 😬

browner12:
bummer. I get the annoyance is niche, but the change doesn't hurt anything, and fixes the complaint.

PR #10107, opened 2025-01-10
Add information about the mysql config `use_db_after_connecting`

TheLevti:
Added information about the MySQL connector specific configuration `use_db_after_connecting`. Setting this to true will prevent an explicit `USE database_name` statement after the connection has been established.

Added in https://github.com/laravel/framework/pull/54132


taylorotwell:
Personally not sure this needs its own section in the docs atm.

PR #10108, opened 2025-01-11
`assertMissing()` should proceed `assertVisible()` as its counterpart

u01jmg3:
🪄


u01jmg3:
When you're reaching for the opposite of `assertPresent()`, it's right there and comes straight after in the form of `assertNotPresent()`.

When you're reaching for the opposite of `assertVisible()`, you have to scroll past `assertPresent()` and then past `assertNotPresent()` before you eventually hit `assertMissing()`. Just think it flows better with my change.

PR #10110, opened 2025-01-13
Add queue batching configuration details to upgrade guide

MJAISE:
This pull request updates the Laravel 11 upgrade guide to include crucial information regarding the new batching key introduced in the queue.php configuration. In Laravel 11, this configuration is essential for proper management of job batches when using the database queue. The update ensures users are aware of the necessary change to prevent issues when processing batched jobs.

The batching key should be added to the queue.php configuration file with the following settings:

```
'batching' => [
    'database' => env('DB_CONNECTION', 'mysql'),
    'table' => 'job_batches',
],
```
This ensures that job batches are correctly stored and managed in the specified database. If you are using a custom database connection for your queues, you should change the DB_CONNECTION value in the .env file accordingly to reflect that connection.

Including this configuration will prevent potential issues when processing batched jobs, ensuring smooth job processing.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -25,8 +25,10 @@
 - [Password Rehashing](#password-rehashing)
 - [Per-Second Rate Limiting](#per-second-rate-limiting)
 - [Spatie Once Package](#spatie-once-package)
+- [Queue Batching Configuration](#queue-batching-configuration)
 
 </div>
+
 
 <a name="low-impact-changes"></a>
 ## Low Impact Changes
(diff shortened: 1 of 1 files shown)


taylorotwell:
This shouldn't be necessary. The configuration should cascade to the framework's own configuration which has this in place.

MJAISE:
When using multiple database connections, the environment variable may not always be `DB_CONNECTION`. Users can define a custom connection name, such as `MAIN_DB_CONNECTION` for the primary database and `QUEUE_DB_CONNECTION` for queue processing.

 Additionally, MySQL may not be the default database—in my case, I use PostgreSQL. 

In such scenarios, relying on the default framework value can lead to failures if it doesn't match the intended database setup.

PR #10113, opened 2025-01-15
Add pessimistic locking extended example

DarkGhostHunter:
I do not expect everyone to be "database masters" to wrap their head around locking, so the added example should be enough to understand when and how it should be used: avoid data races by keeping the data immutable until the transaction completes.


DarkGhostHunter:
Yay!

I'll be on the lockout if I find more topics to expand on.

joshmanders:
> I'll be on the **lockout** if I find more topics to expand on.

Not sure if that was intended or not, but I dig it.

PR #10115, opened 2025-01-15
InteractsWithQueue trait is needed for Batchable jobs

vadimonus:
While testing job batches found, that `then` and `finally` methods relies on InteractsWithQueue trait. 
See https://github.com/laravel/framework/blob/6eaa4b3a4d3961d90ba90def9f141367a82ac995/src/Illuminate/Queue/CallQueuedHandler.php#L183 (CallQueuedHandler::ensureSuccessfulBatchJobIsRecorded)

But documentation misses that requirements, making it difficult to find out, why `then` or `finally` not working.

creating-jobs link also says nothing about InteractsWithQueue trait (only about ShouldQueue interface)

This PR fixes that. 
Fixes also appliable for master, 10.x, 9.x, 8.x

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -1298,7 +1298,7 @@ php artisan migrate
 <a name="defining-batchable-jobs"></a>
 ### Defining Batchable Jobs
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Foundation\Queue\Queueable includes this.

PR #10117, opened 2025-01-15
Note about client sharding works only with predis

vadimonus:
Phpredis extention does not provide client-side sharding, only Predis can do so. Currently this is unclear from docs.

This PR makes this more clear.

Can be applied for all supported versions


taylorotwell:
Should be sent to 11.x.

PR #10118, opened 2025-01-15
Calrifying risks of using truncate

vadimonus:
This PR is some way opposite to https://github.com/laravel/docs/pull/10116

The current documentation does not accurately describe the risks of using truncate (appeared in https://github.com/laravel/docs/pull/6560).
Even it documentated for 4 years, users do not understand risks. See https://github.com/laravel/framework/issues/35157#issuecomment-1625141439  and later in thread.

This PR clarifies potential problems, that comes with truncate 
 * recursive data deletion for postgresql (not only foreign key related records, but all records)
 * data inconsistency for mysql

PR #10119, opened 2025-01-15
[11.x] Delete all models example

vadimonus:
In discussion about incorrect usage of truncate (https://github.com/laravel/docs/pull/10118 , https://github.com/laravel/docs/pull/10116), supprisely found, that Eloquent page does not contain example of deletion all rows in table other than with truncate.

To delete it with query we should write `Model::query()->delete()`, not just `Model::delete()`, and this is not obvious, because this page does not contain any other example starting with `Model::query()`.

As the result, many users use `truncate` for deletion of all records, leading to issues like https://github.com/laravel/framework/issues/35157 and https://github.com/laravel/framework/issues/29506 . Now i understand, is because truncate is documented better and higher in documentation, than delete.

This PR adds example of deletion of all modles in table.

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -954,6 +954,10 @@ Of course, you may build an Eloquent query to delete all models matching your qu
 
     $deleted = Flight::where('active', 0)->delete();
 
+To delete all models you can build query without conditions:
+
+    $deleted = Flight::query()->delete();
+
 > [!WARNING]  
 > When executing a mass delete statement via Eloquent, the `deleting` and `deleted` model events will not be dispatched for the deleted models. This is because the models are never actually retrieved when executing the delete statement.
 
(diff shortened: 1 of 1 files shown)


eloquent.md
```diff
@@ -940,10 +940,6 @@ In the example above, we are retrieving the model from the database before calli
 
     Flight::destroy(collect([1, 2, 3]));
 
-If you are utilizing [soft deleting models](#soft-deleting), you may permanently delete models via the `forceDestroy` method:
-
-    Flight::forceDestroy(1);
-
```
taylorotwell, line 946: Why did you remove this?

vadimonus: Just mistake when rebasing from master. Fixed

PR #10127, opened 2025-01-22
Warn about Auto Discovery of Event Listener

marijoo:
I ran into a problem when upgrading an application from Laravel 10 which turned out to be caused by event listeners that were intentionally unregistered before but where the listener class was still located in the `app/Listeners` directory.

The proposed change to the upgrade guide will help with this, I think. Plus I added some missing references to the TOC.


taylorotwell:
This isn't relevant if you maintain the Laravel 10 application structure, which could be confusing for people.

PR #10131, opened 2025-01-23
[11.x] Add CORS configuration to Vite plugin docs

timacdonald:
Adds docs to help guide users with custom origin to configure CORS with Vite.

This is under "Advanced" as majority of Laravel applications should never need this unless they are doing something very custom. All the first-party tooling is supported out of the box.


onlime:
Super nice crystal clear docs! 👏 Thanks for making it work out of the box for most and easy to understand for everyone else. Thanks for the related https://github.com/laravel/vite-plugin/pull/318

PR #10134, opened 2025-01-24
[11.x] migration docs:  adding clarification for json/jsonb datatypes…

hexathos:
Adding clarification for json/jsonb datatypes on sqlite.


shaedrich:
A feel like this docs page should have a matrix table showing
* which engines support/don't support a feature
* different namings/methods to achieve the same feature per engine
* different behavior for the same feature per engine

PR #10137, opened 2025-01-25
[12.x] Add `previousPath` usage to urls.md

miclaus:
```
     // Get the full URL for the previous request...
     echo url()->previous();

+    // Get the full URL for the previous request without query parameters...
+    echo url()->previousPath();
```


taylorotwell:
I don't think this returns the full URL?

PR #10138, opened 2025-01-26
Add explanation for index-settings for Algolia in Scout

joostdebruijn:
Explanation of the functionality added here https://github.com/laravel/scout/pull/889, which enables the sync of index settings to Algolia in Laravel Scout.


taylorotwell:
I think we would want explanation of when or why you would do this vs. configuring things in the UI.

PR #10140, opened 2025-01-27
Upgrade Guide Title Fix 

yagnikmt:
There is wrong title in upgrade guide

Which said Upgrade to 10 to 11 Which logically said downgrade not upgrade 

So fix it

image


crynobone:
No, the current text is correct. You are uograding **TO** 11.0 **FROM** 10.x

yagnikmt:
Okay got it thanks

PR #10141, opened 2025-01-27
feat(routing): add --except-path docs

maloun96:
Add missing documentation for the --except-path option in route:list command

it was added here but without docs.
https://github.com/laravel/framework/pull/36619


taylorotwell:
The full list of options can be reviewed via `php artisan help route:list`

PR #10143, opened 2025-01-29
[11.x] The WARNING about the lang directory is repeated

AhmedAlaa4611:
The WARNING about the lang directory and the `lang:publish` Artisan command is repeated three times in the documentation. Since it conveys the same information each time, we could consolidate it into one instance to avoid redundancy and make the documentation cleaner.


taylorotwell:
I don't mind repeating it.

PR #10147, opened 2025-01-30
Show database timezone key in example

NikitinUser:
\Illuminate\Database\Connectors\PostgresConnector::class
```
protected function configureTimezone($connection, array $config)
{
    if (isset($config['timezone'])) {
        $timezone = $config['timezone'];

        $connection->prepare("set time zone '{$timezone}'")->execute();
    }
}
```

If the timezone key does not exist in config/database.php for your database (PostgreSQL in my case), then $config['timezone'] will be null. As a result, the database server will use its default timezone instead of the application's timezone (i.e., the value set in config('app.timezone')). 
The existence of the config('database.dbname.timezone') key is not documented anywhere, and its impact went unnoticed until the upgrade to version 11. However, it has a significant effect: if it is not specified, timestamps in the database may become incorrect.

### DIFF
--- a/database.md
+++ b/database.md
@@ -113,6 +113,8 @@ To see how read / write connections should be configured, let's look at this exa
         'options' => extension_loaded('pdo_mysql') ? array_filter([
             PDO::MYSQL_ATTR_SSL_CA => env('MYSQL_ATTR_SSL_CA'),
         ]) : [],
+
+        'timezone' => config('app.timezone'),
     ],
 
 Note that three keys have been added to the configuration array: `read`, `write` and `sticky`. The `read` and `write` keys have array values containing a single key: `host`. The rest of the database options for the `read` and `write` connections will be merged from the main `mysql` configuration array.


shaedrich:
Could you please make the title a little more descriptive?

browner12:
I don't think we should encourage this usage. 

I'd say let's make people dig a little if they truly want to adjust the timezone of the database connection.

shaedrich:
@browner12 You must really hate people the way you constantly want to put spokes in people's wheels

browner12:
you got me, I'm only here cause I hate people.

not that I legitimately think setting a timezone on a database connection is a bad practice that we shouldn't encourage.

taylorotwell:
I don't think adding that here in the docs really solves anything. 🤔

NikitinUser:
@browner12  As you can see in the code I provided, the time zone for the current database session is set anyway. However, without this parameter, you may get a different time zone each time, requiring constant attention to server configuration.

NikitinUser:
@taylorotwell  OK, what would you suggest?

PR #10149, opened 2025-02-03
Fix Password Facade Constant

medabkari:


shaedrich:
You can use both:
https://github.com/laravel/framework/blob/53091c205f110fb57ffdc1584e10bdaaf10d0d88/src/Illuminate/Support/Facades/Password.php#L29
https://github.com/laravel/framework/blob/11.x/src/Illuminate/Support/Facades/Password.php#L59
No need to change that

shaedrich:
The constants were added in 5744f80 on on Dec 19, 2024, two days after the release of [`11.36.1`](https://github.com/laravel/framework/releases/tag/v11.36.1), so in the current version of v11.x, they are available. Regardless of whether it is a good policy to include stuff in the docs which is extremely fresh, this is nothing, that needs fixing. That you can't use it (and I assume, you can't easily update for some reason), is unfortunate.

medabkari:
@shaedrich Well, in that case, I'll go ahead and close the PR.

Thanks for the info!

shaedrich:
I'm not a maintainer, so I might be wrong

medabkari:
> I'm not a maintainer, so I might be wrong

I took your comment for granted. Should I reopen the PR?

PR #10152, opened 2025-02-05
Middleware priority append/prepending

adamjsturge:
I recently was trying to append a middleware to the list and found the documentation confusing. After reading the vendor files and testing I figured it out and wanted to add to the docs


taylorotwell:
Will think about this one!

PR #10153, opened 2025-02-06
Add warning about user running and terminating Horizon must match

tobz-nz:


taylorotwell:
I'm not sure we need to document this at the moment.

tobz-nz:
After spending over an hour figuring out why Horizon wasn't actually restarting - I really think this is needed to be documented.
It's not super obvious that a) the daemon has been restarted. b) that the problem could be the user.

PR #10154, opened 2025-02-07
[11.x] Add documentation for Context::missing()

vbergerondev:
This PR introduces some new documentation for `Context::missing` which has been added to the framework in this PR: https://github.com/laravel/framework/pull/54499

### DIFF
--- a/context.md
+++ b/context.md
@@ -269,6 +269,14 @@ Context::has('key');
 // true
 ```
 
+To determine if a key is not present in the context, you may use the `missing` method. The missing method returns true if the key has not been added to the context:
+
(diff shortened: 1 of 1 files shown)


shaedrich:
Not sure, if one needs an example for the same thing as above except negated 🤔

vbergerondev:
> Not sure, if one needs an example for the same thing as above except negated 🤔

As the person who made the PR that adds the `missing` method to the framework, I think it makes sense to add an example in the docs. Other facades also have their `missing` and `has/exists` methods documented, so I thought it made sense to do the same.

PR #10157, opened 2025-02-09
[11.x] Note about bathes ignore job's connection and queue

vadimonus:
Some discussion here https://github.com/laravel/framework/issues/54286 , some code example (combined from examples from docs) here https://github.com/vadimonus/laravel-batch-bug-report/commit/3c60ccec80c74375dcafe99f15f0b3fd5fa71808

Currently documentation only says, `All batched jobs must execute within the same connection and queue`. But having all jobs have same queue and connection explicitely set with onQueue / onConnection , they're pushed to default queue/connection. And this behaviour is not same as for Bus::chain, while other syntax is mostly the same, leading to mistakes.

This PR clarifies this behaviour.

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -1376,7 +1376,7 @@ Some tools such as Laravel Horizon and Laravel Telescope may provide more user-f
 <a name="batch-connection-queue"></a>
 #### Batch Connection and Queue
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
The current wording sounds fine.

PR #10170, opened 2025-02-17
[11.x] Improve flashed session data docs with @session directive

kaykegaldiano:
Hi!

Since Laravel 10, we have a new blade directive called @session. This directive allows us to replace `@if(session('some_value'))` with `@session('some_value')`, making the code cleaner and more up-to-date.

Would it be possible to update the documentation to reflect this?

Thanks!


taylorotwell:
I think this would need explanation of where "value" comes from.

PR #10180, opened 2025-02-19
[11.x] Enhance eventStream to Support Custom Events and Start Messages

devhammed:
#### **Summary**
This PR enhances the `eventStream` method in `ResponseFactory` by allowing customization of:  

- **Event Name (`$as`)** – Previously hardcoded as `"update"`, now configurable.  
- **Start Stream Message (`$startStreamWith`)** – Optional message sent before streaming begins.  
- **End Stream Message (`$endStreamWith`)** – Now optional instead of required.  

#### **Why This Change?**

- **More Flexible SSE Handling**: Developers can now specify custom event names, making it easier to differentiate between different types of SSE messages and also they can now decide if they want start/end messages.
- **Improved Client Communication**: Some clients expect an initial message when connecting. The `$startStreamWith` parameter allows for this. 
- **Backward Compatible**: Defaults maintain previous behaviour, ensuring no breaking changes.  

#### **Implementation Details**

The method signature now includes:  

  ```php
  public function eventStream(
      Closure $callback, 
      array $headers = [], 
      string $as = 'update', 
      ?string $startStreamWith = '<stream>', 
      ?string $endStreamWith = '</stream>'
  )
```

PR: https://github.com/laravel/framework/pull/54695

### DIFF
--- a/responses.md
+++ b/responses.md
@@ -351,12 +351,17 @@ Route::get('/chat', function () {
 });
 ```
 
-This event stream may be consumed via an [EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) object by your application's frontend. The `eventStream` method will automatically send a `</stream>` update to the event stream when the stream is complete:
+This event stream may be consumed via an [EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) object by your application's frontend. The `eventStream` method will automatically send a `<stream>` update to the event stream when the stream starts and a `</stream>` update when it is completed:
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Reverted this PR. The update "as" name should be customizable per update, not for the entire event stream.

devhammed:
> Reverted this PR. The update "as" name should be customizable per update, not for the entire event stream.

I can't really think of a use case where I would want that though since each endpoint will want to return a specific formatted message...

PR #10186, opened 2025-02-19
[12.x] remove `<?php` opening tag in code blocks

browner12:
we're being explicit now with our fenced code blocks, rather than relying on guessing, so we don't need to give the guesser any extra help.

this also brings consistency because we have a mix of code blocks that use it and ones that don't.

also, IMO, it's just noise on the example


taylorotwell:
I don't think I love this change. They aren't directly copy and pastable anymore in many cases.

browner12:
hmm..ok. wouldn't have thought that's how people are using them.

PR #10192, opened 2025-02-21
Add note to Str::wrap() about differences to Str::finish(Str::start(…), …)

shaedrich:
as discussed in https://github.com/laravel/framework/pull/53987#issuecomment-2557253882


taylorotwell:
I don't think this is needed.

PR #10195, opened 2025-02-23
Replace ternary operator with @selected

Rattone:
Moved after value for better readability


taylorotwell:
I think using the raw version is more informative here of what is going on.

PR #10204, opened 2025-02-25
[12.x] Add UUID v7 section for Models to upgrade guide

HasanQQ:
With laravel 12, the `HasUuids` trait using v7 UUIDs by default. If we compare the examples below, the order of the database records may become confusing.

```
Str::orderedUuid()
Ramsey\Uuid\Rfc4122\UuidV4 {#5343
  uuid: "9e4bb9bd-f73d-41c2-aecf-8c9bf81eb188",
}

Str::uuid7()
Ramsey\Uuid\Lazy\LazyUuidFromString {#5295
  uuid: "01953cbc-ceac-73a4-9829-a14fc6528306",
}
```

I'm not sure about my writing style, but I think there should be such a section in the upgrade documentation.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -134,6 +134,21 @@ $table = Schema::getTableListing(schema: 'main', schemaQualified: false);
 
 The `db:table` and `db:show` commands now output the results of all schemas on MySQL, MariaDB, and SQLite, just like PostgreSQL and SQL Server.
 
+<a name="models"></a>
+### Models
+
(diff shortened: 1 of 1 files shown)


nshiro:
We should also mention that the `HasVersion7Uuids` trait was removed in Laravel 12.
If you are using it, you should migrate to the `HasUuids` trait.
https://github.com/laravel/framework/pull/52433

PR #10208, opened 2025-02-27
 add step parameter to Collection::range() method

Ashot1995:
## Summary
This PR updates the Laravel Collections documentation to reflect the new `step` parameter in the `range()` method. 

## Changes
- Updated `collections.md` to include an example of `range($from, $to, $step)`.
- Clarified how the `step` parameter affects number generation in collections.

## Example Usage
```php
collect()->range(3, 9, 2); // [3, 5, 7, 9]

### CODE REVIEW

File: collections.md [hunk 1]
```diff
@@ -2373,14 +2373,14 @@ $random->all();
...
+The `range` method returns a collection containing numbers between the specified range, with an optional step parameter:
 
 ```php
-$collection = collect()->range(3, 6);
+$collection = collect()->range(3, 6, 2);
 
 $collection->all();
 
-// [3, 4, 5, 6]
+// [3, 5, 7, 9]
```
shaedrich, line 2383: Maybe, the `$step` parameter shouldn't modify the original example but be added as an additional one 🤔

.idea/workspace.xml
shaedrich, line 1: Are you sure, this should have been committed?

taylorotwell:
Should be two different sections imo.

PR #10214, opened 2025-02-28
Add documentation for `Context::scope()`

cosmastech:
Documentation for new `Context::scope()`. https://github.com/laravel/framework/pull/54799

### DIFF
--- a/context.md
+++ b/context.md
@@ -155,7 +155,7 @@ Context::when(
 <a name="scoped-context"></a>
 #### Scoped Context
 
-The `scope()` method provides a way to temporarily modify the context for a given callback and restore to the original state. Any changes made to both context and hidden data within the closure are isolated and automatically discarded once the closure completes. Additionally, you can pass extra data to be merged into the context (as the second and third arguments) while the closure executes.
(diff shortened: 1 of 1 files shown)


context.md
```diff
@@ -152,6 +152,41 @@ Context::when(
 );
 ```
 
+<a name="scoped-context"></a>
+#### Scoped Context
+
+The `scope()` method provides a way to temporarily modify the context for a given callback and restore to the original state when complete. Any changes made to both context and hidden data within the closure are isolated and automatically discarded once the closure completes. Additionally, you can pass extra data to be merged into the context (as the second and third arguments) while the closure executes.
+
+```php
+use Illuminate\Support\Facades\Context;
```
shaedrich, line 161: According to #10186, we prefer having PHP opening tags in the code sample blocks:
```suggestion
<?php

use Illuminate\Support\Facades\Context;
```

context.md
```diff
@@ -152,6 +152,41 @@ Context::when(
...
+Context::all();
+// []
+
+Context::allHidden();
+// [
+//     'user_id' => 123,
+// ]
+```
+
+If an object within the context is modified inside the closure, that mutation will be reflected outside the scope. To avoid unintended side effects, use immutable data or create copies as needed.
```
shaedrich, line 188: You might want to highlight this:
```suggestion
> [!NOTE]
> If an object within the context is modified inside the closure, that mutation will be reflected outside the scope. To avoid unintended side effects, use immutable data or create copies as needed.
```

taylorotwell:
Drafting until release day.

PR #10220, opened 2025-03-01
Clarify ASCII numerals in `alpha_dash` and `alpha_num` validation rules

austincarpenter:
ASCII characters in the `\p{L}` regex ranges are clarified as `a-z` and `A-Z` for the `alpha_dash` and `alpha_num` validation rules. ASCII numerals in the `\p{N}` range (`0-9`) are not.

This PR adds ", and `0-9`" where relevant.

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -1214,7 +1214,7 @@ To restrict this validation rule to characters in the ASCII range (`a-z` and `A-
 
 The field under validation must be entirely Unicode alpha-numeric characters contained in [`\p{L}`](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AL%3A%5D&g=&i=), [`\p{M}`](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AM%3A%5D&g=&i=), [`\p{N}`](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AN%3A%5D&g=&i=), as well as ASCII dashes (`-`) and ASCII underscores (`_`).
 
-To restrict this validation rule to characters in the ASCII range (`a-z` and `A-Z`), you may provide the `ascii` option to the validation rule:
+To restrict this validation rule to characters in the ASCII range (`a-z`, `A-Z` and `0-9`), you may provide the `ascii` option to the validation rule:
 
 ```php
 'username' => 'alpha_dash:ascii',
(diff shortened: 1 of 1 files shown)


austincarpenter:
Should the `alpha_dash` description go one step further?  I.e.:

"To restrict this validation rule to characters in the ASCII range (`a-z`, `A-Z`, `0-9`, `-` and `_`), you may provide..."

I feel like most people would always opt for the `:ascii` variant of the rule and having the ranges comma-separated like this without describing them makes visually identifying the rules much easier.

validation.md
```diff
@@ -1214,7 +1214,7 @@ To restrict this validation rule to characters in the ASCII range (`a-z` and `A-
 
 The field under validation must be entirely Unicode alpha-numeric characters contained in [`\p{L}`](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AL%3A%5D&g=&i=), [`\p{M}`](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AM%3A%5D&g=&i=), [`\p{N}`](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AN%3A%5D&g=&i=), as well as ASCII dashes (`-`) and ASCII underscores (`_`).
```
shaedrich, line 1215: Maybe, we could highlight the "numeric" as `alpha_dash` looks like the `alpha` rule plus an additional dash, not the `alpha_num` rule plus dash:
```suggestion
The field under validation must be entirely Unicode alpha-**numeric** characters contained in [`\p{L}`](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AL%3A%5D&g=&i=), [`\p{M}`](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AM%3A%5D&g=&i=), [`\p{N}`](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AN%3A%5D&g=&i=), as well as ASCII dashes (`-`) and ASCII underscores (`_`).
```

austincarpenter: I agree there's a bit of ambiguity between `alpha_dash` and `alpha_num` rules (the former also accepting numbers). I don't think the addition of bold here fixes the problem though.

shaedrich: What would you suggest then? How would you make it more clear that `alpha_dash` is—in parts—a misnomer, so to speak?

PR #10225, opened 2025-03-03
docs: add decodedPath method to request documentation

mohammadrasoulasghari:
This PR adds documentation for the decodedPath method alongside the existing path method documentation. The addition clarifies how to handle URL-encoded characters in request paths by showing a practical example of the difference between these two methods.


taylorotwell:
I think it's OK for now. 👍

mohammadrasoulasghari:
@taylorotwell 
So why was it closed?!

PR #10232, opened 2025-03-06
Fixed typo in sail:publish artisan command

fabioferreira3:


sail.md
```diff
@@ -560,7 +560,7 @@ If you're using PhpStorm, please review JetBrains' documentation regarding [zero
 Since Sail is just Docker, you are free to customize nearly everything about it. To publish Sail's own Dockerfiles, you may execute the `sail:publish` command:
 
 ```shell
-sail artisan sail:publish
+php artisan sail:publish
```
mohammadrasoulasghari, line 563: When using Laravel Sail, we should use `sail artisan` instead of `php artisan` to ensure commands run inside the container. Running `php artisan` directly while using Sail is incorrect as it executes on the host machine rather than in the containerized environment.

taylorotwell:
I think this is actually fine. See comment above.

PR #10237, opened 2025-03-08
[12.x] Clarify error handling behavior in errors.md

AhmedAlaa4611:
This update clarifies that Laravel ignores certain errors from being logged or sent to external error tracking service, rather than completely ignoring them.
This makes it clearer that the errors still occur but are simply not logged by default.


errors.md
```diff
@@ -199,7 +199,7 @@ class PodcastProcessingException extends Exception implements ShouldntReport
 }
 ```
 
-Internally, Laravel already ignores some types of errors for you, such as exceptions resulting from 404 HTTP errors or 419 HTTP responses generated by invalid CSRF tokens. If you would like to instruct Laravel to stop ignoring a given type of exception, you may use the `stopIgnoring` exception method in your application's `bootstrap/app.php` file:
+Internally, Laravel already ignores some types of errors from being logged or sent to your application's external error tracking service, such as exceptions resulting from 404 HTTP errors or 419 HTTP responses generated by invalid CSRF tokens. If you would like to instruct Laravel to stop ignoring a given type of exception, you may use the `stopIgnoring` exception method in your application's `bootstrap/app.php` file:
```
mohammadrasoulasghari, line 202: This is a good sentence, and thank you. 

If I may suggest a clearer wording for readers:

` from being reported to logs or external monitoring tools`

AhmedAlaa4611: Thanks for the suggestion! I see your point, but I believe my version aligns better with the existing terminology used in the documentation.

For example, in the `throttling-reported-exceptions` section in the same page docs, there’s a section that states:

`If your application reports a very large number of exceptions, you may want to throttle how many exceptions are actually logged or sent to your application's external error tracking service.`

To maintain consistency, I used `logged or sent to your application's external error tracking service`

Would love to hear your thoughts on keeping the wording uniform across the docs!

see:
https://laravel.com/docs/12.x/errors#throttling-reported-exceptions

mohammadrasoulasghari: My goal in suggesting this revision is simply to make it clearer for those who are reading the documentation. The more clear and understandable a product’s documentation is, and the more it uses common terminology, the easier it becomes to learn from that documentation.

For example, `monitoring tools` is a clearer phrase that quickly conveys your intention in adding this sentence, and…

PR #10238, opened 2025-03-08
[12.x] Rework pagination Tailwind JIT description; add v4 syntax

rozsazoltan:
Rework: https://laravel.com/docs/12.x/pagination#tailwind-jit

In Tailwind CSS v3, JIT mode is enabled by default (which was introduced in v2). I believe that Laravel 12.x is far enough from Tailwind CSS v2 that we can drop this naming convention and simply refer to it as v3.

* https://v3.tailwindcss.com/docs/upgrade-guide#migrating-to-the-jit-engine

The release of Tailwind CSS v4 introduced many breaking changes. Automatic source detection was introduced, but unfortunately, the vendor folder was excluded due to the .gitignore file. Adding the source now works with a different syntax compared to v3.

* https://tailwindcss.com/docs/detecting-classes-in-source-files#which-files-are-scanned
* https://tailwindcss.com/docs/functions-and-directives#source-directive

```css
@import "tailwindcss";
@source "./../../vendor/laravel/framework/src/Illuminate/Pagination/resources/views";
```


rozsazoltan:
|Old|New|
|---|---|
|image|image|

AhmedAlaa4611:
Also, I think that https://laravel.com/docs/12.x/vite#working-with-stylesheets needs to be updated like that https://tailwindcss.com/docs/upgrade-guide#using-postcss

PR #10240, opened 2025-03-09
[12.x] Mention different Vite integration for Tailwind v3 vs v4

rozsazoltan:
Before the this PR, the documentation included a reference to the new v4 installation while explaining the old v3 integration.

### What's changed?

I moved the mention of Start Kits to the beginning.

After that, I introduced the TailwindCSS v4 Vite plugin (`@tailwindcss/vite`), illustrating it in a code snippet alongside the `laravel-vite-plugin`.

Then, I kept the original v3 description about PostCSS plugins and their automatic integration into Vite.

### DIFF
--- a/vite.md
+++ b/vite.md
@@ -445,6 +445,30 @@ The following example demonstrates how Vite will treat relative and absolute URL
 <a name="working-with-stylesheets"></a>
 ## Working With Stylesheets
 
+> [!NOTE]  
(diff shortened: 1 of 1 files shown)


rozsazoltan:
I'm not sure how much we want to elaborate on this section. Considering the useful mention of the connection between PostCSS and Vite, and that TailwindCSS v4 only available from January 2025, I thought it makes sense to keep the description of v3, which is still used by many.

AhmedAlaa4611:
Hi @rozsazoltan, I think it should be updated also in the `puls` docs

See: https://laravel.com/docs/12.x/pulse#custom-card-styling-tailwind

PR #10243, opened 2025-03-11
Adds a mention to autoreloading and using multiple processes

tonysm:
### Changed

- Adds a mention to the autoreloading behavior of env variable changes being disabled when using multiple processes (https://github.com/laravel/framework/pull/54972)


taylorotwell:
Let me think about this one. I'm not sure it warrants it's own section in the docs. 🤔

PR #10249, opened 2025-03-13
[12.x] Fix strings whenEmpty code example

RobertBoes:
The code example for the `whenEmpty` method is incorrect, as `whenEmtpy` doesn't execute the closure for a string containing spaces. The string would first need to be trimmed.

```php
$string = Str::of('  ')->whenEmpty(function (Stringable $string) {
    return $string->trim()->prepend('Laravel');
});

// '  '

$string = Str::of('  ')->trim()->whenEmpty(function (Stringable $string) {
    return $string->prepend('Laravel');
});

// 'Laravel'
```

### DIFF
--- a/strings.md
+++ b/strings.md
@@ -3393,8 +3393,8 @@ The `whenEmpty` method invokes the given closure if the string is empty. If the 
 use Illuminate\Support\Str;
 use Illuminate\Support\Stringable;
 
-$string = Str::of('  ')->whenEmpty(function (Stringable $string) {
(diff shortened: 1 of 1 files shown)


RobertBoes:
I see this code snippet has been used since 7.x, probably when the feature was introduced (https://laravel.com/docs/7.x/helpers#method-fluent-str-when-empty). So either the example snippet was always wrong, or behavior of `whenEmpty` changed somewhere.

PR #10250, opened 2025-03-13
[12.x] docs: document container respecting property default values

calebdw:
See: https://github.com/laravel/framework/pull/53522

Thanks!


AhmedAlaa4611:
Also, I think that this should be documented in the `providers.md` or in `container.md` docs

to make the users know how the Resolution will react with default values

PR #10256, opened 2025-03-16
Update collections.md, add diffKeysUsing method

sonichandni:
Add `diffKeysUsing ` a collection method which can be usefull when we need to perform key comparisons with custom logic beyond the default behavior.


taylorotwell:
Will work on integrating this.

PR #10268, opened 2025-03-20
[12.x] Add `ddBody` to http-tests.md

Sammyjo20:
Added documentation for the ddBody method added in Laravel 12.2


AhmedAlaa4611:
Are `ddJson`, and `ddBody` not supported with `PHPUnit`?

PR #10275, opened 2025-03-22
Fix broken code blocks

pxlrbt:


taylorotwell:
This will get fixed when I merge docs forward.

PR #10279, opened 2025-03-24
Add documentation for Collection getOrPut method

mohammadrasoulasghari:
In this PR, I've added documentation for the getOrPut method to the Collection documentation.


taylorotwell:
I don't particularly like using collections as singular objects.

mohammadrasoulasghari:
@taylorotwell 
So I don't think this method should be public. Right?

PR #10280, opened 2025-03-24
[12.x] Pessimistic locking SQL

nevstas:


nevstas:
Also should fix 11 version and master

PR #10282, opened 2025-03-26
Improve `Collection::reduce` docs

timacdonald:


collections.md
```diff
@@ -2402,7 +2402,7 @@ $collection->reduce(function (int $carry, int $item) {
 // 10
 ```
 
-The `reduce` method also passes array keys in associative collections to the given callback:
+The `reduce` method also passes array keys to the given callback:
```
timacdonald, line 2405: The method always receives the key. Having access to the key is also useful for lists.

collections.md
```diff
@@ -2417,9 +2417,9 @@ $ratio = [
     'eur' => 1.22,
 ];
 
-$collection->reduce(function (int $carry, int $value, int $key) use ($ratio) {
+$collection->reduce(function (int $carry, int $value, string $key) use ($ratio) {
```
timacdonald, line 2420: This code example would currently fail as the `$key` in this example is a string.

collections.md
```diff
@@ -2417,9 +2417,9 @@ $ratio = [
...
     return $carry + ($value * $ratio[$key]);
-});
+}, 0);
```
timacdonald, line 2422: In a strict setting, this example would fail as it will attempt to pass `null` to an `int` typed parameter.

PR #10286, opened 2025-03-28
Update concurrency compatibility note

Willem-Jaap:
Update concurrency compatibility note to account for users that have already migrated to `providers.php`


taylorotwell:
I'm not sure people will know what this is.

PR #10301, opened 2025-04-01
fix: states eloquent scopes return affects query building

Hyrran:
States that you need to use Model::query() if your scope has a void return.

Might cause some confusion, as the docs does not use the query() method and the scopes return is void.


Hyrran:
I don't know if this is the expected behavior, but @newtonjob has some good points that we could prevent this kind of confusion.
https://github.com/laravel/framework/pull/54450#discussion_r2023708038

shaedrich:
Not sure if changing the docs is the right approach here. It's unfortunate that some devs have this problem right now in using the attribute, but this might be fixed soon when #55246 is merged, I'd assume.

PR #10302, opened 2025-04-02
[12.x] Clarify that table method accepts both arrays and collections

AhmedAlaa4611:
**Description:**
This update clarifies that the second parameter of the `table` method can be either an `array` or a `collection`. If a collection is provided, it is automatically converted to an array.

See: https://github.com/laravel/framework/blob/a60c75a96a7a53e886ecf9c4579e5e11ca5548ee/src/Illuminate/Console/Concerns/InteractsWithIO.php#L237

### DIFF
--- a/artisan.md
+++ b/artisan.md
@@ -697,7 +697,7 @@ $this->newLine(3);
 <a name="tables"></a>
 #### Tables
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
This is noted in the docblock.

AhmedAlaa4611:
@taylorotwell I think it's important to consider the documentation from the reader's point of view. While the docblock provides technical details for developers working directly with the code, the documentation that users read should be clear, highlighting key features and behaviors in a way that is easy to understand.

PR #10303, opened 2025-04-02
[12.x] Query scopes for `tap` and `pipe`

timacdonald:


queries.md
```diff
@@ -1468,6 +1469,121 @@ DB::transaction(function () {
 });
 ```
 
+<a name="scopes"></a>
+## Scopes
```
shaedrich, line 1473: I wonder if that is "officially" considered a "scope" in the stricter sense 🤔

taylorotwell:
Thanks @timacdonald

queries.md
timacdonald: Turns out: it isn't!

PR #10308, opened 2025-04-04

aryanjoshidev:
now its quite clear that we are talking about foreign key column of some table


taylorotwell:
I think it's already pretty clear.

PR #10318, opened 2025-04-08
[12.x] Add example for passing a callback to collection sort method

AhmedAlaa4611:
**Description:**
This PR adds a practical example demonstrating the use of a callback with the `sort` method on a Laravel collection. The new example sorts the collection by placing even numbers first, followed by odd numbers.

Additionally, the sentence referencing PHP’s `uasort()` has been moved after the new example to maintain logical flow, since the example helps illustrate how sort internally relies on `uasort()`.

**Reasoning:**
Clarifies the purpose and power of using a callback with `sort` method.

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -2839,7 +2839,29 @@ $sorted->values()->all();
 // [1, 2, 3, 4, 5]
 ```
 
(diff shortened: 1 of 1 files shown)


taylorotwell:

PR #10319, opened 2025-04-09
[12.x] Make collection bold in intersect method description

AhmedAlaa4611:
**Description:**
In the `intersect` method documentation, the word `array` is bolded, but `collection` is not. Highlighting both equally (either bold both or none) keeps the emphasis balanced and consistent. If the bolding is being used to emphasize acceptable types, it should be applied to both: array and collection.

From a UX/UI perspective, applying bold styling to both terms ensures visual balance and avoids unintentionally drawing more attention to one accepted input type over the other.

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -1465,7 +1465,7 @@ $collection->implode(function (array $item, int $key) {
 <a name="method-intersect"></a>
 #### `intersect()` {.collection-method}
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Probably neither one of them should be bold. 😅

AhmedAlaa4611:
Solved later on: https://github.com/laravel/docs/commit/b6829f947a084978934266cce2d02b6289499aba

PR #10325, opened 2025-04-12
Update routing.md

aryanjoshidev:
this is quite clear to user now as a beginner in laravel,  just separated both codes nothing much. but i think  if you accept the changes this will be a great help to new comers like me.

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -596,16 +596,13 @@ Since the `$user` variable is type-hinted as the `App\Models\User` Eloquent mode
 Of course, implicit binding is also possible when using controller methods. Again, note the `{user}` URI segment matches the `$user` variable in the controller which contains an `App\Models\User` type-hint:
 
 ```php
-
 use App\Http\Controllers\UserController;
 
 // Route definition...
 Route::get('/users/{user}', [UserController::class, 'show']);
-
 ```
 
 ```php
-
 use App\Models\User;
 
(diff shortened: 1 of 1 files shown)


routing.md
```diff
@@ -597,10 +597,13 @@ Of course, implicit binding is also possible when using controller methods. Agai
...
 
 // Route definition...
 Route::get('/users/{user}', [UserController::class, 'show']);
+```
+
+```php
+use App\Models\User;
 
 // Controller method definition...
 public function show(User $user)
```
shaedrich, line 609: @aryanjoshidev Just fyi: As Laravel uses [Torchlight](https://torchlight.dev) for fences code block rendering, you can use the `filename` info string attribute
```suggestion
```php filename=routes/web.php
use App\Http\Controllers\UserController;

// Route definition...
Route::get('/users/{user}', [UserController::class, 'show']);
```

```php filename=app/Controllers/UserController.php
use App\Models\User;

// Controller method definition...
public function show(User $user)
```

shaedrich: Technically, one could also resort to the use of tabs

````suggestion
```php tab="Route definition"
use App\Http\Controllers\UserController;

Route::get('/users/{user}', [UserController::class, 'show']);
```

```php tab="Controller method definition"
use App\Models\User;

public function show(User $user)
````

PR #10326, opened 2025-04-12
Document Form Request failed validation hooks

panda-madness:
This adds some documentation about these method's existence. They have been in the framework since at least version 5, and there is a lot of posts about them out on the internet, but no mention in the docs currently.


taylorotwell:
I don't think these examples show exactly what their use case would be for most applications.

PR #10327, opened 2025-04-13
Adds `AsColection::map()` section

DarkGhostHunter:
As part of [#55383](https://github.com/laravel/framework/pull/55383).

### DIFF
--- a/eloquent-mutators.md
+++ b/eloquent-mutators.md
@@ -468,7 +468,7 @@ protected function casts(): array
 }
 ```
 
-For better control on how the items should be constructed once inside the Collection, you may set a callable that receives the items as an array and returns an object. 
+For better control on how the items should be constructed once inside the Collection, you may set a callable that receives each item as an array.
 
 ```php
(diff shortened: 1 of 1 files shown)


eloquent-mutators.md
```diff
@@ -449,6 +449,85 @@ protected function casts(): array
 }
 ```
 
+If you desire the collection items to be mapped into an specific class instance, you may use a second parameter or the `map()` method if you want to use the base Collection class.
+
+```php
+use App\ValueObjects\Option;
```
shaedrich, line 455: According to #10186, we prefer having PHP opening tags in the code sample blocks:
```suggestion
```php
<?php

use App\ValueObjects\Option;
```

DarkGhostHunter: Opening tags should go on full-code examples, I believe.

shaedrich: Yeah, it is rather superfluous, but one can one do when Taylor decides that way 🤷🏻 😂

PR #10334, opened 2025-04-19
[12.x] update batch table generation command

Puneetk1103:
I noticed the documentation mentioned that we need to use 
```
php artisan make:queue-batches-table
```

however, Checking `php artisan` the command I think it should be 
```
php artisan queue:batches-table 
```


taylorotwell:
Update your Laravel installation.

PR #10345, opened 2025-04-23
[12.x] use multiline parameters for casting methods

browner12:
I find the side scrolling for code snippets to decrease readability. these lengthy method signatures are often the culprit on this page. switching to multiline parameters to help avoid the necessity to side scroll.

followed the PSR-12 recommendation for when dealing with multiline parameters AND a return type.

I know this will be a constant battle in the documentation, and while changes like this won't completely eliminate the problem, I do think they help.

I would also like to see the Laravel documentation site make better use of the space it has available. On larger screens we have the ability to increase the container size. I would also suggest there is a lot of unused space between the columns of our 3 column layout. Changes here would also go a long way to help remove the need for side scrolling.  Since the `laravel.com` repo has been set to private, I don't believe the community has the ability to make these suggestions anymore.

### DIFF
--- a/eloquent-mutators.md
+++ b/eloquent-mutators.md
@@ -690,8 +690,12 @@ class AsJson implements CastsAttributes
      * @param  array<string, mixed>  $attributes
      * @return array<string, mixed>
      */
-    public function get(Model $model, string $key, mixed $value, array $attributes): array
-    {
+    public function get(
+        Model $model,
+        string $key,
+        mixed $value,
+        array $attributes,
+    ): array {
         return json_decode($value, true);
     }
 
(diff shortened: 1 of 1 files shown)


shaedrich:
> side scrolling

Where do you need side-scrolling? On the phone? It's not like these examples are overly long oO

browner12:
On desktop. You can find an example of side-scrolling on basically every page.

AhmedAlaa4611:
> I would also like to see the Laravel documentation site make better use of the space it has available.

I agree with you, these changes will not solve the problem.

> Since the `laravel.com` repo has been set to private

Why it is set to private? and why also the `Bootcamp` is not available?

PR #10357, opened 2025-04-28
Instructions for composer on existing apps

pimpeters:
These instruction were in previous versions as well, without these it's unable to install the composer packages (including sail itself) without having an local development environs already.

### DIFF
--- a/sail.md
+++ b/sail.md
@@ -9,6 +9,7 @@
 - [Executing Commands](#executing-sail-commands)
     - [Executing PHP Commands](#executing-php-commands)
     - [Executing Composer Commands](#executing-composer-commands)
+    - [Installing Composer Dependencies for Existing Applications](#installing-composer-dependencies-for-existing-projects)
     - [Executing Artisan Commands](#executing-artisan-commands)
     - [Executing Node / NPM Commands](#executing-node-npm-commands)
 - [Interacting With Databases](#interacting-with-sail-databases)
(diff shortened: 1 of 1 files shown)


msonowal:
Hi @taylorotwell,
Thank you for the ongoing work on Laravel and the ecosystem improvements.
I wanted to ask about the recommended approach for users who don't have Composer installed locally. Previously, having a simple Docker-only setup provided a very clean starting point for new projects.
While I understand the benefits of using Herd or php.new, there are some scenarios where these solutions may not be feasible:

Corporate environments where Herd installation isn't permitted
Situations where Docker is the only whitelisted development tool
Cases where developers prefer not to modify their existing PHP installations or extensions
(Say only want to test few things with certain things without system wide changes lets say in Arc/Linux)

Would you be able to share what the recommended workflow would be for these situations?

martinbean:
> I wanted to ask about the recommended approach for users who don't have Composer installed locally. Previously, having a simple Docker-only setup provided a very clean starting point for new projects.

@msonowal I know it’s not the “official” response you’re looking for, but just wanted to reply to say that I personally use Composer via a Docker image on my Mac.

I’ve aliased `composer` to the following:

```sh
function composer {
    docker run --interactive --rm --tty --volume $PWD:/app --volume ${COMPOSER_HOME:-$HOME/.composer}:/tmp composer $*
}
```

This will work just like Composer normally. You can call sub-commands (e.g. `composer install`) as well as pass options.

I’ll also create a **Makefile** for my projects so that I can clone the project, run `make build` and get a Sail-based environment created and instantiated. A sample **Makefile** might look like this:

```make
DOCKER := docker
COMPOSER := $(DOCKER) run --interactive --rm --tty --volume $(shell pwd):/app composer
SAIL := vendor/bin/sail

all: install build

.env:
	cp .env.example .env

install: .env
	$(COMPOSER) install

build: install
	$(SAIL) up --detach
	$(SAIL) npm install
	$(SAIL) npm run build
	$(SAIL) artisan key:generate
	$(SAIL) artisan migrate:fresh --seed
	$(SAIL) open

clean:
	$(SAIL) down --remove-orphans --rmi all --volumes
```

So I can run `sail build` to create my project, or `sail clean` to completely clean up after it.

msonowal:
@martinbean that is a workaround, I know I also have lot of ways but I am talking about in general for a new comers that provides a official instruction which gives less friction

martinbean:
> @martinbean that is a workaround, I know I also have lot of ways but I am talking about in general for a new comers that provides a official instruction which gives less friction

@msonowal It’s not really a “workaround”. It uses a Docker container to install Composer dependencies; it just uses the official `composer` image rather than the `laravelsail/php8*-composer` image Laravel were maintaining for… whatever reason.

That being said, I don’t really think it’s the job of the Laravel framework docs to also teach people how to use Docker.

msonowal:
Hi @martinbean , I understand what you are replying but my original intention was to have something from officially mentioned in the docs so one dont have to search github comments like yours for various method to quickly get it running.

PR #9627, opened 2024-05-02
Document tag replacement in localization docs

LegendEffects:
Added in https://github.com/laravel/framework/pull/51190

Still not 100% on the wording, especially around the unescaped blade syntax, but gets the basis there. 

I'm also not sure on what sort of code styling is best for the example so I've used what StyleCI reformatted the PR to

PR #10364, opened 2025-04-30
Document tag replacement in localization docs

LegendEffects:
Reopening this as it has been almost a year since I last tried to get this stuff documented to no avail. (https://github.com/laravel/docs/pull/9627)

**ORIGINAL**

Added in https://github.com/laravel/framework/pull/51190

Still not 100% on the wording, especially around the unescaped blade syntax, but gets the basis there.

I'm also not sure on what sort of code styling is best for the example so I've used what StyleCI reformatted the PR to


LegendEffects:
Guess it will never be documented then

PR #10368, opened 2025-05-01
Query builder first mention

BhSimon:
This is the first mention of `Builder` on the page. Should the `use` reference be included here since there are several possible imports?

### DIFF
--- a/queries.md
+++ b/queries.md
@@ -568,7 +568,7 @@ $users = DB::table('users')
 If you need to group an "or" condition within parentheses, you may pass a closure as the first argument to the `orWhere` method:
 
 ```php
-use Illuminate\Database\Query\Builder;
+use Illuminate\Database\Query\Builder; 
 
 $users = DB::table('users')
     ->where('votes', '>', 100)
--- a/queries.md
+++ b/queries.md
@@ -568,6 +568,8 @@ $users = DB::table('users')
 If you need to group an "or" condition within parentheses, you may pass a closure as the first argument to the `orWhere` method:
 
 ```php
+use Illuminate\Database\Query\Builder;
+
 $users = DB::table('users')
     ->where('votes', '>', 100)
     ->orWhere(function (Builder $query) {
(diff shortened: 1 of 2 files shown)


scout.md
```diff
@@ -168,7 +168,7 @@ Additional settings and schema definitions for your Typesense collections can be
 <a name="preparing-data-for-storage-in-typesense"></a>
 #### Preparing Data for Storage in Typesense
 
-When utilizing Typesense, your searchable model's must define a `toSearchableArray` method that casts your model's primary key to a string and creation date to a UNIX timestamp:
+When utilizing Typesense, your searchable models must define a `toSearchableArray` method that casts your model's primary key to a string and creation date to a UNIX timestamp:
```
shaedrich, line 171: Good catch! 👍🏻

queries.md
```diff
@@ -568,6 +568,8 @@ $users = DB::table('users')
 If you need to group an "or" condition within parentheses, you may pass a closure as the first argument to the `orWhere` method:
 
 ```php
+use Illuminate\Database\Query\Builder;
```
shaedrich, line 571: According to #10186, we prefer having PHP opening tags in the code sample blocks:
```suggestion
```php
<?php

use Illuminate\Database\Query\Builder;
````

shaedrich (commented):
Fair point 👍🏻

queries.md
AhmedAlaa4611: No, I don't think so and we do not use `<?php` in all examples.
In a simple example like this, we don't use it across the whole docs.
In a large one that includes the whole file with namespace we use `<?php`.
last but not lest, this is not related to that PR

shaedrich: Tell Taylor that you don't want to do what he wants, not me

shaedrich: Taylor misses quite a few things actually. browner removed `<?php` this for the exact case where you advocated adding it: Examples namespaces. Taylor was against that change, hence, he was in favor of having these files keep their `<?php`

AhmedAlaa4611: I am now reading the collections docs, although there are many examples out there; but `<?php` is never used.

PR #10369, opened 2025-05-02
Fix "Facade Class Reference" table rendering (Firefox)

jbruni:
Here is how it currently renders (how I see the table):

image

By removing the `overflow-auto` class, here is how it renders to me:

image

On Windows, this happens in Firefox, but not in Chrome. I have not tested other browsers or systems.


shaedrich:
Alternatively, we could solve the spacing issue already addressed in #10345

AhmedAlaa4611:
> Alternatively, we could solve the spacing issue already addressed in https://github.com/laravel/docs/pull/10345

Yes, I agree with you.

taylorotwell:
Hmm, I'm not seeing any issues on Firefox and I'm a bit afraid to just remove the class since that could cause other issues.

PR #10373, opened 2025-05-05
[12.x] Clarify unless method usage in collections docs

AhmedAlaa4611:
Description
---
This PR updates the description for the `unless` method to clarify that both the collection instance and the first argument are passed to the closure similar to the behavior of the `when` method.

As the `when` method explicitly states that:

> The collection instance and the first argument given to the when method will be provided to the closure:

Let's do the same for the `unless` method to maintain consistency and improve clarity for developers reading the documentation, especially since `when` and `unless` are behave similarly in terms of callback parameters.

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -3457,16 +3457,16 @@ This method has the same signature as the [unique](#method-unique) method; howev
 <a name="method-unless"></a>
 #### `unless()` {.collection-method}
 
-The `unless` method will execute the given callback unless the first argument given to the method evaluates to `true`:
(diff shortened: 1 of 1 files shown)


taylorotwell:
Why do you use `int` when the first value given is a boolean?

AhmedAlaa4611:
Hi @taylorotwell,
I think `bool` is more convenient than `int`, but in the `when` method example we used `int` not `bool`. So for consistency, I used `int` in the `unless` example as well.

Do you think we should update both examples to use `bool` instead?

taylorotwell:
Sorry, I don't understand. The types don't match.

PR #10374, opened 2025-05-06
[12.x] Add docs about using 'exists' rule on array

DeffiDeffi:
I recently stumbled upon this bit of (to my knowledge) undocumented behaviour of the `exists` rule. 

I assumed that `exists` would not work on array values because the documentation did not specify that it would. So I validated incoming arrays of IDs like this:

```php
'user_ids' => ['array'],
'user_ids.*' => ['integer', 'exists:users,id'],
```

This queries the database for every single item in the array. Instead, you can specify the 'exists' rule as:

```php
'user_ids' => ['array', 'exists:users,id'],
```

... which only generates a single `WHERE id IN (...)` query. This PR adds a little note on this to the docs.

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -1677,6 +1677,16 @@ You may explicitly specify the database column name that should be used by the `
 'state' => Rule::exists('states', 'abbreviation'),
 ```
 
+<a name="validate-an-array-of-values-against-database"></a>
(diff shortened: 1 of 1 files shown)


shaedrich:
Good catch 👍🏻

PR #10377, opened 2025-05-07
[12.x] Be more specific about usage of recycle() 

bertheyman:
[Recycle only works for relationships defined inside of the factory](https://github.com/laravel/framework/pull/55271).
This PR makes that more clear in the docs, and refers to the hasAttached section for many-to-many relationships.

### DIFF
--- a/eloquent-factories.md
+++ b/eloquent-factories.md
@@ -616,8 +616,8 @@ public function definition(): array
 <a name="recycling-an-existing-model-for-relationships"></a>
 ### Recycling an Existing Model for Relationships
 
-If you have models that have a one-to-one or one-to-many relationship, you may use the `recycle` method to ensure a single instance of the related model is recycled for all of the relationships created by the factory.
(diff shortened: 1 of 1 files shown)


taylorotwell:
Going to skip this one for now.

bertheyman:
> If you have models that share a common relationship with another model, you may use the recycle method to ensure a single instance of the related model is recycled for all of the relationships created by the factory.

@taylorotwell @shaedrich 
I feel the current explanation isn't fully clear: a many-to-many relationships can also be created by the factory, and will not respect recycle(). Would you like me to tweak the wording for this PR, or drop it altogether?

shaedrich:
I can only speak for myself, but for me, something like this would probably suffice:
```markdown
> NOTE:
> For relationships outside of the factory (like a many-to-many), see [HasAttached()](#pivot-table-attributes).
```

bertheyman:
Thanks, good suggestion!
Can't update this PR, so I'll add a new one and see what Taylor thinks.

PR #10380, opened 2025-05-07
[12.x] Add `Authenticated` Attribute in Container

milwad-dev:
This PR updates the Laravel documentation to include an example of using the new `#[Authenticated]` attribute alongside the #[Auth] attribute in controller constructors.

This addition complements the existing examples of other container attributes like `#[Cache]`, `#[Config]`, and `#[DB]`, providing a more complete picture of how to leverage attribute-based dependency injection in Laravel controllers.


taylorotwell:
I would rather just show one option.

PR #10381, opened 2025-05-08
Update undocumented precision in documentation

DCrepper:
In the Number helper class currency method precision not documented, but exist in the code, there is a lot of currency that do not use any subb currency Like (HUF)


DCrepper:
I think 10 or 11 when introduced need to be also change like 11.x or 10.x i dont know exactly when. but its very use full, and i have idea about set default precision based on local.

PR #10384, opened 2025-05-09
[12.x] Document optional arguments for query builder function: distinct()

AdamGaskins:
See https://github.com/laravel/framework/pull/27675


AdamGaskins:
Actually this doesn't work the way I thought. Probably would be confusing to add to the docs as is.

PR #10389, opened 2025-05-10
[12.x] Add support enum for `assertJsonPath` method

milwad-dev:
Recently, the `assertJsonPath` method supports enum classes.

### DIFF
--- a/http-tests.md
+++ b/http-tests.md
@@ -588,6 +588,12 @@ class ExampleTest extends TestCase
 }
 ```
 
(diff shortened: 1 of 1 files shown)


AhmedAlaa4611:
I will leave that here: https://github.com/laravel/framework/pull/55516

shaedrich:
I wonder, if methods that accept enums should just have an admonition right between the first paragraph and the first code sample

````markdown
 #### Asserting on JSON Paths
If you would like to verify that the JSON response contains the given data at a specified path, you should use the `assertJsonPath` method:

> NOTE:
> The second parameter also accepts enums

```php tab=Pest
````

PR #10393, opened 2025-05-12
Update eloquent-relationships.md to include detail about 'morphs' migration helper

n1c:


shaedrich:
Not sure if this is the right place, as the more intuitive place would be the migration docs page

PR #10400, opened 2025-05-14
[12.x] Add `deleteWhen` method in job middleware

milwad-dev:
https://github.com/laravel/framework/pull/55718

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -738,7 +738,7 @@ public function middleware(): array
 }
 ```
 
-If you would like to delete a job by condition, you can use the `deleteWhen` method:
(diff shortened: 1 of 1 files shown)


queues.md
```diff
@@ -738,6 +738,15 @@ public function middleware(): array
 }
 ```
 
+If you would like to delete a job by condition, you can use the `deleteWhen` method:
```
shaedrich, line 741: As we've discussed in https://github.com/laravel/framework/pull/55718#issuecomment-2874616990, the terminology might not intuitive. Therefore, I suggest enriching the explanation with information taken from your PR description:
```suggestion
Contrary to using `->when(…)`, which indicates that the job should not be released back and instead throw an exception, `->deleteWhen(…)` allows you to exit the circuit of either throw exception or release back onto the queue, and instead to just delete the job. In this way, the job will just be deleted when this exception occurs. (Of course, you can still choose to report the exception just like before.):
```

PR #10419, opened 2025-05-20
[12.x] Add `UseFactory` attribute in Factories

milwad-dev:


taylorotwell:
Let me think on this one. Want to keep our attribute usage consistent across the framework.

PR #10423, opened 2025-05-21
[12.x] Clarify where the code should be placed

AhmedAlaa4611:
Problem
---
The original example lacks context about where the `DB::listen` logic is expected to live, which may confuse developers unfamiliar with the typical usage patterns.

Proposed Fix
---
Add a comment to clarify that the `DB::listen` code should be placed in a service provider.

Reasoning
---
This small change improves readability by helping readers understand where the code belongs within a typical Laravel application structure.

### DIFF
--- a/context.md
+++ b/context.md
@@ -227,6 +227,7 @@ Stacks can be useful to capture historical information about a request, such as 
 use Illuminate\Support\Facades\Context;
 use Illuminate\Support\Facades\DB;
 
+// In our service provider...
 DB::listen(function ($event) {
     Context::push('queries', [$event->time, $event->sql]);
 });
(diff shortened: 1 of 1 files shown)


AhmedAlaa4611:
I use the same tone that we are used in [that section](https://laravel.com/docs/12.x/context#how-it-works) example:

```php
// In our middleware...
Context::add('url', $request->url());
Context::add('trace_id', Str::uuid()->toString());

// In our controller...
ProcessPodcast::dispatch($podcast);
```

context.md
```diff
@@ -227,6 +227,7 @@ Stacks can be useful to capture historical information about a request, such as
 use Illuminate\Support\Facades\Context;
 use Illuminate\Support\Facades\DB;
 
+// In our service provider...
```
shaedrich, line 230: Or you could use the `filename` code fence attribute:
````markdown
```php filename=app/Providers/AppServiceProvider.php
use Illuminate\Support\Facades\Context;
use Illuminate\Support\Facades\DB;

DB::listen(function ($event) {
    Context::push('queries', [$event->time, $event->sql]);
});
```
````

PR #10428, opened 2025-05-22
add documentation for Str::split method

abhihyder:
##  Documentation for  `Str::split` Method

This PR adds documentation for a new `Str::split` method that is being proposed in the Laravel framework.

### About the Method

`Str::split` is a convenient, expressive wrapper around PHP’s native `explode()` function. It allows developers to split a string by a given separator, with optional support for a limit.

This method is particularly helpful for developers working across multiple languages like JavaScript or Python, where `split` is the standard method for string separation—providing familiarity and consistency in Laravel’s `Str` API.

### Example Usage

```php
use Illuminate\Support\Str;

Str::split('-', 'laravel-framework');
// ['laravel', 'framework']

Str::split(',', 'one,two,three,four', 3);
// ['one', 'two', 'three,four']
```

### DIFF
--- a/strings.md
+++ b/strings.md
@@ -89,6 +89,7 @@ Laravel includes a variety of functions for manipulating string values. Many of 
 [Str::singular](#method-str-singular)
 [Str::slug](#method-str-slug)
 [Str::snake](#method-snake-case)
+[Str::split](#method-str-split)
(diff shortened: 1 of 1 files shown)


taylorotwell:

PR #10432, opened 2025-05-22
[12.x] Documenting the unless method in conditional context

AhmedAlaa4611:
Description
---
This PR updates the `Conditional Context` section to include an explanation and example for the `unless` method provided by the `Conditionable` trait.

Just like `when`, the `unless` method is available on the Context facade and may be used to conditionally add data to the context. This addition ensures that both conditional helpers (`when` and `unless`) are clearly documented for developers who may prefer either style.

I’ve used the same tone and structure of the existing `when` method documentation to ensure consistency.

### DIFF
--- a/context.md
+++ b/context.md
@@ -162,6 +162,19 @@ Context::when(
 );
 ```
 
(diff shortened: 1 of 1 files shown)


shaedrich:
Is there really an own paragraph and an example needed to say "hey, it also works the other way round?"

taylorotwell:
I think it's fine to leave it alone.

PR #10438, opened 2025-05-23
[12.x] Move Excluding Middleware inside Middleware Groups section

El-Virus:
The `Excluding Middleware` section was inside `Assigning Middleware to Routes`, where groups have not yet been introduced.
This PR moves `Excluding Middleware` to the `Middleware Groups` section.


taylorotwell:
I think it's fine where it is.

PR #10445, opened 2025-05-25
feat(validation.md): Add explanation and example for nullable validat…

mohammadRezaei1380:
This pull request adds a clear explanation and a practical example for the nullable validation rule in the validation.md documentation for Laravel 12.
It helps developers understand how nullable works and when other validation rules are skipped if the field is null.

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -940,6 +940,15 @@ After defining this value, the validation rule will produce the following error 
 ```text
 The credit card number field is required when payment type is credit card.
 ```
+### nullable
+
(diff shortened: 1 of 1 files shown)


AhmedAlaa4611:
It is already documented, and we mentioned it several many times across the docs.

AhmedAlaa4611:
See: https://laravel.com/docs/12.x/validation#a-note-on-optional-fields

mohammadRezaei1380:
Thanks for the note! I understand that nullable is mentioned in various places throughout the docs.
My goal was to provide a concise and clear explanation with a simple example all in one place for easier reference—especially helpful for new users who might miss scattered mentions.

PR #10448, opened 2025-05-25
[12.x] add explanation and example for prohibited_if rule

mohammadRezaei1380:
This pull request adds a dedicated explanation and usage example for the `prohibited_if` validation rule to the Laravel 12 documentation.

It helps clarify how to conditionally prohibit a field based on the value of another field, which is useful in dynamic form validation scenarios.


taylorotwell:
This seems to already be documented?

PR #10453, opened 2025-05-27
[12.x] Improving the return type annotation

AhmedAlaa4611:
Description
---
This PR updates the PHPDoc for the `backoff()` method to use `list<int>` instead of `array<int, int>`. Since the method returns a numerically indexed array with consecutive integer keys starting from 0, I think `list<int>` more accurately describes the return type.

Note
---
I do not know how the `backoff()` method behaves behind the scenes.

### DIFF
--- a/events.md
+++ b/events.md
@@ -569,7 +569,7 @@ You may easily configure "exponential" backoffs by returning an array of backoff
 /**
  * Calculate the number of seconds to wait before retrying the queued listener.
  *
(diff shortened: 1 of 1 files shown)


shaedrich:
While this is certainly not wrong, overall, it doesn't really matter as they are identical.

If this was a code review comment on the PR introducing the method, this would be a good finding. But making a PR just for this, is a lot of overhead.

PR #10459, opened 2025-05-28
Remove duplicated section

AmirKhalifehSoltani:
The "Performing Additional Validation"  topic with it's description was exactly duplicated in the validation.md file. The wrong one is omitted know.


AmirKhalifehSoltani:
> I believe this is intentional for visibility.

But I got confused for some seconds: "what is the difference?!"
I think it can be confusing for others like me.


Anyway, it’s up to you — whatever you think is best.

PR #10464, opened 2025-06-01
fix frontend.md typo: add comma between Turbo and Hotwire links

fershopls:


shaedrich:
Are you sure, because Turbo is by Hotwire?

PR #10478, opened 2025-06-06
[12.x] Add documentation for higher order static calls on collections

AhmedAlaa4611:
Description
---
This PR adds documentation for the newly introduced support for "higher order static calls" on collections, see: laravel/framework/pull/55880.

Higher order static calls allow developers to call static methods on class names stored within a collection using a clean and expressive syntax, similar to existing "higher order messages" for instance methods.

An example is included to demonstrate how this feature can be used with a collection of transformer classes.

This complements the existing section on "higher order messages" and helps developers understand and take advantage of this new feature when working with collections of class names.

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -5,6 +5,7 @@
     - [Extending Collections](#extending-collections)
 - [Available Methods](#available-methods)
 - [Higher Order Messages](#higher-order-messages)
+- [Higher Order Static Calls](#higher-order-static-calls)
 - [Lazy Collections](#lazy-collections)
     - [Introduction](#lazy-collection-introduction)
     - [Creating Lazy Collections](#creating-lazy-collections)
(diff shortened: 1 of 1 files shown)


collections.md
```diff
@@ -3994,6 +3995,40 @@ $users = User::where('group', 'Development')->get();
 return $users->sum->votes;
 ```
 
+<a name="higher-order-static-calls"></a>
+## Higher Order Static Calls
+
+Collections also provide support for "higher order static calls", which are short-cuts for calling static methods elegantly if you’re storing class names in collections.
```
shaedrich, line 4001: You probably want to use an actual apostrophe here 😉
```suggestion
Collections also provide support for "higher order static calls", which are short-cuts for calling static methods elegantly if you're storing class names in collections.
```

shaedrich, line 4001: Not sure if "shortcut" isn't the more common version 🤔 but I'm not a native speaker, so 🤷🏻‍♂️  
```suggestion
Collections also provide support for "higher order static calls", which are shortcuts for calling static methods elegantly if you’re storing class names in collections.
```

AhmedAlaa4611: I used the same tone that we used in the section above that one, see: [Higher Order Messages](https://laravel.com/docs/12.x/collections#higher-order-messages)

AhmedAlaa4611: In dusk docs we used both `shortcuts` and `short-cuts`, so I will open a PR to unify that across the docks.

Thank you for this.

AhmedAlaa4611: I am not sure, but we do not use `you’re` we use `you're` across the whole docs see:
- https://github.com/search?q=repo%3Alaravel%2Fdocs+%22you%27re%22&type=code
- https://github.com/search?q=repo%3Alaravel%2Fdocs+%22you%E2%80%99re%22&type=code

shaedrich: But **you're** not.

That's what I meant 😉

taylorotwell:
I dunno if we really need to document this at the moment. Thanks though!

PR #10481, opened 2025-06-06
[12.x] Improve readability of HTTP client response methods by using a table

AhmedAlaa4611:
Description
---
This PR updates the section on inspecting "HTTP client responses" by converting the current list of available methods into a table format. This improves readability, allows for quicker scanning, and makes it easier for developers to understand the available methods and their return types at a glance.

Benefits
---
- More accessible for beginners and quicker for experienced developers to reference.
- Follows the pattern used in other parts of the documentation for summarizing methods.

### DIFF
--- a/http-client.md
+++ b/http-client.md
@@ -36,20 +36,24 @@ $response = Http::get('http://example.com');
 
 The `get` method returns an instance of `Illuminate\Http\Client\Response`, which provides a variety of methods that may be used to inspect the response:
 
-```php
-$response->body() : string;
-$response->json($key = null, $default = null) : mixed;
-$response->object() : object;
-$response->collect($key = null) : Illuminate\Support\Collection;
-$response->resource() : resource;
-$response->status() : int;
-$response->successful() : bool;
-$response->redirect(): bool;
-$response->failed() : bool;
-$response->clientError() : bool;
-$response->header($header) : string;
-$response->headers() : array;
-```
+<div class="overflow-auto">
+
+| Method                                           | Returns                       |
+| ------------------------------------------------ | ----------------------------- |
+| `$response->body();`                             | string                        |
+| `$response->json($key = null, $default = null);` | mixed                         |
+| `$response->object();`                           | object                        |
+| `$response->collect($key = null);`               | Illuminate\Support\Collection |
+| `$response->resource();`                         | resource                      |
+| `$response->status();`                           | int                           |
+| `$response->successful();`                       | bool                          |
+| `$response->redirect();`                         | bool                          |
+| `$response->failed();`                           | bool                          |
+| `$response->clientError();`                      | bool                          |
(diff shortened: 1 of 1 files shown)


http-client.md
```diff
@@ -36,20 +36,24 @@ $response = Http::get('http://example.com');
...
+| `$response->object();`                           | object                        |
+| `$response->collect($key = null);`               | Illuminate\Support\Collection |
+| `$response->resource();`                         | resource                      |
+| `$response->status();`                           | int                           |
+| `$response->successful();`                       | bool                          |
+| `$response->redirect();`                         | bool                          |
+| `$response->failed();`                           | bool                          |
+| `$response->clientError();`                      | bool                          |
+| `$response->header($header);`                    | string                        |
+| `$response->headers();`                          | array                         |
```
shaedrich, line 54: Are you sure, you don't want code formatting for the second column?
```suggestion
| Method                                           | Returns                          |
| ------------------------------------------------ | -------------------------------- |
| `$response->body();`                             | `string`                         |
| `$response->json($key = null, $default = null);` | `mixed`                          |
| `$response->object();`                           | `object`                         |
| `$response->collect($key = null);`               | `\Illuminate\Support\Collection` |
| `$response->resource();`                         | `resource`                       |
| `$response->status();`                           | `int`                            |
| `$response->successful();`                       | `bool`                           |
| `$response->redirect();`                         | `bool`                           |
| `$response->failed();`                           | `bool`                           |
| `$response->clientError();`                      | `bool`                           |
| `$response->header($header);`                    | `string`                         |
| `$response->headers();`                          | `array`                          |
```

http-client.md
```diff
@@ -36,20 +36,24 @@ $response = Http::get('http://example.com');
...
-$response->header($header) : string;
-$response->headers() : array;
-```
+<div class="overflow-auto">
+
+| Method                                           | Returns                       |
+| ------------------------------------------------ | ----------------------------- |
+| `$response->body();`                             | string                        |
+| `$response->json($key = null, $default = null);` | mixed                         |
+| `$response->object();`                           | object                        |
```
shaedrich, line 45: This might be an interesting read in terms of the return type:
https://github.com/laravel/framework/pull/54160

http-client.md
AhmedAlaa4611: Thank you for this.

shaedrich: You're welcome.

taylorotwell:
Looks like this gets cut off on my screen 🤔

CleanShot 2025-06-09 at 09 15 42@2x

AhmedAlaa4611:
I have reduced the width of the table by removing bold text, I think now it will be good in all screens. Also, it will be cool if we solved the spacing issue already addressed in #10345 and #10369

taylorotwell:
Hmm, I think I prefer the current version.

PR #10486, opened 2025-06-08
[12.x] Add make:job command with --batched option to queue documentation

amirhshokri:
This PR adds documentation for the `php artisan make:job` command with the `--batched` option, which allows the creation of batchable jobs.

This feature was introduced in: https://github.com/laravel/framework/pull/55929


taylorotwell:
We don't document every option as they are available via `php artisan help make:job`

PR #10489, opened 2025-06-09
Nullable User in Pulse Gate definition

razvan-z:
The `Gate` is failing all the time when defining it in the format from documentation.


AhmedAlaa4611:
Hi @razvan-z, just a quick note; in cases like that PR, if you need to make additional changes, you can simply push another commit to the same branch. There's no need to close the PR and open a new one. Keeping everything in one PR helps maintain continuity and context.

PR #10493, opened 2025-06-10
[12.x] Nullable User in Pulse Gate definition

razvan-z:
The `Gate` is always leading to 403 when defining it in the format from documentation.


taylorotwell:
Possibly because you aren't logged in?

PR #10500, opened 2025-06-12
[12.x] Document missing offsetExists and offsetGet methods on collections

AhmedAlaa4611:
Description
---
This PR adds documentation for the `offsetExists()` and `offsetGet()` methods available on the `Illuminate\Support\Collection` class. These methods are part of the [ArrayAccess interface](https://www.php.net/manual/en/class.arrayaccess.php) implementation and allow for key-based access and checks on collections.

These methods come from `ArrayAccess`, but documenting them directly makes them easier to discover and understand.

Question
---
Should we also document `offsetSet()` and `offsetUnset()`, or simply note that the Collection class implements the `ArrayAccess` interface?

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -179,6 +179,8 @@ For the majority of the remaining collection documentation, we'll discuss each m
 [mode](#method-mode)
 [multiply](#method-multiply)
 [nth](#method-nth)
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't think we really need to document these. 👍

PR #10501, opened 2025-06-13
[12.x] Clarify the behavior of collapseWithKeys method

AhmedAlaa4611:
Description
---
This PR clarifies the behavior of the `collapseWithKeys` method when used on an already flat collection. It helps developers understand that the method will return an empty collection in such cases, which might not be immediately obvious.

See: https://github.com/laravel/framework/pull/56002


AhmedAlaa4611:
Drafting until release day.

PR #10507, opened 2025-06-17
[11.x] Add warning about potential breaking change with eager loaded queries when using pivot columns for filtering or ordering

pintend:
I am updating an old Laravel project (8) and in the process of upgrading 10 -> 11, I came across this issue. Yes its probably the wrong way to do things, but this is legacy code which i didn't write... I think it should be included in the upgrade guide for anyone else who may have been using such a query.


taylorotwell:
This feels so deep into an edge case I'm going to draft for now and think about it.

PR #10510, opened 2025-06-18
[12.x] Typo in Queues > Failing Jobs on Specific Exceptions

cosmastech:
Moved the closing quotation mark to the line above.


shaedrich:
Maybe, there should be a pipeline step to validate the code samples. This seems to happen fairly often and goes unnoticed for quite a while sometimes … 🤔

PR #10511, opened 2025-06-18
[12.x] Include the SFTP driver

AhmedAlaa4611:
Description
---
This PR updates the docs for the `url` method to include the `SFTP` driver, which was previously omitted from the explanation. The PR also adds the missing:

```php
'url' => env('SFTP_URL'),
```

configuration option to the `SFTP` filesystem configuration example to help users understand how to properly configure `SFTP` disks to work with the `url` method, as this parameter is essential for generating accessible URLs.

Note
---
The `SFTP` driver is placed between `local` and `s3` to maintain consistency with the driver ordering established elsewhere in the documentation.


taylorotwell:
I dunno if we really need to emphasize SFTP much. People should be using S3 or similar.

PR #10516, opened 2025-06-20
[12.x] only use and document the `limit` and `offset` methods

browner12:
`skip()` and `take()` are alias methods to `offset()` and `limit()`. IMO we should only use and document the main methods, not the aliases.  This will encourage more performant coding, just like we now do in `laravel/framework`.

IF it's decided we still want to mention the aliases, I'd recommended listing them second as the alternatives.  I can make that change if desired.

follow up to https://github.com/laravel/framework/pull/56080 and https://github.com/laravel/framework/pull/56081

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -431,7 +431,7 @@ The Eloquent `all` method will return all of the results in the model's table. H
 ```php
 $flights = Flight::where('active', 1)
     ->orderBy('name')
-    ->take(10)
+    ->limit(10)
     ->get();
 ```
 
--- a/queries.md
+++ b/queries.md
@@ -1210,16 +1210,7 @@ To build more advanced `having` statements, see the [havingRaw](#raw-methods) me
 <a name="limit-and-offset"></a>
 ### Limit and Offset
 
-<a name="skip-take"></a>
-#### The `skip` and `take` Methods
-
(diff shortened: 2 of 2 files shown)


shaedrich:
Well, what do we have aliases for if we make sure nobody knows about them?

browner12:
agreed, we should get rid of the aliases. but we can't because of BC.

shaedrich:
> we should get rid of the aliases

Well, that is _your_ opinion.

And I'm sure, Taylor has something to say about that, too.

AhmedAlaa4611:
Proposal
---
Remove `skip()` and `take()` in Laravel 13 to reduce ambiguity and enforce consistency with SQL terminology. As aliases for `offset()` and `limit()`, they introduce unnecessary abstraction.

I appreciate that `skip()` and `take()` have served as convenient aliases for many developers, especially those newer to SQL. However, after gaining a deeper understanding of the underlying SQL concepts, I believe these methods can be misleading.

Since Laravel 13 will allow for breaking changes, this might be an appropriate opportunity to consider removing `skip()` and `take()` from the core.

shaedrich:
> I appreciate that `skip()` and `take()` have served as convenient aliases for many developers, especially those newer to SQL.

If you are looking for a lean framework, Laravel might not be what you are looking for. Laravel is and always has been a framework that offered multiple ways of doing something.

PR #10517, opened 2025-06-21
[12.x] Rename the person parameter to persons.

kevinb1989:
I believe that the `person` parameter should be renamed to `persons` in this context.


taylorotwell:
Probably generally ok as-is but if we were gonna change it we would probably use "people"?

AhmedAlaa4611:
I think it may be "users"

PR #10521, opened 2025-06-22
[12.x] Use consistent collect([...]) syntax in random() example output

AhmedAlaa4611:
Description
---
This PR updates the example output of the `random()` method to match the consistent output style used throughout the collections docs when the output is a collection (e.g., in `pop()` and `shift()` examples). This change improves consistency.

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -2376,7 +2376,7 @@ $random = $collection->random(3);
 
 $random->all();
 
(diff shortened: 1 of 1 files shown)


AhmedAlaa4611:
I forget that the `all()` method returns the underlying array represented by the collection.

PR #10522, opened 2025-06-22
Add async to http client

diego-lipinski-de-castro:
Added `async` documentation for the http client.

This is my first time contributing to the project. I am not the greatest writer and I think the explanation could be improved. Any feedback is openly appreciated.


taylorotwell:
We haven't really fleshed out an ergonomic API around Guzzle promises so going to leave this one undocumented for now.

PR #10527, opened 2025-06-24
[12.x] Format the Queued Email Failures code

AhmedAlaa4611:
Description
---
This PR utilizing the imported `Illuminate\Queue\SerializesModels` trait.


shaedrich:
Assuming, the changed line is the only one, I'd recommend renaming the PR to "Remove unused import" (or alternatively, put this into the description if you should insist on the current title)

browner12:
the import is not visible in the code example we see, but is more than likely present in any `Mail` class that someone generates.  this is purely because we're seeing a truncated view, so I'd recommend leaving it in.

PR #10532, opened 2025-06-24
[12.x] Refine phrase

AhmedAlaa4611:
Description
---
In the `abort` function we used the phrase `You may also provide the exception's message...` but in the `abort_if` and `abort_unless` functions we used `you may also provide the exception's response text...`, so let's refine them.


taylorotwell:
I think response text makes it a bit more clear this will actually be sent to the user.

shaedrich:
Doesn't that me, the docs for `abort()` have to be changed instead?

PR #10533, opened 2025-06-24
[12.x] Add broadcast_if and broadcast_unless methods to helpers

AhmedAlaa4611:
Description
---
Add documentation for the newly introduced methods in: https://github.com/laravel/framework/pull/55967

Note
---
I used the same tone that we used in `broadcast()`, `abort_if()`, and `abort_unless` methods.

Question
---
I'm not sure how to document the `FakePendingBroadcast` in this context. Could you take care of it, or do you think it doesn't need documentation?

### DIFF
--- a/helpers.md
+++ b/helpers.md
@@ -170,6 +170,8 @@ Laravel includes a variety of global "helper" PHP functions. Many of these funct
 [bcrypt](#method-bcrypt)
 [blank](#method-blank)
 [broadcast](#method-broadcast)
+[broadcast_if](#method-broadcast-if)
+[broadcast_unless](#method-broadcast-unless)
 [cache](#method-cache)
 [class_uses_recursive](#method-class-uses-recursive)
 [collect](#method-collect)
--- a/helpers.md
+++ b/helpers.md
@@ -2256,6 +2256,24 @@ broadcast(new UserRegistered($user));
 broadcast(new UserRegistered($user))->toOthers();
 ```
 
+<a name="method-broadcast-if"></a>
+#### `broadcast_if()` {.collection-method}
+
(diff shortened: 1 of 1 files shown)


helpers.md
```diff
@@ -2256,6 +2258,24 @@ broadcast(new UserRegistered($user));
...
+
+<a name="method-broadcast-unless"></a>
+#### `broadcast_unless()` {.collection-method}
+
+The `broadcast_unless` function [broadcasts](/docs/{{version}}/broadcasting) the given [event](/docs/{{version}}/events) to its listeners if a given boolean expression evaluates to `false`:
+
+```php
+broadcast_unless($user->isBanned(), new UserRegistered($user));
+```
+
```
shaedrich, line 2278: I wonder if it won't be enough to mention, that `broadcast` has conditional methods for both the `if` and the `unless` case. I'd argue that the concept of conditional methods should be common enough for a Laravel dev so that it doesn't require a reintroduction for each method pair.

AhmedAlaa4611: I think they should be documented for two reasons:

1) We have a documentation for:
- `abort()`, `abort_if()`, and `abort_unless()`.
- `report()`, `report_if()`, and `report_unless()`.

> So, for consistency and completeness they should be documented.

2) We have the `dispatch()` helper, but do not have `dispatch_if()`, and `dispatch_unless()`.

> So, for beginners they should be documented.

shaedrich: > for consistency and completeness they should be documented.

As always, this is not the best argument since maybe, we should question this, too

> We have the dispatch() helper, but do not have dispatch_if(), and dispatch_unless().
>
> > So, for beginners they should be documented.

I don't know what that has to do with our discussion.

Also, that would mean that we have to spell _everything_ out, expecting absolute beginners everywhere. However, that isn't what the documentation currently does, either.

PR #10534, opened 2025-06-24
[12.x] document the password reset `cache` driver

browner12:
document https://github.com/laravel/framework/pull/53428

I modeled the wording and organization after the [session](session.md) docs for hopefully a little consistency.

this is a resubmission of #10032. there were a lot of merge conflicts since I had based that off 11.x, so figured it was easier to start fresh.

@taylorotwell you made a comment in that PR about if I could "implement that automatic hashing of the keys so we don't need to worry about prefixes".  I could not find any prior reference to that request, and I wasn't exactly sure what you were asking, so I wanted to get some clarification before this is merged.

I think my confusion lies in I'm so used to hashing for passwords via `password_hash()` (non-deterministic) that I forgot there's also `hash()` (deterministic). Are you asking if we could use `hash($userEmail)` as our cache key, and then drop the "prefix" completely?   I guess that does reduce chance of collision with userland, but does lock it in and force the user to not use that key, where the prefix gives them control. I'm on the fence so I'll defer to you, but I'll document either way.

The other thing to be aware of with dropping the "prefix" option now is that the feature has been out for a bit, and we'd likely break some implementations. Given that most expirations shouldn't be too long, I don't think it'd be a huge deal, but definitely something to be aware of.

---

### Reference PRs

https://github.com/laravel/laravel/pull/6487
https://github.com/laravel/framework/pull/53448
https://github.com/laravel/framework/pull/53561

### DIFF
--- a/passwords.md
+++ b/passwords.md
@@ -28,8 +28,8 @@ The password reset `driver` configuration option defines where password reset da
 
 <div class="content-list" markdown="1">
 
-- `database` - password resets are stored in a relational database.
-- `cache` - password resets are stored in one of your cache based stores.
+- `database` - password reset data is stored in a relational database.
+- `cache` - password reset data is stored in one of your cache based stores.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


browner12:
I *think* this is what you were asking for:

https://github.com/laravel/framework/pull/56127

PR #10543, opened 2025-06-27
[12.x] Fix anchor links in Collections docs for consistency

AhmedAlaa4611:
Description
---
This PR updates several anchor links in the `collections.md` to use kebab-case to align with the consistent format used throughout the docs. For example:

- #method-chunkwhile → #method-chunk-while
- #method-intersectAssoc → #method-intersect-assoc
- #method-countBy → #method-count-by

These changes improve overall consistency in the docs, as some method anchors such as `first-or-fail`, `first-where`, and `reduce-spread` already follow kebab-case formatting.

### DIFF
--- a/collections.md
+++ b/collections.md
@@ -35,7 +35,7 @@ As mentioned above, the `collect` helper returns a new `Illuminate\Support\Colle
 $collection = collect([1, 2, 3]);
 ```
 
-You may also create a collection using the [make](#method-make) and [fromJson](#method-fromjson) methods.
+You may also create a collection using the [make](#method-make) and [fromJson](#method-from-json) methods.
 
 > [!NOTE]
 > The results of [Eloquent](/docs/{{version}}/eloquent) queries are always returned as `Collection` instances.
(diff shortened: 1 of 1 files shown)


AhmedAlaa4611:
I know that this may be a big PR, but I made it cautiously and reviewed it three times.

taylorotwell:
Unfortunately it would break all existing links on the web to these methods.

AhmedAlaa4611:
@taylorotwell Should we target the master branch instead?

shaedrich:
This would work for `https://laravel.com/docs/13.x/*` links but not for `https://laravel.com/docs/*` (version omitted, shortcut for current version) links

AhmedAlaa4611:
@shaedrich I understand you, but should we leave them not in a kebab-case forever? or there is a way to update them?

shaedrich:
Technically, you could probably create a redirect at https://github.com/laravel/bootcamp.laravel.com/blob/main/routes/web.php to preserve backward compatibility, but I'm not sure if Taylor would be willing to add that kind of complexity to that repository

AhmedAlaa4611:
@shaedrich 

> I'm not sure if Taylor would be willing to add that kind of complexity to that repository

I think that too, can we update them in all `https://laravel.com/docs/*`? what do you think about that?

shaedrich:
I'm afraid, I can't follow 🤔

AhmedAlaa4611:
I mean updating all branches with these changes.

shaedrich:
But the problem will remain as other sites will link here with the current version shortcut. And they won't all update their links

AhmedAlaa4611:
Maybe, but I've seen many articles that share tips or list top ten helpful methods in Laravel without referencing the official documentation. They rely solely on their own explanations, so I don't think it's a very important issue to worry about.

shaedrich:
That may be true. I just wanted to voice my concerns and my interpretation of Taylor's words earlier. But I don't really know the impact, it would potentially have.

PR #10550, opened 2025-06-29
[12.x] Fix grammatical inconsistency

AhmedAlaa4611:
Description
---
This PR updates a sentence to improve grammatical accuracy. The original wording uses "is" which implies present, alongside "escaped" which implies "past" at the same time.

### DIFF
--- a/blade.md
+++ b/blade.md
@@ -161,7 +161,7 @@ Sometimes you may pass an array to your view with the intention of rendering it 
 </script>
 ```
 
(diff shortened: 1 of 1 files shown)


shaedrich:
As far as I know, both is possible:
* when "escaped" is a _past tense verb_ (as you are implying), your assumption holds true
* when "escaped" is an a _adjective-like past participle_, I'd say, it doesn't

but in the end, I'm not a native speaker 😅

PR #10552, opened 2025-06-30
[12.x] Add examples for faking events in testing with fakeExceptFor method

MrPunyapal:
For this one https://github.com/laravel/framework/blob/12.x/src/Illuminate/Support/Facades/Event.php#L107C1-L120C1


taylorotwell:
This feels quite complicated to be honest. Will just leave it as is for now.

PR #10560, opened 2025-07-03
[12.x] Add proper links

AhmedAlaa4611:
Description
---
This PR enhances the `encrypt()` and `decrypt()` helpers by adding links to point to each other.

Note
---
I used the same tone that we used with the `blank()` and `filled()` helpers.

### DIFF
--- a/helpers.md
+++ b/helpers.md
@@ -2425,6 +2425,8 @@ The `decrypt` function [decrypts](/docs/{{version}}/encryption) the given value.
 $password = decrypt($value);
 ```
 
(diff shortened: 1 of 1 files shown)


AhmedAlaa4611:
Should we use the word "function" or the word "helper"?

shaedrich:
or "helper function" (or "utility function", but that would clash with our naming) ^^

AhmedAlaa4611:
> but that would clash with our naming

Yes, I think that because the file name is `helpers.md`, not `utilities.md`.

PR #10561, opened 2025-07-03
[12.x] Clarify collapse method behavior

AhmedAlaa4611:
Description
---
This PR clarifies the behavior of the `collapse` method to reflect that it can collapse both arrays and collections. Currently, the documentation only mentions arrays, but the method also supports nested Collection instances.

This change not only improves clarity and accuracy but also ensures consistency with the `collapse` and `collapseWithKeys` methods that are in the `collections` docs, which explicitly states support for both arrays and collections.

Continuation of: https://github.com/laravel/docs/pull/10370

### DIFF
--- a/helpers.md
+++ b/helpers.md
@@ -313,7 +313,7 @@ $value = Arr::boolean($array, 'name');
 <a name="method-array-collapse"></a>
 #### `Arr::collapse()` {.collection-method}
 
-The `Arr::collapse` method collapses an array of arrays into a single array:
+The `Arr::collapse` method collapses an array of arrays or collections into a single array:
 
 ```php
 use Illuminate\Support\Arr;


helpers.md
```diff
@@ -313,7 +313,7 @@ $value = Arr::boolean($array, 'name');
 <a name="method-array-collapse"></a>
 #### `Arr::collapse()` {.collection-method}
 
-The `Arr::collapse` method collapses an array of arrays into a single array:
+The `Arr::collapse` method collapses an array of arrays or collections into a single array:
 
 ```php
 use Illuminate\Support\Arr;
```
shaedrich, line 319: Wouldn't it make sense to show that in the example as well?
```diff
- $array = Arr::collapse([[1, 2, 3], [4, 5, 6], [7, 8, 9]]);
+ $array = Arr::collapse([[1, 2, 3], collect([4, 5, 6]), [7, 8, 9]]);
```

AhmedAlaa4611: I think that also, but I afraid that the PR get closed. 😂😂

PR #10584, opened 2025-07-08
Include installation of echo helper packages in broadcasting.md

guizoxxv:
The required installation step for helper packages `@laravel/echo-vue` and `@laravel/echo-react` is missing.

The previous phrase "_If you're using Laravel's React or Vue starter kits, you may listen for events using Echo's useEcho hook_" gives the impression that those hooks are available out-of-the-box when using React or Vue starter kits, but you still need to install and configure the helper packages. Also, you can use the helper packages without starter kits.

### DIFF
--- a/broadcasting.md
+++ b/broadcasting.md
@@ -78,7 +78,17 @@ All of your application's event broadcasting configuration is stored in the `con
 <a name="quickstart-next-steps"></a>
 #### Next Steps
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
This is installed by the `install:broadcasting` Artisan command.

guizoxxv:
I see what happened. In Laravel 11, those packages are not installed with `php artisan install:broadcasting`. I was looking at the wrong docs. Still, I feel that the docs give the impression that those packages are only used with starter kits. Maybe this would be more clear:

Laravel Echo provides official helper packages that simplify the integration with Vue and React applications. If you have `vue` or `react` in your dependencies, the corresponding package is automatically added when installing Node dependencies on the `php artisan install:broadcasting` command. To manually install the helper packages:

- Vue
```bash
npm install @laravel/echo-vue
```

- React
```bash
npm install @laravel/echo-react
```

Thanks.

PR #10596, opened 2025-07-11
[12.x] Improve Queues and Horizon documentation

QuentinGab:
## About

This PR is based on [a recent post I shared on X](https://x.com/QuentinGabriele/status/1938168895002558835), where I discussed some pain points around Horizon.

I did my best to avoid injecting strong personal opinions or turning the documentation into a blog post.

### Queues:
Clarifies how job attempts are counted and how failed jobs are handled. Adds updated examples and notes for middleware and retries.

### Horizon:
Adds sections on queue priorities, job attempts, timeouts, and backoff. Improves explanations for environment configuration and worker balancing.

There are still options left to document.

## Reflections on `maxException` and `tries` Defaults

While working on this PR, I strongly felt the need for an option to configure `maxExceptions` at the Horizon supervisor or queue worker level, similar to how `tries` can currently be set.

Additionally, I’m wondering whether the default value for `tries` is appropriate. Many of Laravel’s advanced queue features, such as `WithoutOverlapping` and `RateLimited`, often require increasing the `tries` value significantly, sometimes even setting it to `0`.

In my opinion, the default configuration should be more aligned with these common use cases. A more sensible default might look like:

- `tries`: 25
- `maxException`: 1

This defaults would work better out of the box for most Laravel developers .

### DIFF
--- a/horizon.md
+++ b/horizon.md
@@ -3,10 +3,10 @@
 - [Introduction](#introduction)
 - [Installation](#installation)
     - [Configuration](#configuration)
+    - [Dashboard Authorization](#dashboard-authorization)
     - [Max Job Attempts](#max-job-attempts)
     - [Job Timeout](#job-timeout)
     - [Job Backoff](#job-backoff)
-    - [Dashboard Authorization](#dashboard-authorization)
     - [Silenced Jobs](#silenced-jobs)
 - [Balancing Strategies](#balancing-strategies)
     - [Auto Balancing](#auto-balancing)
@@ -129,6 +129,32 @@ While your application is in [maintenance mode](/docs/{{version}}/configuration#
 
 Within Horizon's default configuration file, you will notice a `defaults` configuration option. This configuration option specifies the default values for your application's [supervisors](#supervisors). The supervisor's default configuration values will be merged into the supervisor's configuration for each environment, allowing you to avoid unnecessary repetition when defining your supervisors.
 
+<a name="dashboard-authorization"></a>
+### Dashboard Authorization
+
+The Horizon dashboard may be accessed via the `/horizon` route. By default, you will only be able to access this dashboard in the `local` environment. However, within your `app/Providers/HorizonServiceProvider.php` file, there is an [authorization gate](/docs/{{version}}/authorization#gates) definition. This authorization gate controls access to Horizon in **non-local** environments. You are free to modify this gate as needed to restrict access to your Horizon installation:
+
+```php
+/**
+ * Register the Horizon gate.
+ *
+ * This gate determines who can access Horizon in non-local environments.
+ */
+protected function gate(): void
+{
+    Gate::define('viewHorizon', function (User $user) {
+        return in_array($user->email, [
+            'taylor@laravel.com',
+        ]);
+    });
+}
+```
+
+<a name="alternative-authentication-strategies"></a>
+#### Alternative Authentication Strategies
+
+Remember that Laravel automatically injects the authenticated user into the gate closure. If your application is providing Horizon security via another method, such as IP restrictions, then your Horizon users may not need to "login". Therefore, you will need to change `function (User $user)` closure signature above to `function (User $user = null)` in order to force Laravel to not require authentication.
+
 <a name="max-job-attempts"></a>
 ### Max Job Attempts
 
@@ -204,32 +230,6 @@ You may also configure "exponential" backoffs by using an array for the `backoff
     ],
 ],
 ```
-
-<a name="dashboard-authorization"></a>
-### Dashboard Authorization
-
-The Horizon dashboard may be accessed via the `/horizon` route. By default, you will only be able to access this dashboard in the `local` environment. However, within your `app/Providers/HorizonServiceProvider.php` file, there is an [authorization gate](/docs/{{version}}/authorization#gates) definition. This authorization gate controls access to Horizon in **non-local** environments. You are free to modify this gate as needed to restrict access to your Horizon installation:
-
-```php
-/**
- * Register the Horizon gate.
- *
- * This gate determines who can access Horizon in non-local environments.
- */
-protected function gate(): void
-{
-    Gate::define('viewHorizon', function (User $user) {
-        return in_array($user->email, [
-            'taylor@laravel.com',
-        ]);
-    });
-}
-```
-
-<a name="alternative-authentication-strategies"></a>
-#### Alternative Authentication Strategies
-
-Remember that Laravel automatically injects the authenticated user into the gate closure. If your application is providing Horizon security via another method, such as IP restrictions, then your Horizon users may not need to "login". Therefore, you will need to change `function (User $user)` closure signature above to `function (User $user = null)` in order to force Laravel to not require authentication.
 
 <a name="silenced-jobs"></a>
 ### Silenced Jobs


queues.md
```diff
@@ -1213,7 +1215,21 @@ class ProcessPodcast implements ShouldQueue
 <a name="max-attempts"></a>
 #### Max Attempts
 
-If one of your queued jobs is encountering an error, you likely do not want it to keep retrying indefinitely. Therefore, Laravel provides various ways to specify how many times or for how long a job may be attempted.
```
QuentinGab, line 1216: This section was confusing to me, as it mixed errors and attempts, so I tried to explain as clearly as possible what an ‘attempt’ is.

horizon.md
```diff
@@ -123,6 +149,24 @@ While your application is in [maintenance mode](/docs/{{version}}/configuration#
...
+    'production' => [
+        'supervisor-1' => [
+            'connection' => 'redis',
+            'queue' => ['high', 'default', 'low'],
+        ],
+    ],
+],
+```
+
+In this example, jobs in the `high` queue will be prioritized over those in the `default` or `low` queues.
```
peaklabs-dev, line 168: I think this is only the case if `balance=false`.

horizon.md
```diff
@@ -132,7 +176,12 @@ Unlike Laravel's default queue system, Horizon allows you to choose from three w
 
 The `auto` strategy, which is the configuration file's default, adjusts the number of worker processes per queue based on the current workload of the queue. For example, if your `notifications` queue has 1,000 pending jobs while your `render` queue is empty, Horizon will allocate more workers to your `notifications` queue until the queue is empty.
 
-When using the `auto` strategy, you may define the `minProcesses` and `maxProcesses` configuration options to control the minimum number of processes per queue and the maximum number of worker processes in total Horizon should scale up and down to:
+When using the auto scaling strategy, you may define the `minProcesses` and `maxProcesses` configuration options:
```
peaklabs-dev, line 179: In my testing these work with any balacing mode or at least also with `balance=false`.

horizon.md
QuentinGab: You are right, I need rewrite this section

horizon.md
QuentinGab: Yes, that's true, but these options do not have the same meaning. To make more sense of it, I've added the `processes` key, which is an alias for `maxProcesses`

horizon.md
QuentinGab: I've just completely rewritten the section about the strategies.

horizon.md
QuentinGab: @peaklabs-dev  Ok, I just realized that I didn't understand the `false` option correctly. So I guess it was really confusing.

Really appreciate your reviews here! I've updated section about this strategy, let me know if it's sounds good to you.

PS: I’m pretty sure I don’t fully understand the `simple` one either. From what you’re saying, I’m guessing that Horizon always assigns the same number of processes to each queue when scaling up?

QuentinGab: Did some testing and it seems that when using the simple strategy, there is no scaling. All worker processes defined by maxProcesses are always running. Let me know if I'm wrong

horizon.md
```diff
@@ -92,6 +95,28 @@ You may also define a wildcard environment (`*`) which will be used when no othe
...
 When you start Horizon, it will use the worker process configuration options for the environment that your application is running on. Typically, the environment is determined by the value of the `APP_ENV` [environment variable](/docs/{{version}}/configuration#determining-the-current-environment). For example, the default `local` Horizon environment is configured to start three worker processes and automatically balance the number of worker processes assigned to each queue. The default `production` environment is configured to start a maximum of 10 worker processes and automatically balance the number of worker processes assigned to each queue.
 
+You may also define Horizon's environment independently of the application's `APP_ENV` by adding an `env` key to your Horizon configuration file:
+
+```php
+
+return [
+    /*
+    | ------------------------------------------------------------------
+    | Horizon Environement
```
browner12, line 105: spelling: "Environment"

browner12, line 105: spelling:

```suggestion
    | Horizon Environment
```

horizon.md
```diff
@@ -126,37 +151,231 @@ Within Horizon's default configuration file, you will notice a `defaults` config
...
+> This option is similar to the `--tries` option when using the Artisan command to process queues.
+
+Fine-tuning the `tries` option is essential when using middlewares such as `WithoutOverlapping` or `RateLimited` because they consume attempts. However, depending on your case, it may be more appropriate to set the job class' `$tries` property.
+
+If you do not specify a value for the `tries` option, jobs will only be attempted once or as many times as specified by the job class `$tries` property.
+
+<a name="job-timeout"></a>
+### Job Timeout
+
+Similarly, you can set a `timeout` value at the supervisor level, which specifies how many seconds a worker process can run a job before it’s forcefully terminated. Once terminated, the job will either be retried or marked as failed, depending on your queue configuration.
```
browner12, line 335: looks like you've got a special apostrophe character in there:

```suggestion
Similarly, you can set a `timeout` value at the supervisor level, which specifies how many seconds a worker process can run a job before it's forcefully terminated. Once terminated, the job will either be retried or marked as failed, depending on your queue configuration.
```

queues.md
```diff
@@ -577,6 +577,8 @@ public function middleware(): array
 }
 ```
 
+Releasing an overlapping job back onto the queue will still increment the job's total number of attempts. You may wish to tune your `tries` and `maxExceptions` properties on your job class accordingly. For exemple, leaving the `tries` property to 1 as it is by default would prevent any overlapping job to be retried later.
```
browner12, line 580: spelling:

```suggestion
Releasing an overlapping job back onto the queue will still increment the job's total number of attempts. You may wish to tune your `tries` and `maxExceptions` properties on your job class accordingly. For example, leaving the `tries` property to 1 as it is by default would prevent any overlapping job to be retried later.
```

queues.md
```diff
@@ -1213,7 +1215,22 @@ class ProcessPodcast implements ShouldQueue
 <a name="max-attempts"></a>
 #### Max Attempts
 
-If one of your queued jobs is encountering an error, you likely do not want it to keep retrying indefinitely. Therefore, Laravel provides various ways to specify how many times or for how long a job may be attempted.
+Job attempts are a core concept of Laravel's queue system and power many advanced features. While they may seem confusing at first, it's important to understand how they work before modifying the default configuration.
+
+When a job is dispatched, it is pushed onto the queue. A worker then picks it up and attempts to execute it, this is known as a job attempt.
```
browner12, line 1220: grammar:

```suggestion
When a job is dispatched, it is pushed onto the queue. A worker then picks it up and attempts to execute it. This is known as a job attempt.
```

queues.md
```diff
@@ -1213,7 +1215,22 @@ class ProcessPodcast implements ShouldQueue
...
+
+However, an attempt does not necessarily mean the job’s handle method was executed. Attempts can be consumed in several ways:
```
browner12, line 1222: apostrophe character:

```suggestion
However, an attempt does not necessarily mean the job's handle method was executed. Attempts can be consumed in several ways:
```

queues.md
```diff
@@ -1213,7 +1215,22 @@ class ProcessPodcast implements ShouldQueue
...
+- The job's `handle` method runs and completes without throwing an exception.
+- The job encounters an unhandled exception during execution.
+- The job timed out.
+- The job is manually released back to the queue using `$this->release()`.
+- Middleware such as `WithoutOverlapping` or `RateLimited` fails to acquire a lock and releases the job.
+
+You likely do not want it to keep retrying indefinitely a job. Therefore, Laravel provides various ways to specify how many times or for how long a job may be attempted.
+
+> [!NOTE]
+> By default, Laravel will only attempt a job once. If your job uses middleware like `WithoutOverlapping` or `RateLimited`, or if you're manually releasing jobs, you’ll likely need to increase the number of allowed attempts.
```
browner12, line 1233: apostrophe character:

```suggestion
> By default, Laravel will only attempt a job once. If your job uses middleware like `WithoutOverlapping` or `RateLimited`, or if you're manually releasing jobs, you'll likely need to increase the number of allowed attempts.
```

queues.md
```diff
@@ -1213,7 +1215,22 @@ class ProcessPodcast implements ShouldQueue
...
+
+However, an attempt does not necessarily mean the job’s handle method was executed. Attempts can be consumed in several ways:
+
+- The job's `handle` method runs and completes without throwing an exception.
+- The job encounters an unhandled exception during execution.
+- The job timed out.
+- The job is manually released back to the queue using `$this->release()`.
+- Middleware such as `WithoutOverlapping` or `RateLimited` fails to acquire a lock and releases the job.
+
+You likely do not want it to keep retrying indefinitely a job. Therefore, Laravel provides various ways to specify how many times or for how long a job may be attempted.
```
browner12, line 1230: grammar:

```suggestion
You likely do not want it to keep retrying a job indefinitely. Therefore, Laravel provides various ways to specify how many times or for how long a job may be attempted.
```

horizon.md
```diff
@@ -126,37 +150,231 @@ Within Horizon's default configuration file, you will notice a `defaults` config
...
+],
+```
+
+> [!NOTE]
+> This option is similar to the `--tries` option when using the Artisan command to process queues.
+
+Fine-tuning the `tries` option is essential when using middlewares such as `WithoutOverlapping` or `RateLimited` because they consume attempts. However, depending on your case, it may be more appropriate to set the job class' `$tries` property.
+
+If you do not specify a value for the `tries` option, jobs will only be attempted once or as many times as specified by the job class `$tries` property.
+
```
stevebauman, line 330:
```suggestion

If `tries` (or `$tries` on the job class) is set to 0, the job may be attempted (and released) an unlimited number of times. This can be helpful in handling circumstances when the number attempts may be unknown.
```

Not sure if we want to recommend setting `$maxExceptions` here to support setting tries to 0? Thoughts?

queues.md
```diff
@@ -1213,7 +1215,22 @@ class ProcessPodcast implements ShouldQueue
...
+
+However, an attempt does not necessarily mean the job's handle method was executed. Attempts can be consumed in several ways:
+
+- The job's `handle` method runs and completes without throwing an exception.
+- The job encounters an unhandled exception during execution.
+- The job timed out.
+- The job is manually released back to the queue using `$this->release()`.
+- Middleware such as `WithoutOverlapping` or `RateLimited` fails to acquire a lock and releases the job.
+
+You likely do not want it to keep retrying a job indefinitely. Therefore, Laravel provides various ways to specify how many times or for how long a job may be attempted.
```
stevebauman, line 1230: It might be worth adjusting this wording to `attempting` instead of `retrying`? Thoughts?

```suggestion
You likely do not want to keep attempting a job indefinitely. Therefore, Laravel provides various ways to specify a limit.
```

queues.md
```diff
@@ -1213,7 +1215,22 @@ class ProcessPodcast implements ShouldQueue
...
+
+> [!NOTE]
+> By default, Laravel will only attempt a job once. If your job uses middleware like `WithoutOverlapping` or `RateLimited`, or if you're manually releasing jobs, you'll likely need to increase the number of allowed attempts.
```
stevebauman, line 1233:
```suggestion
> By default, Laravel will only attempt a job once. If your job uses middleware like `WithoutOverlapping` or `RateLimited`, or if you're manually releasing jobs, you'll likely need to increase the number of allowed attempts via the `tries` option.
```

queues.md
```diff
@@ -2249,6 +2273,18 @@ class ProcessPodcast implements ShouldQueue
 > [!WARNING]
 > A new instance of the job is instantiated before invoking the `failed` method; therefore, any class property modifications that may have occurred within the `handle` method will be lost.
 
+A failed job is not necessarily one that encountered an unhandled exception. A job is considered failed when it has exhausted all of its allowed attempts. These attempts can be consumed in several ways:
```
stevebauman, line 2276:
```suggestion
A failed job is not necessarily one that encountered an unhandled exception. A job may also be considered failed when it has exhausted all of its allowed attempts. These attempts can be consumed in several ways:
```

queues.md
```diff
@@ -2249,6 +2273,18 @@ class ProcessPodcast implements ShouldQueue
...
+
+- The job encounters an unhandled exception during execution.
+- The job timed out.
```
stevebauman, line 2279: Super nit: sort in ascending for prettiness:

```suggestion
- The job timed out.
- The job encounters an unhandled exception during execution.
```

queues.md
```diff
@@ -1213,7 +1215,22 @@ class ProcessPodcast implements ShouldQueue
...
+Job attempts are a core concept of Laravel's queue system and power many advanced features. While they may seem confusing at first, it's important to understand how they work before modifying the default configuration.
+
+When a job is dispatched, it is pushed onto the queue. A worker then picks it up and attempts to execute it. This is known as a job attempt.
+
+However, an attempt does not necessarily mean the job's handle method was executed. Attempts can be consumed in several ways:
+
+- The job's `handle` method runs and completes without throwing an exception.
+- The job encounters an unhandled exception during execution.
+- The job timed out.
+- The job is manually released back to the queue using `$this->release()`.
```
stevebauman, line 1227: Super nit: sort in ascending for prettiness:

```suggestion
- The job timed out.
- The job encounters an unhandled exception during execution.
- The job is manually released back to the queue using `$this->release()`.
- The job's `handle` method runs and completes without throwing an exception.
```

horizon.md
```diff
@@ -126,37 +150,231 @@ Within Horizon's default configuration file, you will notice a `defaults` config
...
+        ],
+    ],
+],
+```
+
+> [!NOTE]
+> This option is similar to the `--tries` option when using the Artisan command to process queues.
+
+Fine-tuning the `tries` option is essential when using middlewares such as `WithoutOverlapping` or `RateLimited` because they consume attempts. However, depending on your case, it may be more appropriate to set the job class' `$tries` property.
+
```
stevebauman, line 328:
```suggestion

If `$tries` is defined on the job class, then it will take precedence over the `tries` provided in the Horizon configuration.

```

Horizon uses the framework `WorkCommand`:

https://github.com/laravel/horizon/blob/2744ca84896012d2de135ea0cbb50a8da26d3827/src/Console/WorkCommand.php#L7

Which passes in `WorkerOptions`:

https://github.com/laravel/framework/blob/c82b43b757c608c0afc1b549aa802b12cc847f7c/src/Illuminate/Queue/Worker.php#L425

And only uses the `WorkerOptions::$maxTries` if `null` is returned from the `maxTries()` method on the job (which returns its `$tries` property if its set):

https://github.com/laravel/framework/blob/c82b43b757c608c0afc1b549aa802b12cc847f7c/src/Illuminate/Queue/Worker.php#L433-L435

https://github.com/laravel/framework/blob/c82b43b757c608c0afc1b549aa802b12cc847f7c/src/Illuminate/Queue/Worker.php#L516-L518

stevebauman (commented):
Nice work @QuentinGab! Apologies it took me a bit to get to this. This is a great improvement. Feel free to disregard my suggestions if you don't like them.

horizon.md
QuentinGab: You're right,  I'll try to reformulate this section to add context about `$maxExceptions` and `tries=0` while keeping it concise

queues.md
QuentinGab: Yes. I've adjusted the section accordingly!

horizon.md
QuentinGab: I tried to work this into the other sentences to keep it concise.

QuentinGab:
Thanks a lot for the suggestions, @stevebauman! I tried to integrate them as smoothly as possible. Let me know if you think I should make any other changes.

stevebauman (commented):
Excellent! Let's see what the Laravel team thinks. These new docs should help developers get clarification on some mysteries that would previously take diving into the source to figure out. Thanks for tackling this @QuentinGab ❤️ 

This all looks great to me! 👍

taylorotwell:
Thanks all!

siarheipashkevich:
@QuentinGab thanks a lot.

I have one question regarding failed queue jobs. My app is using Sentry which included globally. If the queue job has max_attempts is 3 and will throw an exception from `handle` method. Will Sentry catch these exceptions or not? Or I should manually catch any exceptions inside the try/catch section and re-throw exception to the next for handling max attempts logic?

QuentinGab:
> @QuentinGab thanks a lot.
> 
> I have one question regarding failed queue jobs. My app is using Sentry which included globally. If the queue job has max_attempts is 3 and will throw an exception from `handle` method. Will Sentry catch these exceptions or not? Or I should manually catch any exceptions inside the try/catch section and re-throw exception to the next for handling max attempts logic?


Hi @siarheipashkevich 

Yes, the error will be reported to Sentry (or Nightwatch). Here is an full example with Nightwatch:

Capture d’écran 2025-08-13 à 14 05 31

PR #10619, opened 2025-07-16
[12.x] Improve explanation for whereUuid and whereUlid route constraints

gaurav-mehtawebsolution:
**Why this is helpful**

Right now, the docs only include quick examples for these methods, but they don’t actually explain:
- What a UUID or ULID is
- What they look like
- Why you’d want to use these constraints in your routes

For someone new to Laravel (or to UUIDs/ULIDs in general), this can be confusing. So I thought adding a short, clear description with examples would make things much clearer.

**What I’ve added**
A simple explanation of what UUIDs and ULIDs are

- An example of what each format looks like
- A quick note about when and why you’d use these constraints
- Kept the style consistent with the rest of the docs

I hope this small improvement will help beginners understand route parameter constraints a bit better and save them some time Googling!

Thanks so much for reviewing — happy to tweak anything if needed.

### DIFF
--- a/routing.md
+++ b/routing.md
@@ -1,39 +1,40 @@
 # Routing
 
 - [Basic Routing](#basic-routing)
-    - [The Default Route Files](#the-default-route-files)
-    - [Redirect Routes](#redirect-routes)
-    - [View Routes](#view-routes)
-    - [Listing Your Routes](#listing-your-routes)
-    - [Routing Customization](#routing-customization)
+  - [The Default Route Files](#the-default-route-files)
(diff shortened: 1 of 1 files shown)


AhmedAlaa4611:
> Kept the style consistent with the rest of the docs

I believe this PR introduced some style inconsistencies:

- First, the indentation was changed from 4 spaces to 2, which doesn't align with the rest of the documentation.
- Second, a new line was added after the `<a></a>` tag, which also breaks the established formatting style.

Overall, these changes seem to conflict with the styling used elsewhere in the docs.

PR #10620, opened 2025-07-16
[10.x] Clarify migration step in Passport installation instructions

imlolman:
While installing Laravel Passport, I noticed that the documentation suggested running migrations manually before executing `passport:install`. However, the `passport:install` command now handles migrations automatically or prompts to run them if needed.

This update simplifies the instructions by removing the separate migration step and explaining that `passport:install` handles table creation along with encryption key and client generation. This avoids confusion and reflects the current behavior of the command more accurately.

### DIFF
--- a/passport.md
+++ b/passport.md
@@ -68,7 +68,7 @@ To get started, install Passport via the Composer package manager:
 composer require laravel/passport
 ```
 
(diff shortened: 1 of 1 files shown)


AhmedAlaa4611:
According to [https://laravel.com/docs/12.x/releases#support-policy](https://laravel.com/docs/12.x/releases#support-policy), 10.x is not officially supported anymore. I think any new PR should be added the 12.x branch.

PR #10621, opened 2025-07-16
[12.x] Improve formatting in flatten method documentation

AhmedAlaa4611:
Description
---
This PR improves the explanation and formatting in the `flatten()` method section of the `collections.md` file. Specifically:

- Reorders the explanation to introduce the depth parameter before showing the difference in output.
- Formats the example output in a code block for consistency with other parts of the docs.

This change improves readability and aligns the style with other examples in the docs.


taylorotwell:
I think this one is OK as is.

PR #10622, opened 2025-07-16
[12.x] Adds bool:strict docs

peterfox:
Follow up docs for this change in Laravel 12 https://github.com/laravel/framework/pull/56313

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -1346,7 +1346,7 @@ The field under validation must have a size between the given _min_ and _max_ (i
 
 The field under validation must be able to be cast as a boolean. Accepted input are `true`, `false`, `1`, `0`, `"1"`, and `"0"`.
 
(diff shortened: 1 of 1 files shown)


validation.md
```diff
@@ -1346,6 +1346,12 @@ The field under validation must have a size between the given _min_ and _max_ (i
 
 The field under validation must be able to be cast as a boolean. Accepted input are `true`, `false`, `1`, `0`, `"1"`, and `"0"`.
 
+You may use a strict parameter with the rule to validate the field as `true` or `false` exclusively.
```
AhmedAlaa4611, line 1349: I think we need to change `a` with `the`:
```diff
- You may use a strict parameter with the rule to validate the field as `true` or `false` exclusively.
+ You may use the strict parameter with the rule to validate the field as `true` or `false` exclusively.

PR #10625, opened 2025-07-18
Fix return-type typo

kburton-dev:
Not much description needed: Generator return type must be a supertype of Generator, not void.


browner12:
"typo" suggests a spelling error to me. I'd drop that word from your title.

PR #10628, opened 2025-07-19
[12.x] Reformat the link

AhmedAlaa4611:
Description
---
I think only `queued jobs` should be displayed as a link, not `just like queued jobs`.


taylorotwell:
I think this one is OK actually.

PR #10630, opened 2025-07-20
registering and using a custom notification channel

a-sabagh:
if anyone want to add custom notification channel using simple string identifier using manager pattern in channel manager class


taylorotwell:
Not necessary I think.

PR #10639, opened 2025-07-24
Explain cache prefix change when not default value

m4tthumphrey:
When not using an explicit cache prefix the default is used. Removing the prefix in v11 causes new keys to be created and old keys forgotten about.


taylorotwell:
I'm not sure what this adds that we don't already say above. 🤔

PR #10641, opened 2025-07-25
Possible confusion regarding Auth::logoutOtherDevices()

mortenscheel:
The current description seems to suggest the following about invalidating other sessions during password reset:
1. That it is optional.
2. That it should be done manually.

AFAIK none of those are true. After changing the password, the other sessions will be logged out automatically by `AuthenticateSession` because the password hash no longer matches.

I'm actually very curious why `AuthenticateSession` isn't included in the default `web` middleware group? The automatic invalidation seems like a feature you would want to have enabled by default. Is the `hash_equals()` check too expensive?

### DIFF
--- a/authentication.md
+++ b/authentication.md
@@ -508,7 +508,7 @@ public function logout(Request $request): RedirectResponse
 <a name="invalidating-sessions-on-other-devices"></a>
 ### Invalidating Sessions on Other Devices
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Not sure we need to change anything here to be honest. I think we probably could enable that middleware by default on our starter kits to be honest. 👍

PR #10644, opened 2025-07-26
[12.x] Fix grammar

AhmedAlaa4611:
Description
---
Grammarly suggests this correction for grammatical accuracy.

### DIFF
--- a/events.md
+++ b/events.md
@@ -460,7 +460,7 @@ class SendShipmentNotification implements ShouldQueueAfterCommit
 <a name="queued-listener-middleware"></a>
 ### Queued Listener Middleware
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
Original was correct actually.

shaedrich:
Oof, that's indeed a tough one 😂 

> The noun middleware can be countable or uncountable.
>
> In more general, commonly used, contexts, the plural form will also be [middleware](https://www.wordhippo.com/what-is/the-meaning-of-the-word/middleware.html).
>
> However, in more specific contexts, the plural form can also be [middlewares](https://www.wordhippo.com/what-is/the-meaning-of-the-word/middlewares.html) e.g. in reference to various types of middlewares or a collection of middlewares.
> – From [wordhippo.com](https://www.wordhippo.com/what-is/the-plural-of/middleware.html)

PR #10647, opened 2025-07-27
[12.x] docs: Eloquent - Changed wording on use of `destroy()`

vHeemstra:
The wording was such that it seemed like the `destroy` method would not load the models (which would be a great performance benefit).

However, this is not the case! (as a warning 2 paragraphs down mentions)

The benefit of `destroy` is that **_you_** don't have to load the models first before you can delete them.

So I changed the wording to make it less ambiguous and moved the warning to a more appropriate place.

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -1040,7 +1040,7 @@ $flight->delete();
 <a name="deleting-an-existing-model-by-its-primary-key"></a>
 #### Deleting an Existing Model by its Primary Key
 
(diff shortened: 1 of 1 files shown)


AhmedAlaa4611:
The current docs state:

> you may delete the model without explicitly retrieving it by calling the `destroy` method.

So, I think it is clear and fine as it is.

vHeemstra:
Well, "without explicitly retrieving it" could be read as "the system doesn't retrieve it".
Since that's not the case, why not be more precise and state that the user doesn't have to do this?
The wording is not that much longer, and it tackles the ambiguity. Win-win 👍

PR #10650, opened 2025-07-27
[12.x] Fix order of sections in packages docs

AhmedAlaa4611:
Description
---
This PR corrects the order of the "Migrations" and "Routes" sections in the packages docs. Previously, the main content displayed "Routes" before "Migrations", which did not match the order shown in the right sidebar (table of contents). The sections are now reordered to ensure consistency between the sidebar and the main documentation content.

Note
---
While the changes may not be immediately obvious, I only reversed the order of the two sections.

### DIFF
--- a/packages.md
+++ b/packages.md
@@ -136,21 +136,6 @@ public function register(): void
 > [!WARNING]
 > This method only merges the first level of the configuration array. If your users partially define a multi-dimensional configuration array, the missing options will not be merged.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think routes should come before migrations.

AhmedAlaa4611:
Solved in: https://github.com/laravel/docs/commit/e368be1298303d77e04e1dc7fe47975112615f59

PR #10651, opened 2025-07-27
Improve Stringable::whenIsUuid() example

shaedrich:
The old example
* wasn't very descriptive: It extracted the `time_low` part but didn't indicate this anywhere, let alone it looked like a pretty random choice
* didn't treat the UUID as such: It didn't properly parse it (we should strive to show best practices)

The new example
* actually does something with the passed UUID: It formats the extracted UUID segment as it would be expected as a standalone value


taylorotwell:

PR #10653, opened 2025-07-27
[12.x] Improve Artisan console docs

amirhshokri:
This PR improves the Artisan console documentation by:

- Correcting the usage of the `choice()` method parameters
- Adding the missing `alert()` method to the console output methods list

### DIFF
--- a/artisan.md
+++ b/artisan.md
@@ -651,15 +651,15 @@ $name = $this->choice(
     'What is your name?',
     ['Taylor', 'Dayle'],
     $defaultIndex,
-    $maxAttempts = null,
-    $allowMultipleSelections = false
+    attempts: null,
+    multiple: false
 );
 ```
 
 <a name="writing-output"></a>
 ### Writing Output
 
-To send output to the console, you may use the `line`, `info`, `comment`, `question`, `warn`, and `error` methods. Each of these methods will use appropriate ANSI colors for their purpose. For example, let's display some general information to the user. Typically, the `info` method will display in the console as green colored text:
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hmm I find those named args actually less intuitive. Maybe we should rename those arguments?

amirhshokri:
@taylorotwell Done.

artisan.md
```diff
@@ -651,15 +651,15 @@ $name = $this->choice(
     'What is your name?',
     ['Taylor', 'Dayle'],
     $defaultIndex,
-    $maxAttempts = null,
-    $allowMultipleSelections = false
+    $maxAttempts,
+    $allowMultipleSelections
```
AhmedAlaa4611, line 655: What was the reason for removing the default value here?

AhmedAlaa4611: They should be updated like this:

```php
$maxAttempts: null,
$allowMultipleSelections: false
```

amirhshokri: I believe the default values can already be understood from the function definition itself, and the main goal here is just to indicate that these parameters are available to be passed. Also, we can't assign values to arguments unless they're defined in the function signature.

The parameter names in the function definition are different, so if we want to use named arguments, they would have to be written like:

```php
attempts: null,
multiple: false
```

AhmedAlaa4611: I wasn't aware that the argument names are `attempts` and `multiple`. Thank you for clarifying that.

taylorotwell:
I think the current docs are pretty fine on this.

PR #10657, opened 2025-07-28
[12.x] add a `$` when referencing a class property

browner12:
consistently prefix with a `$` when referencing a class property.

this is a follow up to #10645

I explicitly only added the prefix when we were directly referencing and showing an example of a PHP class.  I did not add the prefix when referencing a javascript property of an object, or when the example showed the calling code of the property from an instantiated object (`$model->property`).

if I'm being honest, I think the previous PR went the wrong direction, and instead we should *remove* all the `$` prefixes. it feels like the `$` is implied for all PHP class properties. we also get this inconsistency, because it becomes largely context dependent on what *type* of property we're referencing, and _how_ we're showing it in the example. but I'll obviously leave that up to maintainers to decide on.

### DIFF
--- a/artisan.md
+++ b/artisan.md
@@ -122,7 +122,7 @@ php artisan make:command SendEmails
 <a name="command-structure"></a>
 ### Command Structure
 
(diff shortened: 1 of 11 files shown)


AhmedAlaa4611:
> it becomes largely context dependent on what type of property we're referencing, and how we're showing it in the example.

According to your description, I think we may need to remove the prefix.

browner12:
correct. Taylor can decide if he wants to accept this PR, which matches your changes from the previous PR. or if he'd like to revert and go the other direction.

taylorotwell:
Just reverted the other PR.

PR #10659, opened 2025-07-29
[12.x] clarify dispatch behavior when using beforeCommit() with ShouldQueueAfterCommit interface

ahmad-cit22:
This little update is added in support of the changes made in ([framework#56445](https://github.com/laravel/framework/pull/56445)). It will give developers more clarity about the behavior of job dispatching when using the `beforeCommit()` method alongside the `ShouldQueueAfterCommit` interface.

The will be helpful for devs to ensure more predictable and intentional dispatch timing.


taylorotwell:
I'm really not sure this is needed.

PR #10668, opened 2025-08-01
[12.x] Prevent line break after code-formatted language name and list separator

shaedrich:
| Before | After |
| ------ | ----- |
| grafik | grafik |


AhmedAlaa4611:
I think increasing the grid column span of the middle content section would help address a lot of issues like this. The current layout feels limiting, and this doesn't seem like a long-term solution.

shaedrich:
That should be done as well, but that wouldn't necessarily prevent issues like this.

PR #10683, opened 2025-08-06
[12.x] Improve docs for `Schedule::command` and `Artisan::call()`

amirhshokri:


taylorotwell:
I personally prefer to just leave this as undocumented. It's not the typical use case imo.

PR #10689, opened 2025-08-06
[12.x] Pipeline::withinTransaction() accepts connection

AhmedAlaa4611:
https://github.com/laravel/framework/pull/56550

---

Should we use: `For convenience, you may wish to ...` or `You may also provide ...`?


taylorotwell:
I don't think we have to document this since the type-hint should make it visible.

PR #10693, opened 2025-08-07
[12.x] Document events and message handling in Reverb

lakchii:


taylorotwell:
I think this could maybe be much more condensed. Also I think more modern to just use event discovery on listeners than EventServiceProvider.

PR #10694, opened 2025-08-07
Add note about S3 ACLs (12)

OmniTroid:
See also https://chrysanthos.xyz/article/how-to-fix-s3-error-the-bucket-does-not-allow-acls-in-laravel/

### DIFF
--- a/filesystem.md
+++ b/filesystem.md
@@ -221,6 +221,21 @@ Typically, after updating the disk's credentials to match the credentials of the
 'endpoint' => env('AWS_ENDPOINT', 'https://minio:9000'),
 ```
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Every Laravel installation in the world that uses S3 needs this? Why would it not just be the default in laravel/laravel then?

OmniTroid:
Because it depends on the configuration of the S3 bucket. Buckets created before April 5th will work fine without the 'options', but buckets created after will require it (assuming buckets are created with default configuration, which is typically safe to assume...), which makes it seem pertinent to at least document. Would it be better to deduce the situation in code rather than assuming it via config?

PR #10701, opened 2025-08-07
index.php calls handleRequest

khattaksd:
`index.php` calls `handleRequest` and *not* `handleCommand`.

### DIFF
--- a/lifecycle.md
+++ b/lifecycle.md
@@ -3,7 +3,7 @@
 - [Introduction](#introduction)
 - [Lifecycle Overview](#lifecycle-overview)
     - [First Steps](#first-steps)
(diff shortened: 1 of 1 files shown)


AhmedAlaa4611:
It is not about `index.php` only, it about `index.php` and also the `artisan` file which in the root of any Laravel application that handles the `php artisan` console commands.

This should be closed.

khattaksd:
The lifecycle step mentioned right before this statement does not mention 'artisan' command flow at all.
Clarification is required as to when and how the kernels are called.
That said, I am a big fan of Laravel and this is just an effort to contribute. 🙏🏾
Please read that section in full context.
Thank you!

PR #10705, opened 2025-08-07
[12.x] Improve batch jobs documentation

SebKay:
I was tripped up for a long time by this issue when running multiple queue workers for the same queue.

I first noticed it when using Laravel Cloud from playing around with the queue worker settings.

Then I happened to see a video by Harris Raftopoulos where he explained you can force a batch of jobs to run in parallel by putting them in a nested array inside the batch.

I've updated the documentation to make this clear, as it currently isn't.

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -1533,7 +1533,9 @@ class ImportCsv implements ShouldQueue
 <a name="dispatching-batches"></a>
 ### Dispatching Batches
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hmm, I'm not sure this is actually an improvement. The fact the jobs get processed concurrently is the defining features of job batches. 🤔 I'm not sure it needs a 'warning'.

SebKay:
OK, fair point.

In that case, does it make sense to use a CSV import as the example?

Surely you'd want your CSV rows to import in to the database in the same order they appear in the CSV file? I think it would be confusing if they imported in a random order depending on how you queue workers are set up.

Maybe it's best to remove the warning and just have it as plain text?

PR #10714, opened 2025-08-09
[12.x] The database driver is the default queue driver

AhmedAlaa4611:
Description
---
According to the [.env](https://github.com/laravel/laravel/blob/8d8d3b729afacce5632ca8b42e20e09f5433d2d6/.env.example#L38)  and [config/queue.php](https://github.com/laravel/laravel/blob/8d8d3b729afacce5632ca8b42e20e09f5433d2d6/config/queue.php#L16) files, the default queue driver is the `database` driver.


AhmedAlaa4611:
Is it still necessary to mention that the `sync` driver executes jobs synchronously in the foreground?

PR #10720, opened 2025-08-10
[12.x] Add new methods docs for `ComponentAttributeBag`

KennedyTedesco:
This is a follow-up work of: https://github.com/laravel/framework/pull/56452


taylorotwell:
We may need more reliastic examples here.

KennedyTedesco:
@taylorotwell My main idea was not to give full examples, as the methods themselves are self-explanatory.

But I’ll come up with another version in the next few days when I have some time.

PR #10728, opened 2025-08-12
Typo fix

zigzagdev:
The indefinite article was incorrect (“an” had been written as “a”), so I’ve fixed it.

``` Before
Laravel includes a `Illuminate\Queue\Middleware\ThrottlesExceptions`
```

``` After
Laravel includes an `Illuminate\Queue\Middleware\ThrottlesExceptions`
```

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -688,7 +688,7 @@ class ProviderIsUp
 <a name="throttling-exceptions"></a>
 ### Throttling Exceptions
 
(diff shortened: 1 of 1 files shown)


AhmedAlaa4611:
You will find this in other parts of the docs, see: https://github.com/search?q=repo%3Alaravel%2Fdocs%20%22a%20%60Illuminate%22&type=code

shaedrich:
Debating natural speech grammar in code is kind of philosophical terrain as code is not to be spoken, not even to be used as prose. Similar to sorting imports, different people will probably give you different answers (sort by whole FQN vs sort by class base name).

PR #10733, opened 2025-08-13
Remove duplicate event:list mention from events docs

alsayeedar:
The event:list command is mentioned twice in the Laravel 12 events documentation — once at the end of the Event Discovery section and again immediately after the Manually Registering Events section. Both instances contain identical text and code. This PR removes the second occurrence to avoid redundancy and keep the documentation concise.

### DIFF
--- a/events.md
+++ b/events.md
@@ -134,12 +134,6 @@ public function boot(): void
 }
 ```
 
-The `event:list` command may be used to list all of the listeners registered within your application:
-
-```shell
-php artisan event:list
-```
-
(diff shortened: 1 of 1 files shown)


AhmedAlaa4611:
This PR should be closed because it indicates that, whether you are using **Event Discovery** or **Manually Registering Events**, you can list your events using the `event:list` Artisan command.

taylorotwell:
Documentation doesn't have to follow DRY. We can repeat things if it helps the user.

PR #10741, opened 2025-08-17
[12.x] Add resolve singletons with parameters

AhmedAlaa4611:
See
---
https://github.com/laravel/framework/discussions/56672


AhmedAlaa4611:
Should we phrase this as "the cached singleton" or would "the same object instance" be more accurate?

taylorotwell:
This should probably be given along with an example.

PR #10746, opened 2025-08-18
[12.x] Use Str::plural

AhmedAlaa4611:
Description
---
This fixes the messages so that we don't have messages that say "1 seconds" instead it will say "1 second".


taylorotwell:
I think this is OK for now.

PR #10747, opened 2025-08-18
[12.x] Add resolve singletons with parameters

AhmedAlaa4611:
Rework of: #10741

---

Should we phrase this as "the cached singleton" or would "the same object instance" be more accurate?


taylorotwell:
I think we can leave this one.

PR #10749, opened 2025-08-19
[12.x] Remove outdated phrasing

AhmedAlaa4611:
Description
---
The current phrasing made sense when **Horizon** was first introduced, but since Horizon has been part of Laravel for years, it no longer feels current. This makes the feature stable and keeps the tone clear and timeless.


shaedrich:
Besides, "beautiful" is quite biased 😂

I thought, this was a documentation not an advertisement ^^

PR #10753, opened 2025-08-20
[12.x] Add Limit::perHour to route rate limiters

AhmedAlaa4611:
Description
---
Most of the examples in the rate limiting documentation demonstrate the `Limit::perMinute` method. To add more variety and show that other intervals are also available, I updated **only** the final example to use `Limit::perHour` instead of `Limit::perMinute`.


taylorotwell:
I'm not sure that specific example makes sense though since the daily limit is so high.

PR #10771, opened 2025-08-24
[12.x] Add assertNotReleased method

AhmedAlaa4611:
See
---
https://github.com/laravel/framework/blob/b8e2aa2527fb5601698e8021b3b205632ef2e4ba/src/Illuminate/Queue/InteractsWithQueue.php#L241-L256


taylorotwell:
Not sure it's really needed.

PR #10774, opened 2025-08-25
[12.x] Clean up redundant wording

AhmedAlaa4611:
Before
---
The docs explanation about Swoole feels a bit repetitive.

After
---
The text now introduces the concept generally and leaves the example to demonstrate how to namespace the function. This keeps the section cleaner and easier to follow.


taylorotwell:
I think it's OK.

PR #10782, opened 2025-08-28
[12.x] Enhance trimming section with 'keep' key details

jonatasbaldan:
Context: I needed to extend the pulse trimming interval but I didn't find much information on the web. And the value is too implicit (at least for me), which made me consulting the underlying implementation.

In the [pulse/src/Storage/DatabaseStorage.php](https://github.com/laravel/pulse/blob/1.x/src/Storage/DatabaseStorage.php) on line `130`, the interval is created using the `CarbonInterval::fromString`.

And in "...which is also the minimum value" part is confirmed in the line `134`, when it checks if the `before` var is less than `7 days`, if it is the `before` is set to `7 days`.

```
    /**
     * Trim the storage.
     */
    public function trim(): void
    {
        $now = CarbonImmutable::now();

        $keep = $this->config->get('pulse.storage.trim.keep') ?? '7 days';

        $before = $now->subMilliseconds(
            (int) CarbonInterval::fromString($keep)->totalMilliseconds
        );

        if ($now->subDays(7)->isAfter($before)) {
            $before = $now->subDays(7);
        }
        ....
```

English is not my native language, so I'd appreciate it if anyone could help me with my grammar. :D

### DIFF
--- a/pulse.md
+++ b/pulse.md
@@ -521,6 +521,22 @@ In general, the more entries you have for a particular metric, the lower you can
 
 Pulse will automatically trim its stored entries once they are outside of the dashboard window. Trimming occurs when ingesting data using a lottery system which may be customized in the Pulse [configuration file](#configuration).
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
The `keep` value doesn't change how often the trim function is run though. That works based on the lottery. 🤔

PR #10786, opened 2025-08-28
[12.x] Add toPrettyJson to Eloquent Serialization

AhmedAlaa4611:
See: https://github.com/laravel/framework/pull/56697

I used the same tone that we used here: #10765

Note
---
I didn't remove the example:

```php
return $user->toJson(JSON_PRETTY_PRINT);
```

Because it's a way for documenting things. Also, the current docs state:

https://github.com/laravel/docs/blob/90032baf5391094038f32f2428680ac4a0420ee3/eloquent-serialization.md?plain=1#L54C174-L54C307

### DIFF
--- a/eloquent-serialization.md
+++ b/eloquent-serialization.md
@@ -63,6 +63,16 @@ return $user->toJson();
 return $user->toJson(JSON_PRETTY_PRINT);
 ```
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think it's OK.

PR #10791, opened 2025-09-02
[12.x] Document new Plural param for prependCount

caseydwyer:
Doc updates related to https://github.com/laravel/framework/pull/56802

### DIFF
--- a/strings.md
+++ b/strings.md
@@ -1191,7 +1191,7 @@ $singular = Str::plural('child', 1);
 // child
 ```
 
-You may pass a third argument to prefix the pluralized string with the formatted `$count`, via the `prependCount` parameter:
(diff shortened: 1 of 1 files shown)


AhmedAlaa4611:
Hello @caseydwyer 

Just a small suggestion: instead of using **1234**, maybe **1000** would be a bit clearer and easier for readers to follow at a glance. It also avoids the impression that the number was chosen randomly. What do you think?

caseydwyer:
> Just a small suggestion: instead of using **1234**, maybe **1000** would be a bit clearer and easier for readers to follow at a glance. It also avoids the impression that the number was chosen randomly. What do you think?

Makes sense, thanks for the feedback @AhmedAlaa4611—updated!

PR #10793, opened 2025-09-03
Add warning about observer event timing with `ShouldHandleEventsAfterCommit` interface

stevebauman:
Closes https://github.com/laravel/framework/issues/56400


taylorotwell:
Let me think on this one. It does what it says on the box imo so it somewhat feels redundant to say it again.

PR #10795, opened 2025-09-03
[12.x] document default slot usage

browner12:
while we always have the option for `$slot` content to live anywhere in the component, we can also be explicit by using the unnamed `<x-slot>` tag. this is necessary when you want to pass attributes to your default slot.

this is kind of documentation for https://github.com/laravel/framework/pull/56883 but the ability to have a default slot actually already existed, you just needed to name it "slot".

`<x-slot:slot>`

### DIFF
--- a/blade.md
+++ b/blade.md
@@ -1268,6 +1268,22 @@ You may define the content of the named slot using the `x-slot` tag. Any content
 </x-alert>
 ```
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Probably just going to leave this undocumented for now.

browner12:
@taylorotwell I feel like showing people how to pass attributes to the default slot is a feature that would be useful.

PR #10797, opened 2025-09-03
[12.x] use consistent `blade` code fence

browner12:
similar code blocks use "blade" for their code fence. don't see anything unique on these that would necessitate "xml".


taylorotwell:
I sometimes use that if it rendered better than Blade for some reason.

shaedrich:
Shouldn't the blade syntax highlighting be fixed then?

PR #10800, opened 2025-09-04
docs: add Nginx config example for handling CORS preflight in Laravel

allanmcarvalho:
Added an example to the Nginx configuration in the documentation showing how to forward OPTIONS requests to Laravel.   This ensures that CORS preflight requests are handled by the application instead of Nginx:

    # If you want to handle cors in the application instead of Nginx
    if ($request_method = OPTIONS) {
        rewrite ^ /index.php last;
    }

### DIFF
--- a/deployment.md
+++ b/deployment.md
@@ -77,6 +77,11 @@ server {
 
     error_page 404 /index.php;
 
+    # If you want to handle cors in the application instead of Nginx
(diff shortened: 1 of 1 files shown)


taylorotwell:
Not sure this is totally necessary in our docs.

PR #10801, opened 2025-09-05
Update Enum Type Example Usage

jakegore:
The example usage of defining a column based off an enum gives the following error:

```
  Object of class Modules\Pack\Enums\MessageEmailStatus could not be converted to string

  at vendor/laravel/framework/src/Illuminate/Database/Grammar.php:235
    231▕         if (is_array($value)) {
    232▕             return implode(', ', array_map([$this, __FUNCTION__], $value));
    233▕         }
    234▕ 
  ➜ 235▕         return "'$value'";
    236▕     }
    237▕ 
    238▕     /**
    239▕      * Escapes a value for safe SQL embedding.
```

I've adjusted the documentation to demonstrate passing the value of the enum, which works as expected

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -686,12 +686,12 @@ The `enum` method creates a `ENUM` equivalent column with the given valid values
 $table->enum('difficulty', ['easy', 'hard']);
 ```
 
-Of course, you may use the `Enum::cases()` method instead of manually defining an array of allowed values:
+Alternatively, you can use an existing enum instead of manually defining an array of allowed values:
 
 ```php
 use App\Enums\Difficulty;
 
-$table->enum('difficulty', Difficulty::cases());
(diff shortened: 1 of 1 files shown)


shaedrich:
This is already done in

https://github.com/laravel/framework/blob/5a0715f939b6b4da078610b596ca0b3d7f37d58a/src/Illuminate/Database/Schema/Blueprint.php#L1098-L1110

That was added in laravel/framework#56565/ laravel/framework@b11b3c7/[v12.22.0](https://github.com/laravel/framework/releases/tag/v12.22.0)—which version are you on?

jakegore:
I didn't spot that, I'm using `12.8.1` so that explains why

It'd probably be worth adding a note into the docs to explain that this doesn't work for `12.x` versions before `12.22.0`

shaedrich:
That would be a problem with all features added in the current version. This can currently only be seen in the changelog and the migration guide 🤔

PR #10802, opened 2025-09-05
Shouldn't it be resend/resend-laravel instead of resend/resend-php?

eduardocruz:
I looked at the documentation on Resend ... and they use: composer require resend/resend-laravel

https://resend.com/docs/send-with-laravel/


taylorotwell:
No - we don't use the Laravel specific SDK for our driver, just the general PHP one. 👍

PR #10836, opened 2025-10-01
docs(passport): add guide for retrieving authenticated client

renoirtech:
Added a new subsection explaining how to retrieve the authenticated client instance  using `auth('api')->client()`. Includes example usage inside a controller and a note  clarifying that `auth()->user()` is always null when using the client credentials grant.


taylorotwell:
Will need to format this. Should probably use Auth facade.

PR #10840, opened 2025-10-03
fix(precognition): correctly set file data for avatar

SantosVilanculos:
This pull request refactors the `onChange` handler for the avatar file input to correctly capture and set the `File` object in the form data instead of the input's string `value`.


timacdonald:
Thanks! I've updated the example to match the Vue example.

PR #10841, opened 2025-10-03
fix small typos

sajadtorkamani:


shaedrich:
No, this is actually not a typo, since the a/an rule applies to the _pronunciation_ of the subsequent word, not the _spelling_

sajadtorkamani:
@shaedrich Ah fair enough, I wasn't aware of that. I still think it looks like a typo even if not technically so but please feel free to close 🙏

shaedrich:
I can't close since I'm not a maintainer. But yeah, to a non-native tongue—myself included—this is indeed odd at first glance, so you are probably in good company 😉

shaedrich:
However, the missing space, you corrected, was a good catch and indeed missing 👍🏻

PR #10845, opened 2025-10-06
[12.x] Add `afterCommit` & `afterRollback` method to Database

milwad-dev:


taylorotwell:
I think these are mainly helpful for the framework itself.

PR #11081, opened 2026-03-05
[12.x] Document trait boot and initialize methods

rodvilla:
Add documentation for the `boot{TraitName}` and `initialize{TraitName}` conventions that allow Eloquent traits to hook into model lifecycle. These patterns are used by Laravel's own traits (HasUuids, SoftDeletes, etc.) but were not previously documented.

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -36,6 +36,7 @@
 - [Comparing Models](#comparing-models)
 - [Events](#events)
     - [Using Closures](#events-using-closures)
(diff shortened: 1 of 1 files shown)


rodvilla:
Similar documentation has been suggested on #10903 and #10907 but not accepted. Still, I think these are valuable methods that could benefit from a bit more exposure.

taylorotwell:
I don't think most apps needs this it's more useful for packages and internal stuff.

PR #11111, opened 2026-03-19
Question: what is the exact support period of Laravel 13?

Vectorial1024:
This PR aims to ask a question (should be an Issue).

Laravel 13 was released some hours ago, but the support period of Laravel 13 is unspecified. This PR aims to motivate discussion and clarification, and then subsequently be closed.

### DIFF
--- a/releases.md
+++ b/releases.md
@@ -28,7 +28,7 @@ For all Laravel releases, bug fixes are provided for 18 months and security fixe
 | 10      | 8.1 - 8.3 | February 14th, 2023 | August 6th, 2024    | February 4th, 2025   |
 | 11      | 8.2 - 8.4 | March 12th, 2024    | September 3rd, 2025 | March 12th, 2026     |
 | 12      | 8.2 - 8.5 | February 24th, 2025 | August 13th, 2026   | February 24th, 2027  |
-| 13      | 8.3 - 8.5 | Q1 2026             | Q3 2027             | Q1 2028              |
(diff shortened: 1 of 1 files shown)


SanderMuller:
See https://laravel.com/docs/13.x/releases#support-policy

Vectorial1024:
OK, I think I failed to convey it properly. I know that page, I was expecting exact dates instead of vague date ranges (Released "Q1 2026" but apparently Laravel 13 was released on 18 March 2026?). This could be clarified, but because I am not part of the internal Laravel team, I do not know the exact dates. I could only ask around here.

Or is it that the dates for the latest release have been inexact for many Laravel releases, and this is the tradition/norm?

SanderMuller:
Laravel went to yearly major releases in Q1 of every year. And yes, L13 was released on 17 March 2026, see https://github.com/laravel/framework/releases/tag/v13.0.0

Not sure about exact end of support dates tbh. Upgrades are quite easy, though! What are you mainly worried about?

Vectorial1024:
I have no technical concerns at all (the Laravel team is going great to keep user-facing changes at a minimum). This is purely a documentation concern. For example, PHP itself can clearly state their support period well in advance (see https://www.php.net/supported-versions.php ; they can mention e.g. PHP 8.5 will get security fixes until Dec 2029). I find it odd that Laravel, having a similar support policy as PHP, just doesn't have clear support periods.

But anyway I can just make a PR for the release date later. I hope I motivated some discussion in the internal Laravel team on whether they need to provide/clarify clear support periods.

PR #11124, opened 2026-03-23
Document --discover option for boost:update command

MrPunyapal:
Added information about the --discover option for boost:update command.

### DIFF
--- a/boost.md
+++ b/boost.md
@@ -119,8 +119,6 @@ By default, the `boost:update` command will only update the existing Boost resou
 php artisan boost:update --discover
 ```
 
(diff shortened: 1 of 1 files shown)


boost.md
```diff
@@ -113,6 +113,12 @@ You may also automate this process by adding it to your Composer "post-update-cm
 }
 ```
 
+By default, the `boost:update` command will only update the existing Boost resources already published within your application. If you would like Boost to scan your application for any newly installed packages and offer to publish their corresponding guidelines and skills, you may use the `--discover` option:
```
pushpak1300, line 116: newly installed packages -> newly installed third party packages

@MrPunyapal

PR #11130, opened 2026-03-27
fix wrong method call

zaccaro1980:


zaccaro1980:
hi @taylorotwell, why did you closed it?
class Symfony\Component\HttpFoundation\Response doesn't have a "status" method

PR #11139, opened 2026-04-02
[13.x] Document client credentials token ID collision risk

pushpak1300:
When using the `client_credentials` grant with integer client IDs, the token's `sub` claim can match a user's primary key, causing unintended user resolution. This adds a warning to the Passport docs explaining the risk and how to avoid it.

CC @hafezdivandari

### DIFF
--- a/passport.md
+++ b/passport.md
@@ -991,6 +991,9 @@ $response = Http::asForm()->post('https://passport-app.test/oauth/token', [
 return $response->json()['access_token'];
 ```
 
+> [!WARNING]
+> The underlying OAuth2 server sets the token's `sub` claim to the client's identifier for client credentials tokens. By default, Passport uses UUIDs for clients, so this cannot collide with a user's integer primary key. However, if you have set `Passport::$clientUuids` to `false`, a client credentials token may inadvertently resolve a user whose ID matches the client's ID. Applications using integer client IDs should ensure client credentials routes are protected by the `EnsureClientIsResourceOwner` middleware and are not shared with user-authenticated routes.
+
 <a name="personal-access-tokens"></a>
 ## Personal Access Tokens
 
--- a/passport.md
+++ b/passport.md
@@ -974,7 +974,7 @@ Route::get('/orders', function (Request $request) {
 ```
 
 > [!WARNING]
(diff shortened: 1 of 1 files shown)


hafezdivandari:
1. Would you please move it before the [Retrieving Tokens](https://laravel.com/docs/13.x/passport#retrieving-tokens) section? right after where the middleware is documented.
2. I think it is better to target `12.x` branch instead.
3. Also please add a link to OAuth2 server docs: https://oauth2.thephpleague.com/database-setup/#:~:text=Please%20note%20that,the%20bearer%20token.

Related to https://github.com/thephpleague/oauth2-server/issues/1456#issuecomment-2734989996 and https://github.com/laravel/passport/pull/1902#discussion_r3026389975

passport.md
```diff
@@ -973,6 +973,9 @@ Route::get('/orders', function (Request $request) {
 })->middleware(EnsureClientIsResourceOwner::using('servers:read', 'servers:create'));
 ```
 
+> [!WARNING]
+> The [underlying OAuth2 server](https://oauth2.thephpleague.com/database-setup/#:~:text=Please%20note%20that,the%20bearer%20token.) sets the token's `sub` claim to the client's identifier for client credentials tokens. By default, Passport uses UUIDs for clients, so this cannot collide with a user's integer primary key. However, if you have set `Passport::$clientUuids` to `false`, a client credentials token may inadvertently resolve a user whose ID matches the client's ID. Applications using integer client IDs should ensure client credentials routes are protected by the `EnsureClientIsResourceOwner` middleware and are not shared with user-authenticated routes.
```
hafezdivandari:
```suggestion
> The [underlying OAuth2 server](https://oauth2.thephpleague.com/database-setup/#:~:text=Please%20note%20that,the%20bearer%20token.) sets the token's `sub` claim to the client's identifier for client credentials tokens. By default, Passport uses UUIDs for clients, so this cannot collide with a user's integer primary key. However, if you have set `Passport::$clientUuids` to `false`, a client credentials token may inadvertently resolve a user whose ID matches the client's ID. In such cases, using this middleware cannot guarantee that the incoming token is a client credentials token.
```

PR #11144, opened 2026-04-06
Add link to Prism documentation for agent SSE output

WoutervdBrink:
After using `$agent->stream($prompt)` to stream the agent's response to the browser, I was a bit lost on where I could find more information on what exactly is being sent and how I could process it.

This PR adds a link to the relevant page of the Prism documentation.

Of course, I'm open to other ideas on how we could word this in the documentation.

### DIFF
--- a/ai-sdk.md
+++ b/ai-sdk.md
@@ -410,7 +410,7 @@ $response = (new ImageAnalyzer)->prompt(
 <a name="streaming"></a>
 ### Streaming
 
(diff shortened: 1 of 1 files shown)


WoutervdBrink:
Update: no, apparently the `laravel/ai` package adds its own [translation layer](https://github.com/laravel/ai/tree/0.x/src/Streaming/Events) and the name of events changes, e.g. `thinking_delta` becomes `reasoning_delta`.

Still quite confusing, but this PR does not improve the docs.

PR #11147, opened 2026-04-08
[13.x] Clarify orWhere array behavior when passing conditions as an array

NikhiltGhalme:
This documents the behavior of `orWhere` when passing an array of conditions. Currently, the conditions within the array are joined using `or` (not `and`), which can be unexpected.

This was introduced in laravel/framework#53147 and has been reported in laravel/framework#59516 and laravel/framework#53184.

As suggested by @crynobone in laravel/framework#59516, this PR adds a warning note to the query builder documentation clarifying the behavior and recommending the closure syntax when `and` is needed within the group.


taylorotwell:
I don't know this warrants a big red warning.

PR #11156, opened 2026-04-12
Document refreshApplication method for multi-request tests

restuff:
Added documentation for the `refreshApplication` method in the HTTP Testing section. While single-request tests are preferred, certain scenarios like parity testing between legacy and new endpoints require multiple requests. The `refreshApplication` method allows resetting the service container and application state to prevent state leakage between these requests.

Includes examples for both `Pest` and `PHPUnit`.


taylorotwell:
I really have never had to use this when building any Laravel applications.

PR #11160, opened 2026-04-15
[13.x] Adjust queue route docs

jackbayliss:
This makes it sound like `videos` will be the queue default, but it actually becomes the default for BOTH queue and connection. 

Sorry, you have to keep seeing my face!

### DIFF
--- a/queues.md
+++ b/queues.md
@@ -1307,7 +1307,7 @@ You may also route multiple job classes at once by passing an array to the `rout
 
 ```php
 Queue::route([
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hmm, I find that unexpected to be honest. Are you sure it works that way? If so that seems almost broken to me.

jackbayliss:
@taylorotwell Yeah, I fell into it today.  I'll think about it and prep a PR, unless you beat me too it and close this for now 👍🏻

PR #11167, opened 2026-04-18
Add missing documentation for api key that can be use in .env file

daviddprtma:
I add missing api key that can be used in .env file like:
- AZURE_OPENAI_API_KEY
- DEEPSEEK_API_KEY
- GROQ_API_KEY
- OPENROUTER_API_KEY


vwinck-dev (commented):
Yep, looks right

PR #11180, opened 2026-05-02
Clarify YEAR column limitations in migrations.md

mariopro:
Added a warning about the range limitation of the YEAR column in MySQL. This warning is important to clarify the **limitations** of this type. In many cases, especially in historical projects or projects dealing with **dates before 1901**, for instance, a company established since 1828, the `integer` type should be used.
In these cases, instead of ```$table->year('founded');``` use `$table->unsignedInteger('founded');` to keep your data coherent.

### DIFF
--- a/migrations.md
+++ b/migrations.md
@@ -1199,6 +1199,8 @@ The `year` method creates a `YEAR` equivalent column:
 ```php
 $table->year('birth_year');
 ```
+> [!WARNING]
+> The year method creates the `YEAR` equivalent column, but **limited in range**. YEAR accepts input values in a variety of formats, but always in the range '1901' to '2155'. Years previous to '1901' are not accepted by MySQL.
 
 <a name="column-modifiers"></a>
 ### Column Modifiers
--- a/migrations.md
+++ b/migrations.md
@@ -1200,7 +1200,7 @@ The `year` method creates a `YEAR` equivalent column:
 $table->year('birth_year');
 ```
 > [!WARNING]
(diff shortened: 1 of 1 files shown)


migrations.md
```diff
@@ -1199,6 +1199,8 @@ The `year` method creates a `YEAR` equivalent column:
 ```php
 $table->year('birth_year');
 ```
+> [!WARNING]
+> The year method creates the `YEAR` equivalent column, but **limited in range**. YEAR accepts input values in a variety of formats, but always in the range '1901' to '2155'. Years previous to '1901' are not accepted by MySQL.
```
shaedrich: I would put "MySQL" in front rather than somewhere in the back where it can easily be missed. People will keep assuming, it applies to all database engines until they reach the very last word.

```suggestion
> **When using the MySQL engine:** The year method creates the `YEAR` equivalent column, but **limited in range**. YEAR accepts input values in a variety of formats, but always in the range '1901' to '2155'. Years previous to '1901' are not accepted.
```

mariopro: Absolutely agree with your comment. For me, it’s a go and an important note because I spent 2 hours on this issue while migrating a project.

shaedrich: Yeah, not very intuitive. While I get that 1901 does have some meaning when working with dates, it makes especially the `year` column type way less useful

taylorotwell:
I try not to recreate MySQL specific documentation.

PR #11184, opened 2026-05-04
Clarify skipping validation rules for empty string

amyavari:
[Implicit Rules](https://laravel.com/docs/13.x/validation#implicit-rules) explanation says that empty string will skip the validation rules and the example shows it as well.

But, this case is valid when the user uses `Validator::make()`. If the user uses rules in the **FormRequest**, the middleware will convert empty string to `null`, and then the validation rules won't be skipped, and custom rules, even without `--implicit`, will be run. So, to clarify this to the users, in this PR I added a small note about this.

### DIFF
--- a/validation.md
+++ b/validation.md
@@ -3113,3 +3113,6 @@ php artisan make:rule Uppercase --implicit
 
 > [!WARNING]
 > An "implicit" rule only _implies_ that the attribute is required. Whether it actually invalidates a missing or empty attribute is up to you.
+
+> [!NOTE]
+> As mentioned in [a note on optional fields](#a-note-on-optional-fields), when using [Form Request Validation](#form-request-validation), empty strings are converted to `null` by the `ConvertEmptyStringsToNull` middleware. Therefore, they will be passed to the validation rules as `null` , and the rules will not be skipped.
--- a/validation.md
+++ b/validation.md
@@ -3115,4 +3115,4 @@ php artisan make:rule Uppercase --implicit
 > An "implicit" rule only _implies_ that the attribute is required. Whether it actually invalidates a missing or empty attribute is up to you.
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


validation.md
```diff
@@ -3113,3 +3113,6 @@ php artisan make:rule Uppercase --implicit
 
 > [!WARNING]
 > An "implicit" rule only _implies_ that the attribute is required. Whether it actually invalidates a missing or empty attribute is up to you.
+
+> [!NOTE]
+> As mentioned in [a note on optional fields](#a-note-on-optional-fields), when using [Form Request Validation](#form-request-validation), empty strings are converted to `null` by the `ConvertEmptyStringsToNull` middleware. Therefore, they will be passed to the validation rules as `null` , and the rules will not be skipped.
```
shaedrich: Assuming, the middleware is used like in the default setting

```suggestion
> As mentioned in [a note on optional fields](#a-note-on-optional-fields), when using [Form Request Validation](#form-request-validation), empty strings are converted to `null` by the `ConvertEmptyStringsToNull` middleware if it hasn't been removed from the route or its group. Therefore, they will be passed to the validation rules as `null` , and the rules will not be skipped.
```

shaedrich: Small typo:

```suggestion
> As mentioned in [a note on optional fields](#a-note-on-optional-fields), when using [Form Request Validation](#form-request-validation), empty strings are converted to `null` by the `ConvertEmptyStringsToNull` middleware. Therefore, they will be passed to the validation rules as `null`, and the rules will not be skipped.
```

amyavari: Good point.

Since the middleware is mentioned explicitly, I think leaving this point out would make the note a bit simpler. But I’m happy to include that clarification if you think it improves the accuracy of the docs.

taylorotwell:

PR #11190, opened 2026-05-06
[13.x] Bug fix branch latest is 13.x

legodlygod:
https://github.com/laravel/docs/pull/11187 changed this to 12.x which has confused me...

The text says: "bug fixes should be sent to the latest version that supports bug fixes"

image

The latest version to support bug fixes is 13.x, which is until 2027.

12.x is not the latest... so fixes go to 13.x, which can get backported afaik.

Am I missing something crazy here? or perhaps it needs rewording..


taylorotwell:
12 - Bug fixes until Aug 13th 2026.

PR #11200, opened 2026-05-14
docs: document socket ID retrieval for broadcast()->toOthers() with framework adapters

hectorgrecco:
## Summary

Extends the existing "Only To Others - Configuration" section to document how to retrieve the socket ID when using the framework adapter packages (`@laravel/echo-react`, `@laravel/echo-vue`, `@laravel/echo-svelte`).

Previously, this section only showed the vanilla Echo approach (`Echo.socketId()`), leaving framework adapter users without guidance — which led to confusion and workarounds (see laravel/echo#474).

### What's added

- `echo().socketId()` via the exported `echo()` helper, for interceptor setup and outside-component usage
- `useSocketId()` reactive hook/composable/rune for inside-component usage, with examples for all three adapters

## Related

- laravel/echo#474
- laravel/echo#499 (adds the `useSocketId()` hook itself)


hectorgrecco:
Closing — reopening against 13.x branch.

PR #11210, opened 2026-05-20
Document remember me functionality for Sanctum SPA authentication

LucasCavalheri:
This PR documents how Laravel's remember me functionality works when using Sanctum's SPA authentication mode.

It clarifies that SPA authentication uses Laravel's session authentication services, so `Auth::attempt($credentials, true)` may be used to issue a remember me cookie while continuing to protect routes with the `auth:sanctum` middleware.

### DIFF
--- a/sanctum.md
+++ b/sanctum.md
@@ -14,6 +14,7 @@
 - [SPA Authentication](#spa-authentication)
     - [Configuration](#spa-configuration)
     - [Authenticating](#spa-authenticating)
+    - [Remembering Authenticated Sessions](#remembering-authenticated-sessions)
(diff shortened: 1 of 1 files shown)


taylorotwell:
I feel like this uses a lot of words to just say "use authentication as normal"?

LucasCavalheri:
Thanks for the feedback, @taylorotwell!

I simplified the wording significantly. The intention was mainly to guide users toward Laravel's existing authentication documentation with a small clarification that Sanctum SPA authentication is session based and works with Laravel's standard authentication features.

I opened a simplified follow-up PR here: #11211

PR #11213, opened 2026-05-25
[13.x] Added note: teams prompt requires single-file components to be enabled

jonodonovan:
Adding clarity for why the prompt wasn't showing.


taylorotwell:
Let me think of how to solve this so we don't need this note.

PR #11216, opened 2026-05-27
[13.x] Document database query logging

ziadoz:
I find database query logging really useful for local development debugging, particularly the `getRawQueries()` method. 

I figured it would be nice to document some of this stuff, to save others having to dig around in the codebase to find it.


taylorotwell:
Think I'm going to skip this one for now because I fear people won't notice the memory stuff.

PR #11223, opened 2026-06-04
docs: string convert case helper

omaromp2:
## Summary

Documents the `Str::convertCase` helper in the string helpers documentation.

The new section includes examples demonstrating:

* converting strings to uppercase with `MB_CASE_UPPER`,
* the default `MB_CASE_FOLD` behavior,
* passing a custom encoding as the third argument.

This method is currently available in the framework and API documentation but is not documented in the main string helpers documentation.


taylorotwell:
I think I want to leave this one undocumented for now.

PR #11225, opened 2026-06-04
Add Payment Element integration documentation

### LINKED ISSUE #1541: Payment Element integration documentation
The Cashier Stripe documentation is only explaining how to use the [Stripe UI](https://stripe.com/docs/payments/elements) Card Element, which is now marked as "Legacy". While the new Payment Element is not documented.

It seems like the JS code should now call `await stripe.confirmPayment` instead of `await stripe.confirmCardSetup`. But it requires defining a `return_url`. A `payment_intent_client_secret` value is then passed as GET query string parameter (not ideal), which I guess can be used with Cashier.

I think the documentation should explain how to set this up properly. Or is this not supported officially?

Thanks!

youssef-al-mostafa:
Closes laravel/cashier-stripe#1541

The current documentation only covers the legacy Card Element which Stripe 
now marks as "Legacy". This PR adds a new "Using the Payment Element" section 
covering the three main use cases:

- Subscriptions (Setup Intent flow)
- Single charges (Payment Intent flow)  
- Updating the default payment method

The legacy Card Element docs are kept intact for backward compatibility.

### DIFF
--- a/billing.md
+++ b/billing.md
@@ -23,6 +23,7 @@
     - [Billing Portal](#billing-portal)
 - [Payment Methods](#payment-methods)
     - [Storing Payment Methods](#storing-payment-methods)
+    - [Using the Payment Element](#using-the-payment-element)
     - [Retrieving Payment Methods](#retrieving-payment-methods)
     - [Payment Method Presence](#payment-method-presence)
     - [Updating the Default Payment Method](#updating-the-default-payment-method)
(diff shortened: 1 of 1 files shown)


taylorotwell:
So does this leave the old legacy stuff in the docs as well?

youssef-al-mostafa:
Yes, I kept the old Card Element docs in place since a lot of existing apps still use it. 
Added the Payment Element as a new section below it so new projects have a modern reference without breaking anything for current users.
Happy to add a small deprecation note to the old section pointing to the new one if you'd prefer that approach.

taylorotwell:
Can you fix the styling to make it match the rest of the docs?

youssef-al-mostafa:
Updated formatting and prose style to match the existing docs. Ready for review.

youssef-al-mostafa:
Hi @taylorotwell  , just wanted to gently follow up on this PR, happy to make any additional changes if needed. Thanks for your time

PR #11226, opened 2026-06-05
Document whereNullSafeEquals methods

omaromp2:
## Summary

- document the `whereNullSafeEquals` query builder method
- document the `orWhereNullSafeEquals` query builder method
- add examples showing null-safe equality comparisons

### DIFF
--- a/queries.md
+++ b/queries.md
@@ -916,6 +916,28 @@ $users = DB::table('users')
     ->get();
 ```
 
+**whereNullSafeEquals / orWhereNullSafeEquals**
+
+The `whereNullSafeEquals` method may be used to compare a column's value against a given value while treating two `NULL` values as equal:
+
+```php
+$lastLoginIp = $request->input('last_login_ip');
+
+$users = DB::table('users')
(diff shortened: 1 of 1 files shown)


queries.md
```diff
@@ -916,6 +916,28 @@ $users = DB::table('users')
...
+The `orWhereNullSafeEquals` method may be used to add an "or" clause with a null-safe equality comparison:
+
+```php
+$lastLoginIp = $request->input('last_login_ip');
+
+$users = DB::table('users')
+    ->whereNullSafeEquals('last_login_ip', $lastLoginIp)
+    ->orWhereNullSafeEquals('registration_ip', $lastLoginIp)
+    ->get();
+```
```
shaedrich: Is this necessary to mention for this rule specifically? This is implicit knowledge when knowing how to work with Laravel's query builder

omaromp2: That's a fair point. I added the example mainly to demonstrate a practical use case for orWhereNullSafeEquals alongside whereNullSafeEquals, rather than to explain query builder boolean logic. If it feels redundant, I'm happy to remove it and keep the example more focused on the null-safe comparison itself.

PR #11228, opened 2026-06-06
[13.x] Make attributed `#[Scope]` usage reflect the actual state of the feature

shaedrich:
See https://github.com/laravel/framework/issues/55414#issuecomment-4638675296

TL;DR\
Due to the limitations of PHP as a language and its use in patterns of the framework, it is impossible to have protected attributed scopes accessed from outside the model as explained in the above issue.

The example, as it is in the docs right now, doesn't work.

Thanks to: @siarheipashkevich, @decadence, @JackBayliss

### DIFF
--- a/eloquent.md
+++ b/eloquent.md
@@ -1638,7 +1638,7 @@ class User extends Model
      * Scope a query to only include users of a given type.
      */
     #[Scope]
-    protected function ofType(Builder $query, string $type): void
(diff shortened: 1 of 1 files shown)


jackbayliss:
Thanks @shaedrich

shaedrich:
You're welcome, was long overdue

taylorotwell:
This isn't necessary from outside the model. I just tried it.

PR #11230, opened 2026-06-07
Document command class support for Artisan methods

zainphp:
This PR updates the documentation to reflect that `Artisan::call`, `Artisan::queue`, and `$this->call` methods now support passing the command class directly, in addition to the command signature. The code example for the console test is also updated.

I discovered this behavior was not previously documented and traced it back to the implementation in this framework PR: https://github.com/laravel/framework/pull/56530.

edited:
I apologize for missing the previous discussion on this topic in https://github.com/laravel/docs/pull/10683#issuecomment-3160555631, but I still believe documenting this is helpful for users, as it provides a type-safe and refactor-friendly alternative to string-based signatures, significantly reducing potential runtime errors.

### DIFF
--- a/artisan.md
+++ b/artisan.md
@@ -774,13 +774,13 @@ Route::post('/user/{user}/mail', function (string $user) {
 });
 ```
 
-You may also pass the entire Artisan command to the `call` method as a string:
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #11231, opened 2026-06-07
Add warning about messages() method conflict with RemembersConversati…

EmilioVar:
When using the RemembersConversations trait, defining a messages() method 
in the agent class causes PHP to silently prioritize the class method over 
the trait's implementation, resulting in conversation history never being 
loaded from the database.

This is a common pitfall because the make:agent Artisan command generates 
a messages() method by default. Developers adding RemembersConversations 
to an existing agent may not realize they need to remove it.

Confirmed via ReflectionMethod:
- Agent with messages(): method resolves to the agent class → trait ignored
- Agent without messages(): method resolves to RemembersConversations.php → works correctly

### DIFF
--- a/ai-sdk.md
+++ b/ai-sdk.md
@@ -316,7 +316,7 @@ class SalesCoach implements Agent, Conversational
 }
 ```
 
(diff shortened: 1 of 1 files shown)


EmilioVar:
Thank you for merging this, and thank you for all the work on the SDK.
It's a joy to work with!

PR #11245, opened 2026-06-19
Fix: Correct middleware for Client Credentials Grant in Passport v12 docs

pietrantonio91:
This PR corrects the documentation for the Client Credentials Grant in the 12.x branch to reflect the correct middleware class.

Currently, the documentation references the `EnsureClientIsResourceOwner` middleware. However, this class was introduced in Passport v13 and does not exist in Passport v12. Users following the current guide will encounter a `Class not found` error.

I have updated this section to use `CheckClientCredentials`, effectively reverting it to match the v11 documentation. This accurately reflects the available classes and the correct implementation for the Passport v12 release.

### DIFF
--- a/passport.md
+++ b/passport.md
@@ -955,26 +955,33 @@ Before your application can issue tokens via the client credentials grant, you w
 php artisan passport:client --client
 ```
 
-Next, assign the `Laravel\Passport\Http\Middleware\EnsureClientIsResourceOwner` middleware to a route:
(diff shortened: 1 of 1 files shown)


taylorotwell:
This really reflects the framework's version 12 documentation. Passport 13 is compatible with Laravel 12. Unfortunately we don't version package docs separately from the framework so the package docs are always the latest compatible version with that framework version.

PR #11247, opened 2026-06-22
[13.x] Fix the session cookie name in the upgrade guide

clarkewing:
The "Cache Prefixes and Session Cookie Names" section of the 13.x upgrade guide states that the default session cookie name now uses a hyphenated `-session` suffix. That part is inaccurate.

laravel/framework#56172 hyphenated the cache (`-cache-`) and Redis (`-database-`) prefixes, but it deliberately kept the session cookie underscore-based — see laravel/framework@4f8ee55a442656b9a128d0333ad67185e8ddea18, which sets the default to:

```php
Str::snake((string) env('APP_NAME', 'laravel')).'_session'
```

Since `Str::snake` and the previous `Str::slug(..., '_')` produce the same value for typical application names, the session cookie name — and therefore existing user sessions — is unchanged on upgrade. This corrects the example and clarifies that only the cache and Redis prefixes are affected.

### DIFF
--- a/upgrade.md
+++ b/upgrade.md
@@ -94,11 +94,11 @@ Or, if you are using [Laravel Herd's](https://herd.laravel.com) bundled copy of 
 
 **Likelihood Of Impact: Low**
 
-Laravel's default cache and Redis key prefixes now use hyphenated suffixes.
+Laravel's default cache and Redis key prefixes now use hyphenated suffixes. The default session cookie name is now generated using `Str::snake` rather than `Str::slug`, but it keeps its underscore suffix and is therefore unchanged for typical application names.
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
CleanShot 2026-06-22 at 07 56 09@2x

clarkewing:
Hi @taylorotwell,

Thanks for taking a look at this.

It seems you're working off the `laravel/laravel` skeleton's `config/session.php`, whereas I was working off the framework's, and they don't agree. The framework fallback resolves to `Str::snake(…).'_session'`, while the skeleton ships `Str::slug(…).'-session'`, so the default cookie name depends on which one is in effect.

I've written it up as laravel/framework#60569 so a call can be made on which way it should go.

PR #11284, opened 2026-07-09
[13.x] Remove `prependCount` option from `Str::plural` docs

JosephSilber:
Since we now have [a dedicated `Str::counted` method](https://github.com/laravel/framework/pull/60649), there's no need to document the `prependCount` parameter on `Str::plural`. People should use `Str::counted` instead.

### DIFF
--- a/strings.md
+++ b/strings.md
@@ -1250,16 +1250,6 @@ $singular = Str::plural('child', 1);
 // child
 ```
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hmm, I may do this on next major version.

JosephSilber:
@taylorotwell just making sure you realized that we are not removing it from the code. This is not a breaking change. We're just making it undocumented, since it's clunky compared to `counted`.

PR #11285, opened 2026-07-12
do not need to use Vite because it is aliased

hosni:
We do not need to import the `Vite` facade using the `@use` directive, because it is already aliased in:
```php
Illuminate\Support\Facades::defaultAliases();
```

See: https://github.com/laravel/framework/blob/3093ff3a61f88225f16fec35dda65e1e7c0867a4/src/Illuminate/Support/Facades/Facade.php#L321


taylorotwell:
We usually document this because not everyone uses aliases or has them enabled.

PR #11290, opened 2026-07-14
[13.x] Document the API starter kit

WendellAdriel:
Adds an API entry under the available starter kits and a dedicated section covering token-based authentication with Sanctum, the included endpoints, API versioning via the `X-API-Version` header, and the Scalar/Scramble API reference. Also updates the introduction, authentication, and teams sections to reflect that the API kit uses Sanctum instead of Fortify and supports teams through endpoints rather than screens.

### DIFF
--- a/starter-kits.md
+++ b/starter-kits.md
@@ -7,6 +7,7 @@
     - [Svelte](#svelte)
     - [Vue](#vue)
     - [Livewire](#livewire)
+    - [API](#api)
 - [Starter Kit Customization](#starter-kit-customization)
     - [React](#react-customization)
     - [Svelte](#svelte-customization)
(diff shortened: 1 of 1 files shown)


starter-kits.md
```diff
@@ -55,6 +60,9 @@ composer run dev
 
 Once you have started the Laravel development server, your application will be accessible in your web browser at [http://localhost:8000](http://localhost:8000).
 
+> [!NOTE]
+> The [API starter kit](#api) does not include a frontend, so there are no NPM dependencies to install. After creating your application, you may simply run `composer run dev`.
```
GertjanRoke, line 64: I guess this is going to be `php artisan dev` in the future

WendellAdriel: We plan to keep `composer dev` as the entrypoint and it will run `php artisan dev`

GertjanRoke: Aah cool! Didn't knew that, thanks for sharing

PR #11293, opened 2026-07-17
[13.x] Mail queue and readme.md modified

farad-tech:


shaedrich:
These two changes are unrelated. One would usually open to separate PRs for this. Chances are higher that your PR is approved when it has a description.

PR #11302, opened 2026-07-24
[13.x] Add warning about `shouldRun()` about safe running order

shaedrich:
As discussed in https://github.com/laravel/framework/discussions/60871#discussioncomment-17765733, `shouldRun()` should be used with caution and at one's own risk.

grafik


taylorotwell:
I don't find this note very clear or helpful 🤔

PR #11308, opened 2026-07-28
Document intl requirement for Number methods

igor-kamil:
### Context

We encountered this while using Laravel's `Number` helper in an application.

Our local development environment uses Laravel Herd, where PHP's `intl` extension is already available. The application therefore worked locally and passed the usual development checks. However, our deployed application uses a smaller custom PHP Docker image that did not include `intl`, so the missing dependency became apparent only after deployment when the affected code was executed.

The exception raised by Laravel is helpful once the code reaches that point. However, it was still somewhat surprising that a first-party helper documented as part of Laravel's core API depended on an extension that is neither listed among Laravel's general requirements nor mentioned in the Numbers documentation.

This may also become easier to encounter with AI-assisted development. Laravel Boost recommends using the `Number` helper for display formatting, and Boost provides agents with access to Laravel's documentation through its documentation search tool. Mentioning the dependency here could therefore help both developers reading the documentation and coding agents researching the helper before introducing it.

I found the previous related proposals in laravel/docs#9587 and laravel/docs#9773, and I understand the reasoning behind closing them:

* `intl` is not required to use Laravel as a whole;
* Laravel already throws an informative exception when an affected method is called.

This PR is not suggesting that `intl` should become a general Laravel requirement. It only adds a small note beside the affected feature so that the optional runtime dependency may be discovered before the exception occurs.

Other approaches may also be reasonable, such as documenting the requirement in Laravel Boost's guidance or adding `ext-intl` to Composer's suggested extensions. I started with the documentation note because it may help both human readers and agents using Laravel's documentation, but I would be happy to adjust the approach if another location is considered more appropriate.

Thank you in advance for taking the time to review this and for considering the best way to make this dependency discoverable.

### DIFF
--- a/helpers.md
+++ b/helpers.md
@@ -1638,6 +1638,9 @@ $last = last($array);
 <a name="numbers"></a>
 ## Numbers
 
+> [!NOTE]
+> The `Number` class methods that format, parse, or spell values require PHP's `intl` extension.
+
 <a name="method-number-abbreviate"></a>
 #### `Number::abbreviate()` {.collection-method}
 


shaedrich:
Wouldn't it make more sense to add it to [`./Illuminate/Support/composer.json`](https://github.com/laravel/framework/blob/13.x/src/Illuminate/Support/composer.json)?

igor-kamil:
Thanks. Yes, that may be a better place for this, and I mentioned Composer's `suggest` section as one of the possible approaches in the PR description.

My only uncertainty is whether adding it solely to `src/Illuminate/Support/composer.json` would surface the suggestion for standard Laravel applications. The default application requires `laravel/framework`, which replaces `illuminate/support`, and the framework has its own root `composer.json` with a separate `suggest` section. Would this therefore need to be added to both manifests or only to the root framework manifest?

I initially started with the documentation because it places the requirement beside the affected API and may also make it discoverable to tools searching Laravel's documentation. A Composer suggestion is certainly appropriate as well, although it remains informational and may be easy to miss in installation or deployment logs.

I’m happy to move the change to the framework repository if that is the preferred approach.

shaedrich:
I just have a gut feeling that, since it isn't in the docs, it may not be wanted there. That's why I brought it up 🤞🏻

taylorotwell:
The exception tells you this so I think that is sufficient.

PR #11312, opened 2026-07-31
[13.x] Document dominant color support for image methods

dr-codswallop:
Documents the dominant color support added in laravel/framework#60932.

The `contain` and `rotate` methods now accept `'dominant'` as a background string, and images expose a `dominantColor()` method that returns the average color as a hex string. 

Dominant color is calculated by downsampling the image to a single pixel.


taylorotwell:
Already done.

PR #11313, opened 2026-07-31
[13.x] Documents `--blade` option on `pint`

nunomaduro:


luisprmat:
Is there any indication of how to override the Blade formatter rules?

PR #11332, opened 2026-08-12
Document dropForeignIdFor behavior change in upgrade guide

elhamlin:
[laravel/framework#54102](https://github.com/laravel/framework/pull/54102) was a fix to make the behavior of `dropForeignIdFor` mirror the behavior of `foreignIdFor`.  This is a good change but isn't really documented anywhere easy to find.  The docs don't mention the function and the api auto documentation doesn't reflect any change, I had to dig into the source code and compare to figure out what was going on.

I lost a few hours trying to figure out why our migrations were suddenly breaking when upgrading from 11x to 12x, so I think it'd be good to have in the upgrade guide.


taylorotwell:
Let me take a look at this one.

PR #11346, opened 2026-08-26
[13.x] Document Redis Sentinel configuration for Predis

tgivslife:
Predis has supported Redis Sentinel through Laravel's Redis configuration for years, but the documentation never mentions it, so the configuration shape is only discoverable through third-party packages. 

This adds a "Redis Sentinel" subsection to the Predis section showing the `replication`, `service` and `parameters` options with a list of Sentinel servers, and warns that the default `options.cluster` value must be removed, since Predis would otherwise treat the Sentinel list as a Redis Cluster. Every configuration shown was verified against `RedisManager` on 13.x.

### DIFF
--- a/redis.md
+++ b/redis.md
@@ -178,6 +178,52 @@ In addition to the default configuration options, Predis supports additional [co
 ],
 ```
 
+<a name="predis-sentinel"></a>
+#### Redis Sentinel
+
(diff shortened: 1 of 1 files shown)


taylorotwell:

PR #11352, opened 2026-09-01
[13.x] Clarify the two makeSearchableUsing methods

danielebarbaro:
Hi, I finally noticed that there are 2 methods with same name, `makeSearchableUsing`. It took me a couple of headaches to work out the difference. Maybe a small note helps to "polish" this amazing doc, without giving the same headache to other devs.


taylorotwell:
Hmm let me think on this one!

PR #11356, opened 2026-09-08
[13.x] Clarify meta field in paginated Resource response

gcugnet:
This PR aims to clarify the documentation about the PaginatedResourceCollection return.

Another PR was created in 2021 to change the return behaviour but it has been closed to avoid merging a breaking change : [[8.x] Exclude links from PaginatedResourceResponse - #36101](https://github.com/laravel/framework/pull/36101)

While it is comprehensible, I think the documentation still lacks a bit of information about the nested `links` field.

I opened a discussion yesterday about it, where you can find more information on that case: [https://github.com/laravel/framework/discussions/61473](https://github.com/laravel/framework/discussions/61473)


eloquent-resources.md
```diff
@@ -562,7 +582,9 @@ Or, for convenience, you may use the paginator's `toResourceCollection` method,
 return User::paginate()->toResourceCollection();
 ```
 
-Paginated responses always contain `meta` and `links` keys with information about the paginator's state:
+Paginated responses always contain `meta` and `links` keys at the root level
+with information about the paginator's state. Another `links` field which
+provides a detailed array of page links is also nested under the `meta` field:
```
shaedrich, line 587: Just from this paragraph, I would still not get the difference and would still have to read the related discussion. Maybe, we can add some more details here.

PR #11358, opened 2026-09-09
Fix billing table of contents spelling

haddybhaiya:
The billing table of contents says ‘Canceling Subscriptions’, while the linked section is titled ‘Cancelling Subscriptions’. This updates the table of contents to match the section.


taylorotwell:
Canceling should be the way

haddybhaiya:
> Canceling should be the way

Ngl, It sounds poetic >⁠.⁠<

PR #11361, opened 2026-09-13
docs(auth): remove duplicate regenerate

vexcited:
`Auth::attempt` calls the `Auth::login` method which calls the protected method `Auth::updateSession` that finally calls `$this->session->regenerate(true)`.

Why would the documentation ask you to regenerate the session which has already been regenerated?


taylorotwell:
I think it's fine to leave for now.

PR #11368, opened 2026-09-18
Document vector indexes

Rhaima96:
Vector columns, similarity queries and the `AsVector` cast are documented, but vector indexes are not — even though `vectorIndex()` and `dropVectorIndex()` have both shipped.

- Adds `vectorIndex()` to the available index types table.
- Adds `dropVectorIndex()` to the dropping indexes table.
- Mentions in the `vector()` section that vector columns need pgvector or MariaDB 11.7+, and that calling `index()` on a vector column creates a vector index.

This matters in practice: without an index, a similarity query scans the whole table.


migrations.md
```diff
@@ -1195,12 +1195,18 @@ The `vector` method creates a `vector` equivalent column:
 $table->vector('embedding', dimensions: 100);
 ```
 
-When utilizing PostgreSQL, the `pgvector` extension must be loaded before `vector` columns can be created:
+Vector columns are supported on PostgreSQL connections using the `pgvector` extension and MariaDB 11.7 or later. When utilizing PostgreSQL, the `pgvector` extension must be loaded before `vector` columns can be created:
```
shaedrich: Now, the second sentence sounds a bit redundant—we might be able to improve that a bit:

```suggestion
Vector columns are supported on PostgreSQL connections using the `pgvector` extension and MariaDB 11.7 or later. When utilizing PostgreSQL, `pgvector` must be loaded before `vector` columns can be created:
```
or

```suggestion
Vector columns are supported on PostgreSQL connections using the `pgvector` extension and MariaDB 11.7 or later. When utilizing PostgreSQL, the extension must be loaded before `vector` columns can be created:
```

shaedrich:
Err, something went wrong in 4312e26:

image

```diff
- <a name="column-method-vector"></a>
- #### `vector()` {.collection-method}
+ <a name="column-method-"></a>
+ #### `()` {.collection-method}
```

